PR #295, opened 2015-04-01
Update node_data::remove to use new equals() method

bdutro:
When the equals() function was moved to be a member of the node class, the call to equals() in node_data::remove() was not updated to match the new implementation.

### DIFF
--- a/include/yaml-cpp/node/detail/impl.h
+++ b/include/yaml-cpp/node/detail/impl.h
@@ -133,7 +133,7 @@ inline bool node_data::remove(const Key& key, shared_memory_holder pMemory) {
     return false;
 
   for (node_map::iterator it = m_map.begin(); it != m_map.end(); ++it) {
-    if (equals(*it->first, key, pMemory)) {
+    if (it->first->equals(key, pMemory)) {
       m_map.erase(it);
       return true;
     }
--- a/test/node/node_test.cpp
+++ b/test/node/node_test.cpp
@@ -39,6 +39,13 @@ TEST(NodeTest, SimpleAppendSequence) {
   EXPECT_TRUE(node.IsSequence());
 }
 
+TEST(NodeTest, MapElementRemoval) {
+  Node node;
+  node["foo"] = "bar";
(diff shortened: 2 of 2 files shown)


jbeder:
Wow, that's bad! Thanks!

Can you add a test that exercises this logic? I assume you can write a test that fails to compile without this fix?

WrinklyNinja:
> Presumably there's a test case that we can write that won't compile or won't link?

I added the code below to `test/node/node_test.cpp`, and it produces a compiler error that disappears with this PR merged. (It is a compiler error, not a linker error, I must have confused it with something else.)

```
TEST(NodeTest, MapElementRemoval) {
    Node node;
    node["foo"] = "bar";
    node.remove("foo");
    EXPECT_TRUE(!node["foo"]);
}
```

bdutro:
WrinklyNinja beat me to it- that test was what I had in mind.

jbeder:
@bdutro , can you add that test to your pull request? Please squash the commit, format in the appropriate style, etc.

bdutro:
I'm not too experienced with GitHub, so bear with me. I added a new commit to the pull request that adds the test. Will this work?

jbeder:
No problem - to squash the commit, use git rebase -i and follow the instructions on the resulting message to squash into a single commit. If your repo looks like

A (master)
 \_ B _ C (patch-1)

Your command, when you're at commit C in the branch patch-1, will look something like

git rebase -i master

bdutro:
Sorry this took so long. I pushed the rebased commits.

bdutro:
Ack, sorry. Fixed.

jbeder:
Overall the code looks good, but why are there six commits listed? Can you make it so that there's only one commit, and then I'll pull that?

(I don't really understand the github UI, so I can't really help there.)

bdutro:
Ok, so the problem is that GitHub doesn't like squashing commits that have already been pushed to the repo. I created a new branch with the squashed commit, so I'll close this pull request and open another with the new branch.

PR #297, opened 2015-04-02
Revert compiler flag variables in CMake config.

WrinklyNinja:
This is a partial revert of d59586630e9444f04ee12577e3e5977e93fc9b86, which breaks using CMake (3.2.1) to generate an MSVC project which uses the static runtime. This is due to the introduction of the `yaml_c_flags` and `yaml_cxx_flags` variables.

The main issue is that [this foreach](https://github.com/WrinklyNinja/yaml-cpp/blob/d59586630e9444f04ee12577e3e5977e93fc9b86/CMakeLists.txt#L191-L199) is broken, as `yaml_c_flags` and `yaml_cxx_flags` never contain `/MD`, and their `_RELEASE`, `_DEBUG`, etc. variations are all initialised as empty strings, unlike their `CMAKE_` flags equivalents (which do initially contain `/MD`).

For example, some initial values:

```
CMAKE_CXX_FLAGS = "/DWIN32 /D_WINDOWS /W3 /GR /EHsc"
CMAKE_CXX_FLAGS_RELEASE = "/MD /O2 /Ob2 /D NDEBUG"
yaml_cxx_flags = "/DWIN32 /D_WINDOWS /W3 /GR /EHsc"
yaml_cxx_flags_RELEASE = ""
```

There's a secondary issue in that when you set the [target compile flags](https://github.com/WrinklyNinja/yaml-cpp/blob/d59586630e9444f04ee12577e3e5977e93fc9b86/CMakeLists.txt#L246-L248), you're setting flags that are _in addition_ to the existing flags (which are stored in the `CMAKE_` flag variables). As far as I can tell, that means the switch to the `yaml_` flag variables doesn't do anything other than break the foreach.

As such, I've reverted all the changes relating to the the `yaml_c_flags` and `yaml_cxx_flags` variables. I've tested it on MSVC with static and shared runtimes set, and both work. I haven't tested it on GCC or Clang, but if I've understood everything correctly, there should be no difference in effect.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,9 +118,6 @@ include_directories(${YAML_CPP_SOURCE_DIR}/include)
 ###
 ### General compilation settings
 ###
-set(yaml_c_flags ${CMAKE_C_FLAGS})
-set(yaml_cxx_flags ${CMAKE_CXX_FLAGS})
-
 if(BUILD_SHARED_LIBS)
 	set(LABEL_SUFFIX "shared")
 else()
@@ -179,7 +176,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR
 		set(GCC_EXTRA_OPTIONS "${GCC_EXTRA_OPTIONS} ${FLAG_TESTED}")
 	endif()
 	#
-	set(yaml_cxx_flags "-Wall ${GCC_EXTRA_OPTIONS} -pedantic -Wno-long-long -std=c++11 ${yaml_cxx_flags}")
+	set(yaml_cxx_flags "-Wall ${GCC_EXTRA_OPTIONS} -pedantic -Wno-long-long -std=c++11")
 
 	### Make specific
 	if(${CMAKE_BUILD_TOOL} MATCHES make OR ${CMAKE_BUILD_TOOL} MATCHES gmake)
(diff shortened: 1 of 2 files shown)


jbeder:
The issue is that this messes with the settings that gtest and gmock use to prevent warnings. When we set our own local variables, we don't interfere with other ones.

For example, when I build this using gcc, I get warnings such as

/Users/beder/src/yaml-cpp/test/gmock-1.7.0/gtest/include/gtest/gtest-typed-test.h:239:47: warning: 
      variadic macros are a C99 feature [-Wvariadic-macros]
# define REGISTER_TYPED_TEST_CASE_P(CaseName, ...) \
                                              ^

WrinklyNinja:
Hmm, that shouldn't happen, `yaml-test-flags` contains `-Wno-c99-extensions` and still gets added to the compile flags... I'll have to test it out in a virtual machine myself, maybe I'm missing something.

WrinklyNinja:
OK, I've squashed in a few minor changes to fix the warning suppression, and tested on GCC. The trick was to do the same for the `yaml-cpp` target as you did for the `run-tests` target, sticking in the extra flags (eg. `-Wall`) in a separate variable and adding it to the compile flags via `set_target_properties()`.

I had assumed that the warning suppression flags would override the warning flags, but that was wrong.

WrinklyNinja:
Closing this as I'm no longer using static runtime linking and it looks like 96598c5c25c9dc44f48cb6c0987cd2935842b7c1 may have fixed the issue anyway.

Issue #41, opened 2015-03-30
Unsupported merges


GoogleCodeExporter:
```
Current version of otherwise great yaml-cpp does not seem to support merges
in a way described for example here:
http://en.wikipedia.org/wiki/YAML#Data_merge_and_references

I don't have a deep knowledge of yaml-cpp internals (I am just its user),
but maybe this is somehow linked with the cloning problem - not yet
sufficiently solved.

I have tested the simple form of a merge in yaml-cpp 0.2.0 on RedHat Linux
3.4.6-9 using gcc 3.4.6 (20060404) - it doesn't work, but at least it
doesn't hurt the parser. With that gcc I am unable to compile 0.2.1 because
of new conversion operators in nodeimpl.h (it reported an error on lines
with "node.Read<T>" and the thing that caused that error was very probably
the templated read).

Do you plan to support merges in some future version?
```

Original issue reported on code.google.com by `pole...@gmail.com` on 8 Sep 2009 at 10:38

SimplyKnownAsG:
looks like this is probably a duplicate of #300 now

agirault:
There are many duplicates to this feature request: #353, #617, #638, #683.

@Ortham proposed an implementation in #300 back in 2015 (!) that was garnering a lot of support, then closed in 2021 because it's not officially in the spec: https://github.com/jbeder/yaml-cpp/pull/300#issuecomment-877660781

Now:
- this issue is still open, and mentioned by many projects
- almost every online documentation of YAML that discusses "anchors" will introduce the `<<` merge key capability
- @Ortham did address the concern regarding the lack of spec here: https://github.com/jbeder/yaml-cpp/pull/300#issuecomment-428086289

Based on this, can we re-open the discussion to consider merge keys as a feature (even if optionally built) of yaml-cpp? Or should this issue be closed for the same reason that #300 was closed?

jbeder:
This is a good point, and I'm not sure.

This is a pretty big feature, and would need thorough review. I'm the only maintainer, and I unfortunately don't have time for it. I also (as you can see by this history of this feature) never really thought much of this feature.

I've been unable to find someone to help maintain this project (any takers?). If I found someone who I trusted and is willing to support and review this feature, then I'd be open to accepting this feature.

Also, thanks for the links to the other feature requests, and they should at least be closed as dupes of this one.

alex-thiessen-for-siemens:
[...]

PR #300, opened 2015-04-04
Add support for merge keys.

Ortham:
This would close issue #41. While the merge key type isn't mentioned in the YAML 1.2 spec, there's a [spec](http://yaml.org/type/merge.html) for YAML 1.1, and I know I'm not alone in finding it a very useful feature.

### DIFF
--- a/test/integration/load_node_test.cpp
+++ b/test/integration/load_node_test.cpp
@@ -211,6 +211,24 @@ TEST(NodeTest, NestedMergeKeys) {
   ASSERT_FALSE(!node["a"]);
   EXPECT_EQ(1, node["a"].as<int>());
 }
+
+TEST(NodeTest, AnchorAndMergeKey) {
+  Node node = YAML::Load(R"(
+    a_root: &root_anchor
+      key1: value1
+      key2: value2
+    b_child:
+      <<: *root_anchor
+      key2: value2_override
+  )");
+
+  ASSERT_FALSE(!node["a_root"]);
+  ASSERT_FALSE(!node["b_child"]);
+  EXPECT_EQ("value1", node["a_root"]["key1"].as<std::string>());
+  EXPECT_EQ("value2", node["a_root"]["key2"].as<std::string>());
+  EXPECT_EQ("value1", node["b_child"]["key1"].as<std::string>());
+  EXPECT_EQ("value2_override", node["b_child"]["key2"].as<std::string>());
+}
 #else
 TEST(NodeTest, MergeKeySupport) {
   Node node = Load("{<<: {a: 1}}");
--- a/test/integration/load_node_test.cpp
+++ b/test/integration/load_node_test.cpp
@@ -185,6 +185,39 @@ TEST(LoadNodeTest, DereferenceIteratorError) {
   EXPECT_THROW(node.begin()->begin()->Type(), InvalidNode);
 }
 
+#ifdef YAML_CPP_SUPPORT_MERGE_KEYS
+TEST(NodeTest, MergeKeyScalarSupport) {
+  Node node = Load("{<<: {a: 1}}");
+  ASSERT_FALSE(!node["a"]);
+  EXPECT_EQ(1, node["a"].as<int>());
+}
+
+TEST(NodeTest, MergeKeyExistingKey) {
+  Node node = Load("{a: 1, <<: {a: 2}}");
+  ASSERT_FALSE(!node["a"]);
+  EXPECT_EQ(1, node["a"].as<int>());
+}
+
+TEST(NodeTest, MergeKeySequenceSupport) {
+  Node node = Load("<<: [{a: 1}, {a: 2, b: 3}]");
+  ASSERT_FALSE(!node["a"]);
+  ASSERT_FALSE(!node["b"]);
+  EXPECT_EQ(1, node["a"].as<int>());
+  EXPECT_EQ(3, node["b"].as<int>());
+}
+
+TEST(NodeTest, NestedMergeKeys) {
+  Node node = Load("{<<: {<<: {a: 1}}}");
+  ASSERT_FALSE(!node["a"]);
+  EXPECT_EQ(1, node["a"].as<int>());
+}
+#else
+TEST(NodeTest, MergeKeySupport) {
+  Node node = Load("{<<: {a: 1}}");
+  ASSERT_FALSE(node["a"]);
+}
+#endif
+
 TEST(NodeTest, EmitEmptyNode) {
   Node node;
   Emitter emitter;
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ os:
 compiler:
   - clang
   - gcc
+env:
+  - YAML_CPP_SUPPORT_MERGE_KEYS=ON
+  - YAML_CPP_SUPPORT_MERGE_KEYS=OFF
 before_install:
   - |
     if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
@@ -17,7 +20,7 @@ before_install:
 before_script:
   - mkdir build
   - cd build
-  - cmake ..
+  - cmake .. -DYAML_CPP_SUPPORT_MERGE_KEYS=$YAML_CPP_SUPPORT_MERGE_KEYS
 script:
   - make
   - test/run-tests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ enable_testing()
 option(YAML_CPP_BUILD_TESTS "Enable testing" ON)
 option(YAML_CPP_BUILD_TOOLS "Enable parse tools" ON)
 option(YAML_CPP_BUILD_CONTRIB "Enable contrib stuff in library" ON)
+option(YAML_CPP_SUPPORT_MERGE_KEYS "Support YAML merge keys ('<<') in yaml-cpp's executable targets. Use '#define YAML_CPP_SUPPORT_MERGE_KEYS' instead when linking from another project." OFF)
 
 ## Build options
 # --> General
@@ -95,6 +96,10 @@ if(YAML_CPP_BUILD_CONTRIB)
 	file(GLOB contrib_private_headers "src/contrib/[a-zA-Z]*.h")
 else()
 	add_definitions(-DYAML_CPP_NO_CONTRIB)
+endif()
+
+if (YAML_CPP_SUPPORT_MERGE_KEYS)
+    add_definitions(-DYAML_CPP_SUPPORT_MERGE_KEYS)
 endif()
 
 set(library_sources
--- a/include/yaml-cpp/node/detail/node.h
+++ b/include/yaml-cpp/node/detail/node.h
@@ -122,12 +122,23 @@ class node {
     // NOTE: this returns a non-const node so that the top-level Node can wrap
     // it, and returns a pointer so that it can be NULL (if there is no such
     // key).
-    return static_cast<const node_ref&>(*m_pRef).get(key, pMemory);
+    node* value = static_cast<const node_ref&>(*m_pRef).get(key, pMemory);
+#ifdef YAML_CPP_SUPPORT_MERGE_KEYS
+    if (!value || value->type() == NodeType::Undefined) {
+      return get_value_from_merge_key(key, value, pMemory);
+    }
+#endif
+    return value;
   }
   template <typename Key>
   node& get(const Key& key, shared_memory_holder pMemory) {
     node& value = m_pRef->get(key, pMemory);
     value.add_dependency(*this);
+#ifdef YAML_CPP_SUPPORT_MERGE_KEYS
+    if (value.type() == NodeType::Undefined) {
+      return *get_value_from_merge_key(key, &value, pMemory);
+    }
+#endif
     return value;
   }
   template <typename Key>
@@ -159,6 +170,33 @@ class node {
   }
 
  private:
+#ifdef YAML_CPP_SUPPORT_MERGE_KEYS
+  template <typename Key>
+  inline node* get_value_from_merge_key(const Key& key, node* currentValue,
+                                        shared_memory_holder pMemory) const {
+    node* mergeValue =
+      static_cast<const node_ref&>(*m_pRef).get(std::string("<<"), pMemory);
+    if (!mergeValue) {
+      return currentValue;
+    }
+    if (mergeValue->type() == NodeType::Map) {
+      return &mergeValue->get(key, pMemory);
+    }
+    if (mergeValue->type() == NodeType::Sequence) {
+      for (const_node_iterator it = mergeValue->begin();
+            it != mergeValue->end(); ++it) {
+        if (it->pNode && it->pNode->type() == NodeType::Map) {
+          node* value = it->pNode->get(key, pMemory);
+          if (value && value->type() != NodeType::Undefined) {
+            return value;
+          }
+        }
+      }
+    }
+    return currentValue;
+  }
+#endif
+
   shared_node_ref m_pRef;
   typedef std::set<node*> nodes;
   nodes m_dependencies;


include/yaml-cpp/node/impl.h
```diff
@@ -369,9 +369,8 @@ inline const Node Node::operator[](const Key& key) const {
   EnsureNodeExists();
   detail::node* value = static_cast<const detail::node&>(*m_pNode)
                             .get(detail::to_value(key), m_pMemory);
-  if (!value) {
-    return Node(ZombieNode);
-  }
+  if (!value || value->type() == NodeType::Undefined)
```
jbeder: Add curly braces.

include/yaml-cpp/node/impl.h
```diff
@@ -381,10 +380,37 @@ inline Node Node::operator[](const Key& key) {
     throw InvalidNode();
   EnsureNodeExists();
   detail::node& value = m_pNode->get(detail::to_value(key), m_pMemory);
+  if (value.type() == NodeType::Undefined)
```
jbeder: Add curly braces.

test/integration/load_node_test.cpp
```diff
@@ -185,6 +185,20 @@ TEST(LoadNodeTest, DereferenceIteratorError) {
   EXPECT_THROW(node.begin()->begin()->Type(), InvalidNode);
 }
 
+TEST(NodeTest, MergeKeyScalarSupport) {
```
jbeder: Needs more testing. Specifically, tests where it takes the main entry instead of the merge.

jbeder: Also, multiple levels of merges (that's legal, right?)

WrinklyNinja: Yes, I think multiple-level merges are legal, I'll add a test for them.

> Specifically, tests where it takes the main entry instead of the merge.

I'm not sure what you mean by this.

jbeder:
In addition to the nits on the CL, I have a couple high-level questions:

1) Why implement this at the YAML::Node level rather than the YAML::detail::node level? Couldn't node::get() look up the merge too? (honest question - I haven't tried, so I'm not sure of the answer)

2) Right now, if you look up node["<<"], it'll give you that sequence or map; is that what we want? Or should that be undefined since we've "used" the value for merges.

3) I really worry about this giving unexpected results to people, especially since it's not in the YAML 1.2 spec. If you don't know about merges, this is very strange behavior. I'd love to be able to turn them on or off (default off, I think), but I'm not sure how best to do that. What do you think?

test/integration/load_node_test.cpp
jbeder: Like, the main dictionary has the key "foo", and so does the merge entry - the test should confirm that "foo" refers to the main dictionary.

WrinklyNinja:
> 1) Why implement this at the YAML::Node level rather than the YAML::detail::node level? Couldn't node::get() look up the merge too?

TBH I haven't looked at `YAML::detail::node`. This was just the most straightforward way to implement a feature I wanted. If you feel that putting it in `YAML::detail::node` would be better design, I can take a look at it.

> 2) Right now, if you look up node["<<"], it'll give you that sequence or map; is that what we want? Or should that be undefined since we've "used" the value for merges.

I think there are probably situations where being able to access the merge node itself would be beneficial, eg. during document conversion, where you might want to change what gets merged into a node before emitting the YAML.

> 3) I really worry about this giving unexpected results to people, especially since it's not in the YAML 1.2 spec. If you don't know about merges, this is very strange behavior. I'd love to be able to turn them on or off (default off, I think), but I'm not sure how best to do that. What do you think?

Wrap the code in a preprocessor definition, eg. `YAML_CPP_SUPPORT_MERGE_KEYS`? I can't think of any use-case where having a runtime switch would be beneficial. I'd be perfectly happy with a default-off switch though.

jbeder:
I suspect it'd be better to put the implementation in YAML::detail::node - the purpose of YAML::Node is just for wrapping the actual work in an API.

The problem with a preprocessor definition is that it's going in the headers, so we have to make sure people set it when they install the library. Maybe a CMake flag that generates the definition? That seems super hacky...

WrinklyNinja:
> I suspect it'd be better to put the implementation in YAML::detail::node - the purpose of YAML::Node is just for wrapping the actual work in an API.

Okay, I'll do it then.

> The problem with a preprocessor definition is that it's going in the headers, so we have to make sure people set it when they install the library. Maybe a CMake flag that generates the definition? That seems super hacky...

Well, what if the `#ifdef` block was inside `GetValueFromMergeKey()`, so that without the definition the function just didn't do anything, eg.

```
template <typename Key>
inline Node Node::GetValueFromMergeKey(const Key& key,
                                       detail::node* currentValue) const {
#ifdef YAML_CPP_SUPPORT_MERGE_KEYS
  detail::node* mergeValue = static_cast<const detail::node&>(*m_pNode)
                                 .get(detail::to_value("<<"), m_pMemory);
  if (mergeValue && mergeValue->type() != NodeType::Undefined) {
    Node mergeNode(*mergeValue, m_pMemory);
    // Merge keys are used with maps and sequences of maps. Aliases may be used,
    // but yaml-cpp resolves them automatically.
    if (mergeNode.IsMap()) {
      return mergeNode[key];
    } else if (mergeNode.IsSequence()) {
      for (const_iterator it = mergeNode.begin(); it != mergeNode.end(); ++it) {
        if (it->IsMap() && (*it)[key]) {
          return (*it)[key];
        }
      }
    }
  }
#endif
  if (!currentValue) {
    return Node(ZombieNode);
  }
  return Node(*currentValue, m_pMemory);
}
```

Then there wouldn't be any need for changing the header files, and you'd just need to tell people to either pass a `-DYAML_CPP_SUPPORT_MERGE_KEYS` arg to CMake, or you could add a CMake toggle that would do the defining.

jbrezmorf:
Can not build since to_value wanted in GetValueFromMergeKey is defined in node/impl.h. 
I have used explicit conversion to std::string in my fork, and it works. 
Looking forward to have this feature in official master.

WrinklyNinja:
> Can not build since to_value wanted in GetValueFromMergeKey is defined in node/impl.h.

Works for me, but I suppose it's bad design anyway. I've replaced `to_value` with `std::string` too now.

WrinklyNinja:
@jbeder: I've just rebased this PR on top of the latest `master`. Is there anything I need to address to get this merged?

jbeder:
Did we ever work out a good API to make sure that this doesn't confuse people (and can be turned on/off)?

WrinklyNinja:
No, but I'm still partial to using a `YAML_CPP_SUPPORT_MERGE_KEYS` preprocessor definition.

E.g. in the yaml-cpp code it would be

```
#ifdef YAML_CPP_SUPPORT_MERGE_KEYS
    if (value.type() == NodeType::Undefined) {
       return *GetValueFromMergeKey(key, &value, pMemory);
    }
#endif
```

Somewhere in yaml-cpp's `CMakeLists.txt` there would be

```
option(YAML_CPP_SUPPORT_MERGE_KEYS "Build with support for YAML merge keys ('<<')" OFF)

IF (YAML_CPP_SUPPORT_MERGE_KEYS)
  add_definitions(-DYAML_CPP_SUPPORT_MERGE_KEYS)
ENDIF()
```

If I wanted to include merge key support I'd build it with `cmake ... -DYAML_CPP_SUPPORT_MERGE_KEYS=ON`, and include yaml-cpp as

```
#define YAML_CPP_SUPPORT_MERGE_KEYS
#include <yaml-cpp/yaml.h>
```

That way users have to explicitly opt-in to the behaviour, so there shouldn't be any nasty surprises.

WrinklyNinja:
I've committed my proposal for turning on/off merge key support.

joseph-ireland:
I disagree that this feature would be unexpected. Every major implementation does this with no easy option to disable: pyyaml, ruby stdlib, java snakeyaml, etc. Also it is mentioned in common introductory material, including the YAML wikipedia page.

Also, it is (almost) part of the 1.2 spec\*:

> ###10.4. Other Schemas
> ...
>  In addition, it is strongly recommended that such schemas make as much use as possible of the the YAML tag repository at http://yaml.org/type/. This repository provides recommended global tags for increasing the portability of YAML documents between different applications. 

which links to a list of recommended *interoperable schemas* to support, which includes this merge feature.

\*The *interoperable schemas* are intentionally left out of the official spec to allow them to evolve independently.

WrinklyNinja:
@jbeder How's my implementation of the support switch? Does @joseph-ireland's spec find (which I'd missed) change your stance on whether the support should be default on, default off or always on? What's (if anything) blocking this getting merged?

jbrezmorf:
Strong support to merge this. Currently, we use the old version of "the merge support" and looking forward to the merge into master.

joseph-ireland:
There is an issue with editing documents in this PR. If you try to assign `node["key"]=x`, and there is a default value merged in with `<<: &DEFAULTS`, then it will assign into the defaults, and potentially change lots of other merged maps too.

I think a better approach would be to implement it at the NodeBuilder level, and do a single-level copy (this is what pyyaml does), i.e. OnMapEnd, if (node["<<"].IsMap()), loop over its keys and insert them into the node if not overriden, then remove the "<<". This will eliminate the *DEFAULT  ref in the above case, but not any nested references, acting consistently with other implementations

WrinklyNinja:
> There is an issue with editing documents in this PR. If you try to assign node["key"]=x, and there is a default value merged in with <<: *DEFAULTS, then it will assign into the defaults, and potentially change lots of other merged maps too.

I'm not sure I understand, what would that look like (eg. as a test)?

I don't think I'm ever in a position where I edit a Node that may contain merged keys, so I guess I missed something...

joseph-ireland:
I'd expect something like this to fail:
```
TEST(MergeKey, CheckAssignmentDoesntOverwriteDefault) {
    auto node = YAML::Load(R"(
defaults : &DEFAULTS
    merged: default
a:
    <<: *DEFAULTS
b:
    <<: *DEFAULTS
)");

    node["a"]["merged"] = "updated"     
    // The node returned & assigned to will be node["defaults"]["merged"]
    // So I expect this will fail with node["b"]["merged"] == "updated"
    EXPECT_EQ(node["b"]["merged"].Scalar(), "default"); 
}
```

WrinklyNinja:
Ah, I understand now. I'll see about changing my implementation so it passes that test.

WrinklyNinja:
So I tried a naive reimplementation in NodeBuilder:

```
void NodeBuilder::OnMapEnd() {
  detail::node& node = *m_stack.back();
  const std::string mergeKey = "<<";
  detail::node& mergeValue = node.get(mergeKey, m_pMemory);

  if (mergeValue.type() == NodeType::Sequence) {
    for (detail::node_iterator sit = mergeValue.begin();
         sit != mergeValue.end(); ++sit) {
      InsertAll(*sit->pNode, node);
    }
  } else {
    InsertAll(mergeValue, node);
  }

  node.remove(mergeKey, m_pMemory);

  assert(m_mapDepth > 0);
  m_mapDepth--;
  Pop();
}

void NodeBuilder::InsertAll(detail::node& sourceNode, detail::node& destinationNode) {
  if (sourceNode.type() == NodeType::Map) {
    for (detail::node_iterator it = sourceNode.begin();
         it != sourceNode.end(); ++it) {
      destinationNode.insert(*it->first, *it->second, m_pMemory);
    }
  }
}
```

However, this fails because it's inserting references to existing nodes, [as documented for the higher-level API](https://github.com/jbeder/yaml-cpp/wiki/Tutorial#building-nodes). I had a look through the code but couldn't find any way to create a copy of a node (which would have to be recursive for protection against overwriting merged map values).

I may be missing something, but it seems like editing nodes is broken by design in general, and not just for merged nodes. Merging nodes just makes it easier to hit that brokenness, which I think is way beyond the scope of this PR to fix.

WrinklyNinja:
@jbeder I've rebased this PR off a recent `master` commit, is there anything blocking it from being merged?

hofbi:
Could you also add this as a Unit Test. This is related to issue #617 and I'm not sure if this will also work.

```cpp
TEST(NodeTest, AnchorAndMergeKey) {
  Node node = Load("{a_root: &root_anchor {key1: value1, key2: value2}, b_child: {<<: *root_anchor, k ey2: value2_override}}");
  ASSERT_FALSE(!node["a_root"]);
  ASSERT_FALSE(!node["b_child"]);
  EXPECT_EQ(value1, node["a_root"]["key1"].as<std::string>());
  EXPECT_EQ(value2, node["a_root"]["key2"].as<std::string>());
  EXPECT_EQ(value1, node["b_child"]["key1"].as<std::string>());
  EXPECT_EQ(value2_override, node["b_child"]["key2"].as<std::string>());
}
```

hofbi:
Looks goot to me. 

I just saw that PR #567 will also add such kind of test, so this PR should also fix the test of #567

hofbi:
@WrinklyNinja Could you rebase and squash again that we can merge this

WrinklyNinja:
@hofbi Who is 'we'? You don't appear to have any authority to merge this (you're not the owner and aren't shown as a collaborator).

I added the previous test thinking that you were asking as a maintainer, I'm unwilling to act in the interests of an unknown party without knowing how it'll affect the chances of this PR finally getting merged. For all I know, @jbeder wouldn't want the test I added for you, or want this squashed.

hofbi:
@WrinklyNinja 
1. Where does @jbeder say he doesn't want this test. The suggestion for this test was the outcome of issue #617.
1. How should a Unit Test affect the outcome of this pull request. The scope of your PR is to support merge key. The test is about valid yaml with usage if merge key. Either you see the test passing which means this kind of test works or the test fails. In this case you know there is something which works not as expected.

hofbi:
@jbeder Could you review the changes again please?

jbeder:
I'm still not sure if we want this. It's not in the official spec, and I
don't like the idea of a preprocessor config.

WrinklyNinja:
> I'm still not sure if we want this. It's not in the official spec, and I don't like the idea of a preprocessor config.

As joseph-ireland noted, I was wrong in my initial PR message, as the spec does strongly recommend using merge keys, along with other types not included in the spec itself. Quoting the relevant section:

> Interoperable schemas make use of global tags (URIs) that represent the same data across different programming languages. In addition, an interoperable schema may provide additional tag resolution rules. Such rules may provide additional regular expressions, as well as consider the path to the node. This allows interoperable schemas to use untagged nodes. 
>
> It is strongly recommended that such schemas be based on the core schema defined above. In addition, it is strongly recommended that such schemas make as much use as possible of the the YAML tag repository at http://yaml.org/type/. This repository provides recommended global tags for increasing the portability of YAML documents between different applications.
>
> The tag repository is intentionally left out of the scope of this specification. This allows it to evolve to better support YAML applications.

Not supporting merge keys because they're not in the spec document itself seems to directly contradict the text.

hofbi:
Also check node anchors of the spec http://yaml.org/spec/1.2/spec.html#id2785586

timn:
We would like to use this to avoid repetitions of defaults in configurations, which can be rather annoying. I would say the cat is already out of the bag regarding merge keys, they are used heavily, e.g., for CI tool configuration (cf. [CircleCI blog entry](https://circleci.com/blog/circleci-hacks-reuse-yaml-in-your-circleci-config-with-yaml/)), and it's mentioned in [tutorials](https://learnxinyminutes.com/docs/yaml/). We also have an example in our [Buildkite pipeline configuration](https://github.com/fawkesrobotics/fawkes/blob/233ee67c43229b15a9ccd98544d144efdcf8f120/.buildkite/pipeline.yml#L8-L14), where we would need to repeat this snippet several times. Now, we would like to have the same capabilities in our YAML-based configuration implemented using yaml-cpp.

I think it would be good for yaml-cpp to merge this for compatibility with the rest what's out there in YAML-land.

noizex:
> I'm still not sure if we want this. It's not in the official spec, and I don't like the idea of a preprocessor config.

Can't this be supported with a toggle? If someone wants "pure" spec-compliant version they can compile it with such flag, and those who find merge functionality to be one of the best things about YAML file composability and removing duplication could use the feature with yaml-cpp, rather than try to find some other way to do this. As far as I know this is widely supported by libs from other languages too?

wieset:
I was surprised today to find that yaml-cpp does not support merge keys. Super useful to avoid repetitions. Very much in favor of seeing this implemented!

CZDanol:
This implementation does not work with when iterating over the node keys using for loop.

andrewerf:
It's 2021 now. Any plans on adding this feature?

perrysteger:
Would very much appreciate merge keys being supported in yaml-cpp. Took me a while to realize I was doing nothing wrong when I had a node key of "<<"

jbeder:
I think the place this PR ended up is that I don't think it should be in the mainline of yaml-cpp because it's not part of the standard YAML 1.2 spec. I'm going to close this PR so that's clear.

noizex:
Well you not thinking it should be in the mainline was the start of this years long "conversation" where people tried to convince you that having this as a switchable feature would be beneficial for the library. At least that gives us official stance on this feature and maybe some incentive to start a fork.

jbeder:
Yeah, that's totally fine. This probably would be a good feature on top of the library (like a call to Merge or something). I'm also looking for someone to take over yaml-cpp, so if that future person wants to merge this feature, that's fine too.

Issue #284, opened 2015-03-30
Implement support for timestamps


GoogleCodeExporter:
```
Hi,

I'm currently trying to incorporate the yaml-cpp in a Qt-based project.  One 
thing I need to serialise and deserialise is timestamps.  I haven't yet 
stumbled on a way to do this.  Specifically, I'm after a QDateTime, but I can 
tolerate a time_t or some generic date/time class.

The YAML way of giving these seems to be ISO8601 format, see example 2.2 here:
http://www.yaml.org/spec/1.2/spec.html#id2761292

I'm guessing the way forward is to implement custom serialisation methods.  If 
I come up with something I'll post back here.
```

Original issue reported on code.google.com by `stua...@longlandclan.yi.org` on 10 Mar 2015 at 2:40

c0yote:
I'm implementing this with `std::time_point` for my own purposes.  I'll put in a pull request when finished so you can include it in the C++11 upgrade if it's not to ugly.  This bit is causing some interesting parsing adventures:

    [...] serialized using a subset of the ISO8601 format and the formats 
    proposed by the W3C note on datetime. In addition, a more relaxed 
    format [...]

I'm almost done with it and it supports all the format examples on the [spec page](http://yaml.org/type/timestamp.html):

    canonical:        2001-12-15T02:59:43.1Z
    valid iso8601:    2001-12-14t21:59:43.10-05:00
    space separated:  2001-12-14 21:59:43.10 -5
    no time zone (Z): 2001-12-15 2:59:43.10
    date (00:00:00Z): 2002-12-14

I'll write a test as well.  Is there style guide I should follow?

jbeder:
Thanks for looking into this. The style is basically: run clang-format using the style file in the base of the repo. And just try to follow the style of nearby code.

It's basically google style. See https://google-styleguide.googlecode.com/svn/trunk/cppguide.html for more details.

PR #302, opened 2015-04-08
Support parsing of timestamp strings to C++11 time_points.

c0yote:
This is added encode & decode support for timestamps represented as C++11 type `std::chrono::time_point` which is a optional solution for the enhancement mentioned in issue [#284](https://github.com/jbeder/yaml-cpp/issues/284).

I know it's a cart \ horse type thing since the C++11 version isn't started yet, but I was building this anyway for another project.

Figured I'd just leave it here until it was relevant.


jbeder:
OK, thanks - I'll look again once yaml-cpp uses C++11.

PR #303, opened 2015-04-08
Update node_data::remove to use new equals() method

bdutro:
- Update the call to equals() in node_data::remove() to match the new implementation
- Add unit test for node::remove() to catch this type of bug in the future

### DIFF
--- a/include/yaml-cpp/node/detail/impl.h
+++ b/include/yaml-cpp/node/detail/impl.h
@@ -133,7 +133,7 @@ inline bool node_data::remove(const Key& key, shared_memory_holder pMemory) {
     return false;
 
   for (node_map::iterator it = m_map.begin(); it != m_map.end(); ++it) {
-    if (equals(*it->first, key, pMemory)) {
+    if (it->first->equals(key, pMemory)) {
       m_map.erase(it);
       return true;
     }
--- a/test/node/node_test.cpp
+++ b/test/node/node_test.cpp
@@ -39,6 +39,13 @@ TEST(NodeTest, SimpleAppendSequence) {
   EXPECT_TRUE(node.IsSequence());
 }
 
+TEST(NodeTest, MapElementRemoval) {
+  Node node;
+  node["foo"] = "bar";
+  node.remove("foo");
+  EXPECT_TRUE(!node["foo"]);
+}
+
 TEST(NodeTest, SimpleAssignSequence) {
   Node node;
   node[0] = 10;


pbeeson:
Any reason why you cannot fix this equals() call in remove() for 0.5.2, so  that Ubuntu 16.04 (once it pulls changes) would have a working yaml.cpp out of the box?

akimd:
Ping.

jbeder:
@pbeeson and @akimd  what is your question? This has been fixed at HEAD, and will be in the next release.

akimd:
Yes, sorry for being so uninformative...  I was interrupted when I was about to write.

My problem is that Debian ships 0.5.2 so my application is broken on it.  Are there any possible work around for these versions?  I'm reluctant to embed a patched version of the guilty header in my software, but...  Besides, I don't think I saw a CPP symbol that defines the version of yaml-cpp, something I could use to `#if` that include.

pbeeson:
The problem is straightforward.  0.5.2 is broken in that the header won't compile.  Current Debian and Ubuntu releases are pegged to 0.5.2.

My solution right now is either to install 0.5.3 on every machine in my building that might use yaml-cpp (we use ROS so that's that every machine) or to patch the use/include/ header for the shipped 0.5.2 release.  This is not ideal.  Perhaps Debian/Ubuntu need to promote 0.5.3 to their current stable/LTS releases, and I would have hoped the actual software architect would has some influence in that.

jbeder:
Has anyone asked Debian about this?

pbeeson:
I put in a request to the Debian package maintainer, but haven't heard back as of yet.  Given that nothing that includes the impl header from 0.5.2 will compile, it baffles me that no one has caught this sooner.

akimd:
@pbeeson I don't think the header is not _that_ broken: the function `node_data::remove` is templated, so it's only if you use it that the compilation fails.  So you and I are affected because we use it, but most users probably don't remove nodes.

pbeeson:
@akimd I suggest trying this PPA.  It works for me.

https://launchpad.net/~winterheart/+archive/ubuntu/openxcom

willstott101:
FWIW for those finding this, it looks like this has been patched for 0.5.2 by ubuntu (at least) downstream.

[changelog here](http://changelogs.ubuntu.com/changelogs/pool/universe/y/yaml-cpp/yaml-cpp_0.5.2-4ubuntu1/changelog)

Issue #150, opened 2015-03-30
Conversion operators


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. This is a feature request;
2. $ ack operator (in the root directory)
3. Look what operators exists

It would be really handy to define conversion operators for some usual cases 
(int, std::string mainly). It might even be possible to template

Thank you for the great work. Love the new API

Best regards,
Jonatan Olofsson
```

Original issue reported on code.google.com by `jonatan....@gmail.com` on 23 Jan 2012 at 9:23

SGSSGene:
This is a enhancement that is currently of scope.

PR #310, opened 2015-05-08
Added template parameter to convert struct to be able to use std::enable_if.

timohl:
I would like to use std::enable_if and std::is_base_of to write a convert struct that only converts classes derived from one base class.
Therefore I added a template parameter to the convert struct with a default value of void.
This compiles well with gcc (4.8.3 & 4.9.2) and passes all 871 tests.

I added a test application to show the usage of the additional template parameter.
This test is in test_app/main.cpp, because I have not writen tests yet and did not know where to put it in.
Maybe in test/integration/load_node_test.cpp ?
Please let me know if this is the right place and my proposed change is accaptable.
After that I will write a proper test and look at clang-format.

By the way in issue https://github.com/jbeder/yaml-cpp/issues/150 somebody wrote about inheritance in a math library, which would be solved by this change as I understand his problem.

Search for "3) I may have over generalized" in this issue, because I do not know how to link to particular comments in github.

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -23,7 +23,7 @@
 namespace YAML {
 class Binary;
 struct _Null;
-template <typename T>
+template <typename T, typename Enable>
 struct convert;
 }  // namespace YAML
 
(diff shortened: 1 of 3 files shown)


BenjaminNavarro:
I would be very interested in this feature. Why this PR was closed without being merged?

jbeder:
Good question, I have no idea. This does seem useful. If you want to open a new PR with this (and appropriate tests), please do!

Issue #305, opened 2015-04-11
Exception thrown in as method when default supplied


JonHodgson:
Perhaps I'm missing something here, but I've been successfully using an earlier version of yaml-cpp with my code containing lines such as this

    std::string group = attributes["group"].as<std::string>("");

the node "group" may not exist, but that's why there's a default.

However I just updated the yaml-cpp source, and now there is an additional check in "as"

    if (!m_isValid)
      throw InvalidNode();

So, if the node doesn't exist, rather than just returning the default value, an exception is thrown.

This doesn't make sense to me. I can understand throwing an exception when no default value is supplied, but not when it is.

jbeder:
Thanks, this is definitely a mistake.

jbeder:
This was fixed by #316, b426fafff6238dda8d86fa668f585cba732dd272.

PR #315, opened 2015-05-29
Don't throw exception in iterator or .as(fallback) on missing node

JonnyH:
This appears to be a behaviour change caused by 1025f76df1b32b6ec3571ca928d7797a768a3341 - where a missing node is no longer considered 'valid'.
This is referenced in issue #305 - but also affects a number of other paths causing changes in behaviour.

I noticed this running OpenXcom (http://github.com/SupSuper/OpenXcom) which regularly uses .as(fallback) on optional nodes, expecting the fallback if it's missing (not an exception), and iterates over optional nodes (which it expects .begin() == .end() on such missing nodes).

There are likely other behaviour changes causing by this on missing nodes (I noticed that .size() would have returned 0 instead of throwing an exception, for example) but this change request does not fix that at this time.

I have been unable to find any documentation of what is expected in these cases, but these were working in 0.5.1 and broke in 0.5.2.

### DIFF
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -282,26 +282,26 @@ inline std::size_t Node::size() const {
 
 inline const_iterator Node::begin() const {
   if (!m_isValid)
-    throw InvalidNode();
+    return const_iterator();
   return m_pNode ? const_iterator(m_pNode->begin(), m_pMemory)
                  : const_iterator();
 }
 
 inline iterator Node::begin() {
   if (!m_isValid)
-    throw InvalidNode();
+    return iterator();
   return m_pNode ? iterator(m_pNode->begin(), m_pMemory) : iterator();
 }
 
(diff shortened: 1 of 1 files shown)


jbeder:
Thanks! Can you write tests for each of these?

JonnyH:
Sure, I'll try to get a proper fix this weekend.
By the way, it seems to be a 'const &Node' behaviour change, the following passes on 0.5.1 but not 0.5.2. This might be a clue to a 'better' fix, I'll look into it when I get time.

TEST(NodeTest, UndefinedConstNodeWithFallback) {                                
  Node node;                                                                    
  const Node& cn = node;                                                        
  EXPECT_EQ(cn["undefined"].as<int>(3), 3);                                     
}

jbeder:
Cool, thanks.

JonnyH:
Closing as #316 does the same with tests

PR #316, opened 2015-06-01
0.5.2 regressions

JonnyH:
Fixes at least three regressions from 0.5.1 -> 0.5.2 with tests - all introduced in 1025f76df1b32b6ec3571ca928d7797a768a3341 and relating to 'undefined' const Node& objects returned from a operator[] with no matching node.

- .as(fallback) returns the fallback value, 0.5.2 throws an exception (issue #305)
- const_iterator .begin()/.end() returned const_iterator() in 0.5.1 but throws an exception in 0.5.2
- iterator .begin()/.end() returned const_iterator() in 0.5.1 but throws an exception in 0.5.2

This should replace by previous pull request #315

### DIFF
--- a/test/node/node_test.cpp
+++ b/test/node/node_test.cpp
@@ -80,6 +80,12 @@ TEST(NodeTest, MapWithUndefinedValues) {
   EXPECT_EQ(2, node.size());
 }
 
+TEST(NodeTest, UndefinedConstNodeWithFallback) {
+  Node node;
+  const Node& cn = node;
+  EXPECT_EQ(cn["undefined"].as<int>(3), 3);
+}
+
 TEST(NodeTest, MapIteratorWithUndefinedValues) {
   Node node;
   node["key"] = "value";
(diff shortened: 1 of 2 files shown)


test/node/node_test.cpp
```diff
@@ -91,6 +97,30 @@ TEST(NodeTest, MapIteratorWithUndefinedValues) {
   EXPECT_EQ(1, count);
 }
 
+TEST(NodeTest, ConstIteratorOnConstUndefinedNode) {
+  Node node;
+  const Node& cn = node;
+  const Node& undefinedCn = cn["undefined"];
+
+  std::size_t count = 0;
+  for (const_iterator it = undefinedCn.begin(); it != undefinedCn.end(); ++it)
```
jbeder: Add curly braces to this "if" statement. (I haven't changed all existing ones, but new code should put curly braces around all control flow blocks, even one-liners.)

test/node/node_test.cpp
```diff
@@ -91,6 +97,30 @@ TEST(NodeTest, MapIteratorWithUndefinedValues) {
...
+
+TEST(NodeTest, IteratorOnConstUndefinedNode) {
+  Node node;
+  const Node& cn = node;
+  const Node& undefinedCn = cn["undefined"];
+
+  Node& nonConstUndefinedNode = const_cast<Node&>(undefinedCn);
+
+  std::size_t count = 0;
+  for (iterator it = nonConstUndefinedNode.begin(); it != nonConstUndefinedNode.end(); ++it)
```
jbeder: ditto

test/node/node_test.cpp
```diff
@@ -91,6 +97,30 @@ TEST(NodeTest, MapIteratorWithUndefinedValues) {
...
+    count++;
+  EXPECT_EQ(0, count);
+}
+
+TEST(NodeTest, IteratorOnConstUndefinedNode) {
+  Node node;
+  const Node& cn = node;
+  const Node& undefinedCn = cn["undefined"];
+
+  Node& nonConstUndefinedNode = const_cast<Node&>(undefinedCn);
```
jbeder: Does this need to go through a const node for the appropriate behavior to be tested?

jbeder:
Thanks for doing this. Can you squash these down into one commit? Feel free to wait until after the review before doing so.

test/node/node_test.cpp
JonnyH: I believe so - as it's only "const Node Node::operator[](const Key& key) const" that returns a 'zombie' node - "Node Node::operator[](const Key& key)", however, will always create a new node, with a NULL node* - which does not hit the if(!m_isValid) exception.

Then , as far as I know, the const_cast<> is required to get a non-const iterator out of the undefined 'zombie' node returned from the const operator[] version. (Though I see it as if something breaks because of something you can only get access to by const_cast<>ing a value to non-const it's kinda your own fault - I only did the same for the non-const interator functions for completeness)

jbeder: Yeah, that makes sense. It's a little weird, as you point out, but we ought to support this behavior. SGTM.

jbeder:
Looks good overall. Please squash and I'll pull.

winterheart:
Eh, ping?

jbeder:
Sorry, I didn't get a notification when the updated commit was posted.

@JonnyH , can you rebase against HEAD?

JonnyH:
Rebased on top of b0a4de3dd9aa4841f3747d70156e8c206bdc4924

PR #330, opened 2015-08-13
Allow using a Node as the key in force_insert.

md5i:
Node::force_insert() uses convert<> to convert its key to a node.
Add a specialization for convert<Node>.


jbeder:
Cherry-picked into master as 320b02b14ac9eed2dcafbe6d7d43a0526d3bd0d1.

PR #331, opened 2015-08-13
auto_ptr to unique_ptr

amikhalev:
Changed all uses of auto_ptr to unique_ptr


jbeder:
Thanks for this, but there's another pull request to migrate many features to C++11, and when I get to it, I'll just go that route.

PR #341, opened 2015-11-09
Fixed warining of C4819 under MSVC 14.0

kcauchy:
```
[ 53%] Building CXX object test/gmock-1.7.0/CMakeFiles/gmock.dir/gtest/src/gtest-all.cc.obj
gtest-all.cc
[ 55%] Building CXX object test/gmock-1.7.0/CMakeFiles/gmock.dir/src/gmock-all.cc.obj
gmock-all.cc
C:\cygwin64\home\karl\ports\jbeder\yaml\yaml\test\gmock-1.7.0\src/gmock-matchers.cc: error C2220: 警告被视为错误 - 没有生成“object”文件
C:\cygwin64\home\karl\ports\jbeder\yaml\yaml\test\gmock-1.7.0\src/gmock-matchers.cc: warning C4819: 该文件包含不能在当前代码页(936)中表示的字符。请将该文件保存为 Unicode 格式以防止数据丢失
NMAKE : fatal error U1077: “C:\PROGRA~2\MICROS~1.0\VC\bin\cl.exe”: 返回代码“0x2”
```
It seems both master and  version 0.5.2 have the problem of C2220 and C4819 under Visual Studio 2015.
And also, it is a boost problem as well. (Boost 1.5.9) (And you used a very strict standard)

### DIFF
--- a/test/gmock-1.7.0/src/gmock-matchers.cc
+++ b/test/gmock-1.7.0/src/gmock-matchers.cc
@@ -1,4 +1,4 @@
-// Copyright 2007, Google Inc.
+﻿// Copyright 2007, Google Inc.
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without


jbeder:
I'm not sure what you're changing - the diff doesn't make that clear.

Also, this is just a change to gmock. Should you instead file a bug against them?

kcauchy:
It's a problem compiling under Visual Studio 2015. It is because some characters were stored in other character set, not "Unicode", then the visual studio complains about "You should use 'unicode'". 
(Warning C4819.)

Normally it is not a problem, but because gmock uses CMake's "cxx_strict“ standard, the warning becomes an error, and make the system can't be compiled.

I have tried a lot of methods to fix the problem, but the easiest one is just open this file in notepad, and save as a "UTF-8" copy. That's all I have done.

jbeder:
So are you suggesting that a gmock file is stored in a non-UTF-8 format?

kcauchy:
I am afraid yes. I tried to use Git 2.6.2, and downloaded the zip file directly from github, both files the same.

jbeder:
Can you file a bug against gmock and post the link here?

PR #343, opened 2015-11-25
Added CMake config scripts to installation

theNerd247:
Added generated cmake configuration scripts to the files to be installed. These scripts will be installed to `.../lib/cmake/yaml-cpp`.


paulnovo:
I just noticed this pull request, after I put up my own to solve the same issue. Oops. Anyway, I don't think your solution works correctly, it will install the cmake scripts that point back to the files in the Build directory, not their installed locations (library and includes). That is why in the pull request I put up, it actually makes a second yaml-cpp-config.cmake, hides it under ${PROJECT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}, and then installs it during installation.

theNerd247:
Oops! Thanks for pointing that out. I'm building from source on my system and I tend to keep the source code for documentation purposes (especially for this project as there is hardly any) so I didn't realize the bad paths (as it is working on my system). I'll close this pull request.

Issue #158, opened 2015-03-30
performance


GoogleCodeExporter:
```
Maybe some wrongs with my codes, the codes is following:
the code of parsing this attached file in C++:
    std::ifstream file("test.yaml");
    YAML::Parser parser(file);
    YAML::Node document;
    parser.GetNextDocument(document);

the code of parsing this attached file in C:
    file = fopen("test.yaml", "rb");
    yaml_parser_initialize(&parser);
    yaml_parser_set_input_file(&parser, file);
    while (!done)
    {
        if (!yaml_parser_load(&parser, &document)) {
            error = 1;
            break;
        }
        done = (!yaml_document_get_root_node(&document));
        yaml_document_delete(&document);

        if (!done) count ++;
    }
    fclose(file);

```

Original issue reported on code.google.com by `cmt75...@gmail.com` on 9 Apr 2012 at 3:20

Attachments:
- [test.yaml](https://storage.googleapis.com/google-code-attachments/yaml-cpp/issue-158/comment-0/test.yaml)

SGSSGene:
If performance issues exists, please open a new ticket

PR #354, opened 2016-01-12
Cache regexes for scalars -- improves parsing performance

swolchok:
I noticed a lot of time in the scanner was spent futzing with `RegEx`es, so I cached the specific ones that the scanner uses for tokens instead of re-constructing them each time. The two commits combined seem to improve parsing performance by nearly 2x on the sample attached to #158.

### DIFF
--- a/src/scanscalar.h
+++ b/src/scanscalar.h
@@ -19,7 +19,8 @@ enum FOLD { DONT_FOLD, FOLD_BLOCK, FOLD_FLOW };
 
 struct ScanScalarParams {
   ScanScalarParams()
-      : eatEnd(false),
+      : end(nullptr),
+        eatEnd(false),
         indent(0),
         detectIndent(false),
         eatLeadingWhitespace(0),
@@ -32,7 +33,8 @@ struct ScanScalarParams {
         leadingSpaces(false) {}
 
   // input:
-  RegEx end;          // what condition ends this scalar?
+  const RegEx* end;   // what condition ends this scalar?
+                      // unowned.
   bool eatEnd;        // should we eat that condition when we see it?
   int indent;         // what level of indentation should be eaten and ignored?
   bool detectIndent;  // should we try to autodetect the indent?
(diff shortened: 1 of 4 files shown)


src/scanscalar.h
```diff
@@ -19,7 +19,8 @@ enum FOLD { DONT_FOLD, FOLD_BLOCK, FOLD_FLOW };
 
 struct ScanScalarParams {
   ScanScalarParams()
-      : eatEnd(false),
+      : end(0),
```
jbeder: initialize to nullptr

src/scanscalar.h
```diff
@@ -32,7 +33,8 @@ struct ScanScalarParams {
         leadingSpaces(false) {}
 
   // input:
-  RegEx end;          // what condition ends this scalar?
+  const RegEx* end;   // what condition ends this scalar?
+                      // NOTE: does not have ownership and must not be freed.
```
jbeder: Comment can simply be "unowned" - no need for the detailed info

jbeder:
Nice find! I made a few comments above. I didn't get a chance to test this performance but I will shortly.

src/scanscalar.h
swolchok: Oh, didn't realize that C++11 was allowed in yaml-cpp already. Will do!

swolchok:
Anything I can do to help with performance testing?

jbeder:
Sorry, some personal stuff came up so I haven't been able to look at this. Hopefully next week/weekend.

swolchok:
Thanks for the update. Hope your stuff gets resolved smoothly, no rush on my end!

jbeder:
Awesome, pulled as 8c35a8ffabea1c989d9f449bff053081fdc987e1 and 005a6a19ee0af3e284637abe9ecea5501993fc4c (again, cherry-picked).

I added some repetition to the "read" binary (9e37409b4b40e4ca5b026ea9de7ea9784da80ff2) and running

`time util/read -n 100 -c test.yaml`

Before:

```
real 0m26.006s
user 0m25.987s
sys  0m0.010s
```

After:

```
real 0m5.331s
user 0m5.327s
sys  0m0.003s
```

5x improvement!

PR #355, opened 2016-01-13
read benchmark: accept a filename as an argument

swolchok:
On my Macbook Pro, reading from standard input incurs a bunch of locking
overhead, which complicates profiling and (IMO) adds noise to
results. This adds the option to read from a file, which doesn't incur
this overhead.

### DIFF
--- a/util/read.cpp
+++ b/util/read.cpp
@@ -1,3 +1,4 @@
+#include <fstream>
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


jbeder:
Pulled as a5b72f7ae69e5a28434fca02faba6f48ca0d09a1.

github seems to get confused when you cherry-pick

PR #356, opened 2016-01-13
add some tests for RegEx

swolchok:
Just some unit tests to catch regressions.

### DIFF
--- a/test/regex_test.cpp
+++ b/test/regex_test.cpp
@@ -1,0 +1,177 @@
+#include "gtest/gtest.h"
+#include "regex_yaml.h"
+#include "stream.h"
+
+using YAML::RegEx;
+using YAML::Stream;
+
+namespace {
+const auto MIN_CHAR = Stream::eof() + 1;
+
+TEST(RegExTest, Empty) {
+  RegEx empty;
+  EXPECT_TRUE(empty.Matches(std::string()));
+  EXPECT_EQ(0, empty.Match(std::string()));
+  for (int i = MIN_CHAR; i < 128; ++i) {
+    auto str = std::string(1, char(i));
+    EXPECT_FALSE(empty.Matches(str));
+    EXPECT_EQ(-1, empty.Match(str));
+  }
+}
+
+TEST(RegExTest, Range) {
+  for (int i = MIN_CHAR; i < 128; ++i) {
+    for (int j = MIN_CHAR; j < 128; ++j) {
+      RegEx ex((char)i, (char)j);
+      for (int k = MIN_CHAR; k < 128; ++k) {
+        auto str = std::string(1, char(k));
+        if (i <= k && k <= j) {
+          EXPECT_TRUE(ex.Matches(str));
+          EXPECT_EQ(1, ex.Match(str));
+        } else {
+          EXPECT_FALSE(ex.Matches(str));
+          EXPECT_EQ(-1, ex.Match(str));
+        }
+      }
+    }
+  }
+}
+
+TEST(RegExTest, EmptyString) {
+  RegEx ex = RegEx(std::string());
+  EXPECT_TRUE(ex.Matches(std::string()));
+  EXPECT_EQ(0, ex.Match(std::string()));
+
+  // Matches anything, unlike RegEx()!
+  EXPECT_TRUE(ex.Matches(std::string("hello")));
+  EXPECT_EQ(0, ex.Match(std::string("hello")));
+}
+
+TEST(RegExTest, SingleCharacterString) {
+  for (int i = MIN_CHAR; i < 128; ++i) {
+    RegEx ex(std::string(1, (char)i));
+    for (int j = MIN_CHAR; j < 128; ++j) {
+      auto str = std::string(1, char(j));
+      if (j == i) {
+        EXPECT_TRUE(ex.Matches(str));
+        EXPECT_EQ(1, ex.Match(str));
+        // Match at start of string only!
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ file(GLOB test_new_api_sources new-api/[a-z]*.cpp)
 list(APPEND test_sources ${test_new_api_sources})
 add_sources(${test_sources} ${test_headers})
 
-include_directories(${YAML_CPP_SOURCE_DIR}/test)
+include_directories(${YAML_CPP_SOURCE_DIR}/test ${YAML_CPP_SOURCE_DIR}/src)
```
jbeder: Is this change necessary?

test/regex_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+using YAML::RegEx;
+using YAML::Stream;
+
+namespace {
+TEST(RegExTest, Empty) {
+  RegEx empty;
+  EXPECT_TRUE(empty.Matches(std::string()));
+  EXPECT_EQ(0, empty.Match(std::string()));
+  for (int ii = Stream::eof() + 1; ii < 128; ++ii) {
```
jbeder: 1) I'd prefer "i" to "ii"

2) Starting with Stream::eof() + 1 is probably not necessary. I'd hard-code the constant here. If you prefer, feel free to make a named constant at the top of this file (and if you do that, do one for the 128 as well.

test/regex_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    EXPECT_FALSE(empty.Matches(char(ii)));
```
jbeder: Rather than casting to char, what do you think about making the loop variable a char? You won't be able to have < as the terminus of the loop, but maybe this way you could make, e.g., a constexpr vector<char> at the top of this file that consists of all of the chars that you're trying to test, and then just loop through that.

test/regex_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+      } else {
+        EXPECT_FALSE(ex.Matches(char(jj)));
+        EXPECT_EQ(-1, ex.Match(char(jj)));
+      }
+    }
+  }
+}
+
+TEST(RegExTest, Range) {
+  for (int ii = Stream::eof() + 1; ii < 127; ++ii) {
```
jbeder: Do you mean to ensure that ii < jj? If so, just make the second loop go from ii + 1 to 128.

test/regex_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+          EXPECT_FALSE(ex.Matches(char(kk)));
+          EXPECT_EQ(-1, ex.Match(char(kk)));
+        }
+      }
+    }
+  }
+}
+
+TEST(RegExTest, EmptyString) {
+  RegEx ex = RegEx(std::string());
```
jbeder: "" rather than std::string()

test/regex_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+  EXPECT_TRUE(ex.Matches(std::string()));
+  EXPECT_EQ(0, ex.Match(std::string()));
+
+  // Matches anything, unlike RegEx()!
+  EXPECT_TRUE(ex.Matches(std::string("hello")));
+  EXPECT_EQ(0, ex.Match(std::string("hello")));
+}
+
+TEST(RegExTest, SingleCharacterString) {
```
jbeder: If you're testing this, what's the point in testing the SingleCharacter, given that you even had to add to the public API in order to test individual characters?

test/regex_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include "regex_yaml.h"
+#include "stream.h"
+
+using YAML::RegEx;
+using YAML::Stream;
+
+namespace {
+TEST(RegExTest, Empty) {
+  RegEx empty;
+  EXPECT_TRUE(empty.Matches(std::string()));
```
jbeder: "" instead of std::string()

test/regex_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+  }
+}
+
+TEST(RegExTest, EmptyString) {
+  RegEx ex = RegEx(std::string());
+  EXPECT_TRUE(ex.Matches(std::string()));
+  EXPECT_EQ(0, ex.Match(std::string()));
+
+  // Matches anything, unlike RegEx()!
+  EXPECT_TRUE(ex.Matches(std::string("hello")));
```
jbeder: And throughout; why not just the string literals?

jbeder:
This is a great addition, thanks!

test/regex_test.cpp
swolchok: I intended to test j > i as well. I figured testing everything is easy and cheap.

test/regex_test.cpp
swolchok: I originally intended to make some changes to the RegEx implementation that would have made this make sense, but they didn't pan out and you're right that it doesn't make sense as is.

swolchok:
Not sure if you deleted the comments or I just can't find them, but I see in my email that you asked why I call the `std::string` ctor explicitly. It's because there are templated versions of the ctor and Match that are a better match for `const char *` than the `std::string` versions are.

jbeder:
Cherry-picked as 0a1352525aa4decefa6cc5e8ae9f0962175f5753.

Issue #289, opened 2015-03-30
does not parse null value


GoogleCodeExporter:
```
Emitter correctly emits "~" for emtpy/null node but parser loads "~" as a 
string.
Tested versions 0.5.1, trunk.
```

Original issue reported on code.google.com by `sir...@dasir.cz` on 27 Mar 2015 at 9:50

jbeder:
Fixed by 52bcefa1f18947bac98288b00188551046edb948, test added in bfb5703ce92b18e1f2d974a7c567a003184b64b0.

Issue #230, opened 2015-03-30
Null is expected to output empty string, instead of tilde, so that it is loadable to Null.


GoogleCodeExporter:
```
What steps will reproduce the problem?
  YAML::Node node = YAML::Load("---\n...");
  std::ostream << node << std::endl;


What is the expected output? What do you see instead?
an empty string is expected.
currently shows a tilde(~),
which is loaded to a string of "~", instead of Null.

What version of the product are you using? On what operating system?
The latest (on 2013/12/12) revision at the hg repository,
Mac OS X Mavericks.

Please provide any additional information below.
http://stackoverflow.com/questions/20544174/how-to-specify-the-output-format-for
-null-node-in-yaml-cpp
```

Original issue reported on code.google.com by `inouetak...@gmail.com` on 14 Dec 2013 at 1:06

jbeder:
Fixed by 52bcefa1f18947bac98288b00188551046edb948, test added in bfb5703ce92b18e1f2d974a7c567a003184b64b0.

PR #357, opened 2016-01-16
Made null handling YAML 1.2 compliant

TripleWhy:
See http://www.yaml.org/spec/1.2/spec.html#id2805071
Also fixes Bugs #289 and #230 (I think).

### DIFF
--- a/src/emitterutils.cpp
+++ b/src/emitterutils.cpp
@@ -9,6 +9,7 @@
 #include "stringsource.h"
 #include "yaml-cpp/binary.h"  // IWYU pragma: keep
 #include "yaml-cpp/ostream_wrapper.h"
+#include "yaml-cpp/null.h"
 
 namespace YAML {
 namespace Utils {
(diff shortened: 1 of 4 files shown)


src/emitterutils.cpp
```diff
@@ -157,7 +157,7 @@ bool IsValidPlainScalar(const std::string& str, FlowType::value flowType,
   }
 
   // check against null
-  if (str == "null") {
+  if (str == "~" || str == "null" || str == "Null" || str == "NULL") {
```
jbeder: Can you extract these two places into a separate function? Perhaps put it in null.h / null.cpp.

TripleWhy:
As you wish.

src/null.cpp
```diff
@@ -2,4 +2,8 @@
 
 namespace YAML {
 _Null Null;
+
+bool IsNullString(const std::string& str) {
+  return (str.empty() || str == "~" || str == "null" || str == "Null" || str == "NULL");
```
jbeder: You don't need the extra parens around the return value.

jbeder:
Sorry for the delay. Please squash and rebase against HEAD.

TripleWhy:
Done

PR #358, opened 2016-01-22
Useless forward declaration

Yankes:
It should be `YAML::Node` or `YAML::detail::node`. Because nobody noticed it, it should be deleted.


jbeder:
I pulled this as dfbb388409fcfb9f6d2ef07cb0ef5bacadd08ca7 (the CL description needed tweaking, and it was easier just to make a new commit.).

Thanks!

PR #360, opened 2016-01-26
Removed quoted variables to avoid CMP0054 policy warnings on CMake 3

Kwizatz:
This is just a small change to silence warnings regarding cmake policy 54 on cmake 3.3.0.
I have verified that the change does not affect the build process when running cmake  2.8.12.2 as shipped with Ubuntu Linux LTS 14.04 and does what the conditional is expected to do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,8 +148,8 @@ if(WIN32)
 endif()
 
 # GCC or Clang specialities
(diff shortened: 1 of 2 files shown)


jbeder:
Cherry-picked as 57805dfd6a741e55477ea8d4d5b3b6f0272d1f0e.

PR #365, opened 2016-03-10
for node_map changing map<*,*> to list<pair<*,*>>

ysyrov:
It's pointless to sort by pointer value, plus it shuffles files contents (after read-write).
Please pull this into master.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,8 +148,8 @@ if(WIN32)
 endif()
 
 # GCC or Clang specialities
-if("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" OR
-   "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR
+   CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 	### General stuff
 	if(WIN32)
 		set(CMAKE_SHARED_LIBRARY_PREFIX "")	# DLLs do not have a "lib" prefix
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -8,8 +8,8 @@ if(WIN32 AND BUILD_SHARED_LIBS)
   add_definitions("-DGTEST_LINKED_AS_SHARED_LIBRARY")
 endif()
 
-if("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" OR
-   "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR
+   CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set(yaml_test_flags "-Wno-c99-extensions -Wno-variadic-macros -Wno-sign-compare -std=c++11")
 endif()
 


src/node_data.cpp
```diff
@@ -256,7 +256,7 @@ void node_data::reset_map() {
 }
 
 void node_data::insert_map_pair(node& key, node& value) {
-  m_map[&key] = &value;
+  m_map.push_back({ &key, &value });
```
ysyrov: agree, or, to be consistent with line below it:
push_back(kv_pair(&key, &value));

jbeder: Please go with emplace_back, and update the nearby line to be consistent.

jbeder:
Please change the commit message to be in imperative present tense (e.g., "Change ...")

ysyrov:
I disagree with commit message to be in imperative.
It's a description to an artifact and should reflect what it is, not what someone else needs to do.
;)

include/yaml-cpp/node/detail/node_data.h
```diff
@@ -114,7 +114,7 @@ class YAML_CPP_API node_data {
   mutable std::size_t m_seqSize;
 
   // map
-  typedef std::map<node*, node*> node_map;
+  typedef std::list< std::pair<node*, node*> > node_map;
```
jbeder: Make sure you run this change through clang-format. You shouldn't need the extra spaces around the < and >

jbeder:
Can you squash these commits, and also capitalize the first letter and put a period at the end of the sentence? Also, please put a little explanation of *why* you're changing this in the commit message.

Btw, check out http://git.kernel.org/cgit/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD for some discussion of the tense, e.g.

> Describe your changes in imperative mood, e.g. "make xyzzy do frotz"
> instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy
> to do frotz", as if you are giving orders to the codebase to change
> its behaviour.

jbeder:
Also, make sure you rebase to HEAD, of course.

ysyrov:
Ok, it makes sense when change is not an item, but action ("patch" can be both, BTW). Then "change" is imperative verb. Arguing semantics, aren't we?
How about:

Change node_map type from map<*,*> to list<pair<*,*>>.

ysyrov:
the cb26bbf looks good

jbeder:
You've got 16 commits in this pull request. Just put one in and rebase against HEAD.

include/yaml-cpp/node/detail/node_data.h
```diff
@@ -114,7 +114,7 @@ class YAML_CPP_API node_data {
   mutable std::size_t m_seqSize;
 
   // map
-  typedef std::map<node*, node*> node_map;
+  typedef std::list<std::pair<node*, node*> > node_map;
```
jbeder: Extra space between > >

include/yaml-cpp/node/detail/node_iterator.h
```diff
@@ -37,7 +37,7 @@ struct node_iterator_value : public std::pair<V*, V*> {
 };
 
 typedef std::vector<node*> node_seq;
-typedef std::map<node*, node*> node_map;
+typedef std::list<std::pair<node*, node*> > node_map;
```
jbeder: ditto

src/node_data.cpp
```diff
@@ -256,9 +256,9 @@ void node_data::reset_map() {
 }
 
 void node_data::insert_map_pair(node& key, node& value) {
-  m_map[&key] = &value;
+  m_map.emplace_back(kv_pair(&key, &value));
```
jbeder: You can just emplace_back(&key, &value), right?

include/yaml-cpp/node/detail/node_data.h
ysyrov: that's clang-format

src/node_data.cpp
ysyrov: yes, all compiles to same thing.
I personally prefer {}

include/yaml-cpp/node/detail/node_data.h
jbeder: My bad, I didn't make sure the style used C++ 11. Please try again at HEAD, it should adjust it to >>

src/node_data.cpp
jbeder: I don't understand what you're saying

```
emplace_back(kv_pair(&key, &value))
```

and

```
emplace_back(&key, &value)
```

don't compile to the same thing. The former uses pair's copy-constructor and the latter uses pair's two-arg constructor.

PR #369, opened 2016-03-30
Add missing include for minimalists

ishitatsuyuki:
This is one case:
```
#include <yaml-cpp/node/node.h>
#include <yaml-cpp/node/iterator.h>
for(const YAML::Node &b : a) {...}
```


jbeder:
I merged this too eagerly; it actually broke the build. Reverted in 883a1e8540268d31e6327b4e865b87d84c328ea0.

ishitatsuyuki:
Seems I need to add more. Will create a pull request after finished.

Current situation is possibly forward declaration. It sometimes cause link problems.

PR #371, opened 2016-04-06
Change node_map type from map<ptr,ptr> to list<pair<ptr,ptr>>.

ysyrov:


jbeder:
This appears to fail the NodeForceInsert test; I could be convinced that Node::force_insert(Node, V) should have a different behavior than is tested, but I'm not sure. What do you think?

ysyrov:
yes, 
node_data::insert_map_pair
is bogus, it fails to test presence of the pair...
list is not a map, my mistake. I will correct the change

PR #377, opened 2016-05-12
only add -std=c++11 for clang and gcc

billhoffman:
Not sure if  I should have used yaml_cxx_flags instead, but the current code will add -std=c++11 to the MS compiler which does not know that flag.

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -1,14 +1,18 @@
 add_sources(parse.cpp)
 add_executable(parse parse.cpp)
 target_link_libraries(parse yaml-cpp)
-set_target_properties(parse PROPERTIES COMPILE_FLAGS "-std=c++11")
 
 add_sources(sandbox.cpp)
 add_executable(sandbox sandbox.cpp)
 target_link_libraries(sandbox yaml-cpp)
-set_target_properties(sandbox PROPERTIES COMPILE_FLAGS "-std=c++11")
 
 add_sources(read.cpp)
 add_executable(read read.cpp)
 target_link_libraries(read yaml-cpp)
-set_target_properties(read PROPERTIES COMPILE_FLAGS "-std=c++11")
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR
+    CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set_target_properties(parse PROPERTIES COMPILE_FLAGS "-std=c++11")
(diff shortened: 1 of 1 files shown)


jbeder:
I think using yaml_cxx_flags is probably right here. Can you update it?

winterheart:
I think, there better to use 
```
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
But this snippet requires CMake 3.1 (or even 3.2).

billhoffman:
yaml_cxx_flags will add some other stuff as well:
-Wall ${GCC_EXTRA_OPTIONS} -pedantic -Wno-long-long -std=c++11 ${yaml_cxx_flags}

util/CMakeLists.txt
```diff
@@ -1,14 +1,18 @@
...
 target_link_libraries(sandbox yaml-cpp)
-set_target_properties(sandbox PROPERTIES COMPILE_FLAGS "-std=c++11")
 
 add_sources(read.cpp)
 add_executable(read read.cpp)
 target_link_libraries(read yaml-cpp)
-set_target_properties(read PROPERTIES COMPILE_FLAGS "-std=c++11")
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR
+    CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
breznak: It's usually suggested to leave the door open for the future, so rather than whitelisting some, explicitly blacklist the broken. So something like 
`If (NOT compiler MATCHES "MSVC") ...`

billhoffman:
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
is the best way to go.  The flags for this can vary per compiler.  CMake now figures out the flags for you.

breznak:
>  CMake now figures out the flags for you. 

@billhoffman sounds perfect, can you update the PR please? 
What will happen if the compiler does not support the requested standard?

billhoffman:
Hmm.  This might be an issue: https://github.com/jbeder/yaml-cpp/blob/master/CMakeLists.txt#L4  I am not sure what OSX has to do with needing cmake 2.6, but cmake 2.6 will not have the c++-11 stuff in it for sure.

breznak:
cmake 2.6 is quite ancient, it's up to @jbeder  to consider bumping the minimum required version. 
OR, depending on the found cmake version, can you make some `if` that will enable the c++11stuff for the newer cmake?

PR #378, opened 2016-05-17
Fix some clang warnings

xpac27:
Those warnings appear when compiling a C++14 project that includes yaml-cpp/yaml.h


xpac27:
The travis error seems to be unrelated to those changes.

PR #379, opened 2016-05-17
Change node_map type from map<ptr,ptr> to list<pair<ptr,ptr>>.

ysyrov:

### DIFF
--- a/include/yaml-cpp/node/detail/node_data.h
+++ b/include/yaml-cpp/node/detail/node_data.h
@@ -114,7 +114,7 @@ class YAML_CPP_API node_data {
   mutable std::size_t m_seqSize;
 
   // map
-  typedef std::map<node*, node*> node_map;
+  typedef std::list<std::pair<node*, node*>> node_map;
   node_map m_map;
 
   typedef std::pair<node*, node*> kv_pair;
--- a/src/node_data.cpp
+++ b/src/node_data.cpp
@@ -256,9 +256,20 @@ void node_data::reset_map() {
 }
 
 void node_data::insert_map_pair(node& key, node& value) {
-  m_map[&key] = &value;
+
+  bool found = false;
+  for (node_map::iterator it = m_map.begin(); it != m_map.end(); ++it) {
+    if (it->first == &key) {
+      it->second = &value;
+      found = true;
+      break;
+    }
+  }
+  if (!found)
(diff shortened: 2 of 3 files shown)


ysyrov:
please stop crazy shuffling of maps!!!

include/yaml-cpp/node/detail/node_data.h
```diff
@@ -114,7 +114,7 @@ class YAML_CPP_API node_data {
   mutable std::size_t m_seqSize;
 
   // map
-  typedef std::map<node*, node*> node_map;
+  typedef std::list<std::pair<node*, node*>> node_map;
```
jbeder: Prefer `std::vector` to `std::list`

src/node_data.cpp
```diff
@@ -256,9 +256,20 @@ void node_data::reset_map() {
 }
 
 void node_data::insert_map_pair(node& key, node& value) {
-  m_map[&key] = &value;
+
+  bool found = false;
+  for (node_map::iterator it = m_map.begin(); it != m_map.end(); ++it) {
```
jbeder: Use the syntax `for (auto& kv : m_map)`

src/node_data.cpp
```diff
@@ -256,9 +256,20 @@ void node_data::reset_map() {
...
+    if (it->first == &key) {
+      it->second = &value;
+      found = true;
+      break;
+    }
+  }
+  if (!found)
```
jbeder: curly braces around the "if" block

jbeder:
Looks good, just a couple style nits.

c0nk:
Please merge this. The non-deterministic map order is quite bad. Can you just fix the style nits yourself?

jbeder:
@c0nk FYI YAML doesn't prescribe an ordering to maps. So this is entirely extra-spec.

And I figured I'd give @ysyrov a chance to do update this PR. If you're in a hurry, feel free to submit a new PR.

c0nk:
I think std::vector would also be better here for performance. But std::list has more similar iterator invalidation semantics to std::map. Is that important here?

jbeder:
The underlying container is an implementation detail, so all that matters is how it affects the external API.

c0nk:
I made a new pull request #386.

asmaloney:
Looks like this can be closed - superseded by #386

PR #384, opened 2016-06-07
Add array converter

xpac27:
Code review much welcomed 😃

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -247,7 +247,7 @@ template <typename T, std::size_t N>
 struct convert<std::array<T, N>> {
   static Node encode(const std::array<T, N>& rhs) {
     Node node(NodeType::Sequence);
-    for (auto element : rhs) {
+    for (const auto& element : rhs) {
       node.push_back(element);
     }
     return node;
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -241,6 +241,35 @@ struct convert<std::list<T> > {
   }
 };
 
+// std::array
+template <typename T, std::size_t S>
+struct convert<std::array<T, S> > {
+  static Node encode(const std::array<T, S>& rhs) {
+    Node node(NodeType::Sequence);
+    for (typename std::array<T, S>::const_iterator it = rhs.begin();
+        it != rhs.end(); ++it)
+      node.push_back(*it);
+    return node;
+  }
+
+  static bool decode(const Node& node, std::array<T, S>& rhs) {
+    if (!node.IsSequence())
+      return false;
+
+    std::size_t index = 0;
+    for (const_iterator it = node.begin(); it != node.end(); ++it) {
+#if defined(__GNUC__) && __GNUC__ < 4
+      // workaround for GCC 3:
+      rhs.at(index) = it->template as<T>();
+#else
+      rhs.at(index) = it->as<T>();
+#endif
+      ++index;
+    }
+    return true;
+  }
+};
+
 // std::pair
 template <typename T, typename U>
 struct convert<std::pair<T, U> > {
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -254,10 +254,9 @@ struct convert<std::array<T, N>> {
   }
 
   static bool decode(const Node& node, std::array<T, N>& rhs) {
-    if (!node.IsSequence())
+    if (!isNodeValid(node)) {
       return false;
-    if (node.size() != N)
-        return false;
+    }
 
     for (auto i = 0u; i < node.size(); ++i) {
 #if defined(__GNUC__) && __GNUC__ < 4
@@ -268,6 +267,12 @@ struct convert<std::array<T, N>> {
 #endif
     }
     return true;
+  }
+
+private:
+
+  static bool isNodeValid(const Node& node) {
+    return node.IsSequence() && node.size() == N;
   }
 };
 
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -7,12 +7,12 @@
 #pragma once
 #endif
 
+#include <array>
 #include <limits>
 #include <list>
 #include <map>
 #include <sstream>
 #include <vector>
-#include <array>
 
 #include "yaml-cpp/binary.h"
 #include "yaml-cpp/node/impl.h"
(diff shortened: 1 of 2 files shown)


include/yaml-cpp/node/convert.h
```diff
@@ -12,6 +12,7 @@
 #include <map>
 #include <sstream>
 #include <vector>
+#include <array>
```
jbeder: Please alphabetize standard imports.

include/yaml-cpp/node/convert.h
```diff
@@ -241,6 +242,35 @@ struct convert<std::list<T> > {
   }
 };
 
+// std::array
+template <typename T, std::size_t S>
+struct convert<std::array<T, S> > {
```
jbeder: No need for space between > >

include/yaml-cpp/node/convert.h
```diff
@@ -241,6 +242,35 @@ struct convert<std::list<T> > {
...
+  static Node encode(const std::array<T, S>& rhs) {
+    Node node(NodeType::Sequence);
+    for (typename std::array<T, S>::const_iterator it = rhs.begin();
```
jbeder: Use for ( : )

include/yaml-cpp/node/convert.h
```diff
@@ -241,6 +242,35 @@ struct convert<std::list<T> > {
   }
 };
 
+// std::array
+template <typename T, std::size_t S>
```
jbeder: I'd use N for the length, rather than S, so it looks more like a number rather than a type.

include/yaml-cpp/node/convert.h
```diff
@@ -241,6 +242,35 @@ struct convert<std::list<T> > {
...
+      node.push_back(*it);
+    return node;
+  }
+
+  static bool decode(const Node& node, std::array<T, S>& rhs) {
+    if (!node.IsSequence())
+      return false;
+
+    std::size_t index = 0;
+    for (const_iterator it = node.begin(); it != node.end(); ++it) {
```
jbeder: I think it'd be more readable to iterate through the sequence with an index, rather than iterator.

include/yaml-cpp/node/convert.h
```diff
@@ -241,6 +242,35 @@ struct convert<std::list<T> > {
...
+    Node node(NodeType::Sequence);
+    for (typename std::array<T, S>::const_iterator it = rhs.begin();
+        it != rhs.end(); ++it)
+      node.push_back(*it);
+    return node;
+  }
+
+  static bool decode(const Node& node, std::array<T, S>& rhs) {
+    if (!node.IsSequence())
+      return false;
```
jbeder: This should probably also fail to decode if the node has the wrong size, right? I wouldn't expect to parse [1, 2] as an array of length 1 or 3.

include/yaml-cpp/node/convert.h
```diff
@@ -241,6 +242,35 @@ struct convert<std::list<T> > {
...
+
+    std::size_t index = 0;
+    for (const_iterator it = node.begin(); it != node.end(); ++it) {
+#if defined(__GNUC__) && __GNUC__ < 4
+      // workaround for GCC 3:
+      rhs.at(index) = it->template as<T>();
```
jbeder: Use indexing rather than at() (which will be clear if you have the two indexing changes noted above).

test/node/node_test.cpp
```diff
@@ -154,6 +154,20 @@ TEST(NodeTest, SimpleSubkeys) {
...
+  evens[2] = 6;
+  evens[3] = 8;
+  evens[4] = 10;
+
+  Node node;
+  node["evens"] = evens;
+  std::array<int, 5> actualEvens = node["evens"].as<std::array<int, 5> >();
+  EXPECT_EQ(evens, actualEvens);
+}
+
```
jbeder: Please add some failing tests, e.g., for parsing a map, scalar, and wrong-size array.

test/node/node_test.cpp
```diff
@@ -154,6 +154,20 @@ TEST(NodeTest, SimpleSubkeys) {
   EXPECT_EQ("monkey", node["username"].as<std::string>());
 }
 
+TEST(NodeTest, StdArray) {
+  std::array<int, 5> evens;
```
jbeder: Might as well initialize this with an initializer list.

test/node/node_test.cpp
```diff
@@ -154,6 +154,20 @@ TEST(NodeTest, SimpleSubkeys) {
...
+  evens[0] = 2;
+  evens[1] = 4;
+  evens[2] = 6;
+  evens[3] = 8;
+  evens[4] = 10;
+
+  Node node;
+  node["evens"] = evens;
+  std::array<int, 5> actualEvens = node["evens"].as<std::array<int, 5> >();
```
jbeder: No need for space between > >

jbeder:
Great idea, thanks!

xpac27:
Thanks a lot for the code review! Glad you like the idea. This new converter makes things way simpler in my project.

include/yaml-cpp/node/convert.h
xpac27: There seems to be a few > > left in those files:
- include/yaml-cpp/exceptions.h
- include/yaml-cpp/node/convert.h
- include/yaml-cpp/node/detail/node_iterator.h
- test/node/node_test.cpp

I could fix them too in an other PR?

include/yaml-cpp/node/convert.h
```diff
@@ -243,29 +243,29 @@ struct convert<std::list<T> > {
...
-struct convert<std::array<T, S> > {
-  static Node encode(const std::array<T, S>& rhs) {
+template <typename T, std::size_t N>
+struct convert<std::array<T, N>> {
+  static Node encode(const std::array<T, N>& rhs) {
     Node node(NodeType::Sequence);
-    for (typename std::array<T, S>::const_iterator it = rhs.begin();
-        it != rhs.end(); ++it)
-      node.push_back(*it);
+    for (auto element : rhs) {
```
jbeder: const auto&

(no need to copy)

include/yaml-cpp/node/convert.h
```diff
@@ -243,29 +243,29 @@ struct convert<std::list<T> > {
...
+      node.push_back(element);
+    }
     return node;
   }
 
-  static bool decode(const Node& node, std::array<T, S>& rhs) {
+  static bool decode(const Node& node, std::array<T, N>& rhs) {
     if (!node.IsSequence())
```
jbeder: Forgot to say before, always put {} around your "if" clauses (we've obviously not 100% consistent in yaml-cpp, but going that way, and new code should always use this).

test/node/node_test.cpp
```diff
@@ -155,19 +163,21 @@ TEST(NodeTest, SimpleSubkeys) {
...
-  std::array<int, 5> actualEvens = node["evens"].as<std::array<int, 5> >();
+  std::array<int, 5> actualEvens = node["evens"].as<std::array<int, 5>>();
   EXPECT_EQ(evens, actualEvens);
 }
 
+TEST(NodeTest, StdArrayWrongSize) {
+  std::array<int, 3> evens {{ 2, 4, 6 }};
+  Node node;
+  node["evens"] = evens;
+  EXPECT_THROW_REPRESENTATION_EXCEPTION((node["evens"].as<std::array<int, 5>>()),
```
jbeder: I'd prefer inlining this rather than making a separate macro, since you're only using it once.

test/node/node_test.cpp
```diff
@@ -155,19 +163,21 @@ TEST(NodeTest, SimpleSubkeys) {
...
 
 TEST(NodeTest, StdArray) {
-  std::array<int, 5> evens;
-  evens[0] = 2;
-  evens[1] = 4;
-  evens[2] = 6;
-  evens[3] = 8;
-  evens[4] = 10;
-
+  std::array<int, 5> evens {{ 2, 4, 6, 8, 10 }};
```
jbeder: The {{ }} is correct but looks a little funky; prefer

... evens = {...};

include/yaml-cpp/node/convert.h
jbeder: Sure, SGTM.

xpac27: Will do ^^

test/node/node_test.cpp
xpac27: std::array<int, 5> evens = { 2, 4, 6, 8, 10 };

Actually produces:

    warning: suggest braces around initialization of subobject [-Wmissing-braces]
      std::array<int, 5> evens = { 2, 4, 6, 8, 10 };
                                   ^~~~~~~~~~~~~~
                                   {             }

test/node/node_test.cpp
xpac27: Would it be even better to inline and create 2 tests in order to have only 1 assertion per test? I think it would also make the tests less big and thus easier to read.

    TEST(NodeTest, StdArrayWrongSizeThrowsRightException) {
      ...
    TEST(NodeTest, StdArrayWrongSizeThrowsExceptionWithRightMessage) {
      ...

I personally don't find the camelcased test names very readable but I wrote it like that to stay consistant with the rest of the test suite :-/

include/yaml-cpp/node/convert.h
xpac27: of course... my bad.

PR #386, opened 2016-06-08
Change node_map type from map<ptr,ptr> to vector<pair<ptr,ptr>>

c0nk:
Map nodes are now iterated over in document order.

### DIFF
--- a/src/node_data.cpp
+++ b/src/node_data.cpp
@@ -256,18 +256,7 @@ void node_data::reset_map() {
 }
 
 void node_data::insert_map_pair(node& key, node& value) {
-
-  bool found = false;
-  for (auto& entry : m_map) {
-    if (entry.first == &key) {
-      entry.second = &value;
-      found = true;
-      break;
-    }
-  }
-  if (!found) {
-    m_map.emplace_back(&key, &value);
-  }
+  m_map.emplace_back(&key, &value);
 
   if (!key.is_defined() || !value.is_defined())
     m_undefinedPairs.emplace_back(&key, &value);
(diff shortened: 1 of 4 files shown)


ysyrov:
I'm cool with this one. Thank you, Graham.
BTW, uses C++11, when the rest of code is not.

src/node_data.cpp
```diff
@@ -256,9 +256,21 @@ void node_data::reset_map() {
 }
 
 void node_data::insert_map_pair(node& key, node& value) {
-  m_map[&key] = &value;
+
+  bool found = false;
```
jbeder: Rereading this, I realized that this makes inserting into a map node a linear operation, which isn't good. The upshot, I think, is that we should always just append this to the list, and make a breaking change to force_insert so that it might duplicate a key when added twice that way.

Sorry for the churn.

src/node_data.cpp
```diff
@@ -256,9 +256,21 @@ void node_data::reset_map() {
...
+  for (auto& entry : m_map) {
+    if (entry.first == &key) {
```
ysyrov: shouldn't this be 
entry.first->is(key)
?

src/node_data.cpp
ysyrov: On the plus side, get(key) is much faster with vector!
and that's going to dominate access pattern

ysyrov: I see your concern, and share it...
Adding std::unordered_map<node*, size_t> is simple and will take care of it
;)

c0nk:
Is there something that still needs to be done here?

ysyrov:
looks good

c0nk:
Can we get this merged?

c0nk:
@jbeder please come back. Don't abandon us. lol

PR #399, opened 2016-07-26
Use _NOEXCEPT on Visual Studio instead of noexcept

elnormous:
Since noexcept is supported only on Visual Studio 2015, use _NOEXCEPT on older versions of VS.

### DIFF
--- a/include/yaml-cpp/exceptions.h
+++ b/include/yaml-cpp/exceptions.h
@@ -7,6 +7,12 @@
 #pragma once
 #endif
 
+#ifdef _MSC_VER
+#define NOEXCEPT _NOEXCEPT
+#else
+#define NOEXCEPT noexcept
+#endif
+
 #include "yaml-cpp/mark.h"
 #include "yaml-cpp/traits.h"
 #include <stdexcept>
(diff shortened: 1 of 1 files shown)


include/yaml-cpp/exceptions.h
```diff
@@ -7,6 +7,12 @@
 #pragma once
 #endif
 
+#ifdef _MSC_VER
+#define NOEXCEPT _NOEXCEPT
```
jbeder: I'm uncomfortable defining something so common in a header.

Maybe narrow the scope of the definition and #undef it at the end?

benapetr: how about just prefixing it with something that make it look less common? like `#define YAML_CPP_NOEXCEPT _NOEXCEPT`

benapetr:
This is extremely useful and needed change +1 to implement this

include/yaml-cpp/exceptions.h
jbeder: YAML_CPP_NOEXCEPT is fine with me

jbeder: ... and also #undef it at the end of the header

benapetr:
OK, I didn't create this so I had to submit a new one: https://github.com/jbeder/yaml-cpp/pull/517

jbeder:
Merged that one, so closing this.

PR #342, opened 2015-11-11
Replace Boost usage with C++11 features

blair1618:
- Adds 'std=c++11' compiler flags
 - Replaces boost::type_traits with std::type_traits
 - Replaces boost::shared_ptr with std::shared_ptr
 - Replaces std::auto_ptr with std::unique_ptr
 - Replaces raw pointers with std::unique_ptr in ptr_vector, ptr_stack, and SettingChanges
 - Replaces boost::noncopyable with deleted copy and assignment operators
 - Replaces boost::next with std::next
 - Replaces boost::enable_if with std::enable_if
 - Replaces boost::is_convertible with std::is_convertible
 - Replaces ptrdiff_t with std::ptrdiff_t
 - Replaces boost::iterator_facade and boost::iterator_adaptor with std::iterator, borrowing the 'proxy reference' technique from boost
 - Removes Boost dependency from CMakeLists
 - Formats changed files using clang-format

PR #402, opened 2016-08-10
add information about cmake's -DBOOST_ROOT parameter

michaelkonecny:
I always have to look for this, so adding it to the readme might save some people some time.


blair1618:
yaml-cpp no longer uses Boost as of https://github.com/jbeder/yaml-cpp/pull/342

PR #425, opened 2016-10-17
remove const from as_if::operator()() return

vmatare:
Returning a const value prevents the use of non-copyable return types
such as unique_ptr.

The return value needs to be mutable so it can be moved from. This also allows the compiler to do copy elision with copyable types.


vmatare:
Btw, this patch also removes the need for unwieldy hacks when returning abstract types. We don't want to return raw pointers since `convert::decode()` is always expected to throw. So as long as we have to copy the `convert::decode()` return value we're basically stuck with `shared_ptr`s, which has potentially undesirable performance and ownership implications.

PR #437, opened 2016-11-26
RegExp templates

hjanetzek:
This branch uses templates for the pattern matching instead of the dynamic RegExp module. 

This is still WIP but seems to work correctly already - feedback is welcome!


hjanetzek:
This benchmark shows quite a huge speed and memory improvement from switching to static regexp https://github.com/tangrams/tangram-es/pull/1118. There are other changes in that branch, but the RegExp allocations stood out.

PR #446, opened 2016-12-05
cmake: make yaml-cpp package relocatable

sergiud:
This PR makes `yaml-cpp-config.cmake` relocatable allowing the package to be used both from the build tree and the install path. The changes are backwards compatible.

Typical usage of `yaml-cpp-config.cmake`:

```cmake
find_package (yaml-cpp 0.5.3 REQUIRED)

add_executable (myproject main.cpp)
target_link_libraries (myproject yaml-cpp)
```

The `yaml-cpp-config.cmake` is now always installed under `${CMAKE_INSTALL_PREFIX}/(lib|lib64)/cmake/yaml-cpp`. This change was necessary to simplify the config logic. Previously, `yaml-cpp-config.cmake` was installed under `${CMAKE_INSTALL_PREFIX}/CMake` for Win32 (excluding cygwin), and `${CMAKE_INSTALL_PREFIX}/(lib|lib64)/cmake/yaml-cpp` on other platforms. The different config paths that depend on the platform in use are unnecessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -264,15 +264,15 @@ else()
 	set(_library_dir lib)
 endif()
 
-set(INCLUDE_INSTALL_ROOT_DIR ${CMAKE_INSTALL_PREFIX}/include)
+set(INCLUDE_INSTALL_ROOT_DIR include)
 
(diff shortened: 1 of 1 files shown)


ligfx:
+1 This solved my issue with propagating include paths on the `yaml-cpp` target when built inside a larger target.

I can't speak to the change in `yaml-cpp-config.cmake` location.

PR #452, opened 2016-12-09
Fixed compiler warning -Wdeprecated with clang.

jbreitbart:
Starting with C++11 implicit copy-constructors are deprecated when the class
has a user defined destructor.

### DIFF
--- a/include/yaml-cpp/parser.h
+++ b/include/yaml-cpp/parser.h
@@ -47,7 +47,7 @@ class YAML_CPP_API Parser : private noncopyable {
   void Load(std::istream& in);
 
   /**
-   * Handles the next document by calling events on the {@param eventHandler}.
+   * Handles the next document by calling events on the {@code eventHandler}.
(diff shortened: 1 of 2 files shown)


include/yaml-cpp/parser.h
```diff
@@ -47,7 +47,7 @@ class YAML_CPP_API Parser : private noncopyable {
   void Load(std::istream& in);
 
   /**
-   * Handles the next document by calling events on the {@param eventHandler}.
+   * Handles the next document by calling events on the { @param eventHandler }.
```
jbeder: Since this seems to be an incorrect way to refer to a parameter (I can't seem to find any indication it's even allowed to refer to a @param inline like this), can you just change it to @code? I don't want to special-case this one for having a space at the end.

jbreitbart:
Fixed. Sorry, I must have overlooked the comment style in the rest of the file.

PR #463, opened 2017-01-31
Add more std container converters for YAML::Node

rwols:
These commits add conversion ~~and emitter support~~ for

* std::set -- Rendered as a _yaml sequence_. When decoding, duplicate values are silently ignored from a yaml sequence (by virtue of `std::set::insert`/`std::set::emplace`).
* std::unordered_set -- Rendered the same as std::set
* ~~std::multimap -- Rendered as a sequence of pairs~~
* ~~std::unordered_multimap -- Rendered the same as std::multimap~~
* std::unordered_map -- Rendered the same as std::map
* std::deque -- Rendered the same as std::vector
* std::forward_list -- Rendered as a sequence, but decoding happens from back to front
* std::bitset -- Rendered as a yaml string of zeroes and ones
* std::tuple -- Rendered as a yaml sequence

They all have an accompanying unit test.

std::multimap and std::unordered_multimap were not added because we didn't find a canonical way to serialize those data structures to and from yaml.
std::queue, std::priority_queue and std::stack are deliberately not added.

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -8,10 +8,19 @@
 #endif
 
 #include <array>
+#include <bitset>
+#include <cassert>
+#include <deque>
+#include <forward_list>
 #include <limits>
 #include <list>
 #include <map>
+#include <set>
 #include <sstream>
+#include <tuple>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
 #include <vector>
 
 #include "yaml-cpp/binary.h"
@@ -76,7 +85,7 @@ struct convert<const char*> {
 
 template <std::size_t N>
 struct convert<const char[N]> {
-  static Node encode(const char(&rhs)[N]) { return Node(rhs); }
+  static Node encode(const char (&rhs)[N]) { return Node(rhs); }
 };
 
 template <>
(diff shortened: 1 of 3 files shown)


rwols:
I'm not sure what the deal is with travis-ci, the tests run fine here. Their website seems to be down...

include/yaml-cpp/node/convert.h
```diff
@@ -161,83 +167,272 @@ struct convert<bool> {
   YAML_CPP_API static bool decode(const Node& node, bool& rhs);
 };
 
-// std::map
-template <typename K, typename V>
-struct convert<std::map<K, V>> {
-  static Node encode(const std::map<K, V>& rhs) {
+template <class T> class node_push_backer
```
jbeder: I think these helper classes obfuscate what's going on here. Can you just inline the loops instead?

include/yaml-cpp/node/convert.h
```diff
@@ -161,83 +167,272 @@ struct convert<bool> {
...
+      m_node->force_insert(t.first, t.second);
+      return *this;
+  }
+
+  node_inserter& operator*()     { return *this; }
+  node_inserter& operator++()    { return *this; }
+  node_inserter  operator++(int) { return *this; }
+};
+
+template <class T> struct convert_as_sequence
```
jbeder: (these are fine)

include/yaml-cpp/node/convert.h
```diff
@@ -161,83 +167,272 @@ struct convert<bool> {
...
+    Node node(NodeType::Sequence);
+    
+    std::copy(sequence.begin(), sequence.end(), 
+      node_push_backer<typename T::value_type>(node));
+
+    return node;
+  }
+  static bool decode(const Node& node, T& sequence)
+  {
+    if (!node.IsSequence())
```
jbeder: braces around single-line "if"

(throughout, including code that already exists)

include/yaml-cpp/node/convert.h
```diff
@@ -161,83 +167,272 @@ struct convert<bool> {
...
   }
 };
 
+// std::vector
+template <typename T>
+struct convert<std::vector<T>> 
+: public convert_as_sequence<std::vector<T>> {};
+
+// std::deque
+template <class T, class Alloc>
```
jbeder: Why are you adding the full template params to some of these but not others?

include/yaml-cpp/node/convert.h
```diff
@@ -161,83 +167,272 @@ struct convert<bool> {
...
+    return node;
+  }
+
+  static bool decode(const Node& node, T& rhs) {
     if (!node.IsMap())
       return false;
 
     rhs.clear();
-    for (const_iterator it = node.begin(); it != node.end(); ++it)
+    for (const auto& kv : node)
```
jbeder: What if the value is non-null? Then it's not a set, and should return false.

include/yaml-cpp/node/convert.h
```diff
@@ -161,83 +167,272 @@ struct convert<bool> {
...
 #endif
     return true;
   }
 };
 
-// std::vector
-template <typename T>
-struct convert<std::vector<T>> {
-  static Node encode(const std::vector<T>& rhs) {
+// A multimap is a sequence of (sorted) key-value pairs.
```
jbeder: This is a bit surprising. There's no mention of multimap in the YAML spec; where is this definition from?

include/yaml-cpp/node/convert.h
```diff
@@ -275,6 +468,27 @@ struct convert<std::array<T, N>> {
...
+  {
+    return convert<std::string>::encode(rhs.to_string());
+  }
+
+  static bool decode(const Node& node, value_type& rhs)
+  {
+    std::string representation;
+    if (!convert<std::string>::decode(node, representation)) return false;
+    if (representation.size() != N) return false;
+    rhs = value_type(representation);
```
jbeder: "If any characters examined in str are not zero or one, it throws std::invalid_argument."

... but decode is supposed to return false.

test/node/node_test.cpp
```diff
@@ -211,6 +211,20 @@ TEST(NodeTest, StdVector) {
   EXPECT_EQ(primes, node["primes"].as<std::vector<int>>());
 }
 
+TEST(NodeTest, StdDeque) {
```
jbeder: Can you add failing tests for bad conversions to the various types?

rwols: Will do!

include/yaml-cpp/node/convert.h
rwols: I'll fix that.

include/yaml-cpp/node/convert.h
rwols: I'll apply clang format.

include/yaml-cpp/node/convert.h
rwols: Right. Here is a larger problem:

In stlemitter.h, an `std::set` is serialized as a _yaml sequence_, not as a yaml map as I'm doing here. So what should it be? I'm now leaning more towards realizing an `std::set` as a yaml sequence.

include/yaml-cpp/node/convert.h
rwols: Good catch, I'll fix that.

include/yaml-cpp/node/convert.h
rwols: There is indeed no notion of a "yaml multimap", so I made up my own definition. It's up to debate what the correct one is or how one should serialize a multimap. I thought that serializing it as a sequence of pairs was the "correct" thing to do.

rwols:
OK, the travis website seems to be working now and I just learned that GCC 4.6.2 does not have `std::map::emplace`, so I'll write a fix for that.

rwols:
The builds succeed and the tests pass but one of them timed out now (clang's). I don't know why it took 13 minutes...

I'm moving on to stlemitter.h for arbitrary STL container support. Then I want to add unit tests for STL container -> emit -> parse -> decode YAML::Node -> should equal original STL container.

rwols:
I added conversion tests between various STL containers, can you check wether it's enough coverage? I'm gonna go ahead and add `std::tuple` support now 👍 , after that I think I'm done.

rwols:
OK I'm pretty much satisfied with this set of commits. I have some more ideas for tuples now that `std::tuple` is supported, but I'll do that in another PR.

test/integration/emit_parse_decode_test.cpp
```diff
@@ -0,0 +1,300 @@
...
+TEST_F(EmitParseDecodeTest, StdList) {
+  MockList<int> primes;
+  primes.push_back(2);
+  primes.push_back(3);
+  primes.push_back(5);
+  primes.push_back(7);
+  primes.push_back(11);
+  primes.push_back(13);
+  std::list<int> sameprimes(primes.begin(), primes.end());
+  // check(primes);  // <--- this fails to compile for some reason
```
rwols: I still don't understand the compile error here

rwols:
After thinking about it for a while, the std::enable_if's in stlemitter.h are pointless, so I am removing them.

rwols:
I've been thinking about the various sequence and map converters in convert.h now, and I believe they are too "strict". For example, consider these two yaml snippets (a typical travis file):

```yaml
language: cpp
os: linux # [1]
# etc
```
```yaml
language: cpp
os: # [2]
  - linux
  - osx
# etc
```

In [1], `node["os"].IsSequence() == false` and `node["os"].IsScalar() == true`, while in [2] it's the other way around. But, in my opinion, the call
```c++
const auto os = node["os"].as<std::vector<std::string>>();
```
should still succeed for [1], it's just going to be a vector with 1 element in it. Even more so, you could have `node["os"].IsNull() == true` and return a vector with 0 elements in it. This would remove a case-by-case basis that users now need to take care of. So, users would just do
```c++
const auto os = node["os"].as<std::vector<std::string>>();
for (const auto& item : os) {
  // do things with item
}
```
and if the node was null, no problem: no iteration takes place. If the node was a scalar, no problem: precisely one item iteration takes place. If the node was a sequence, multiple iteration takes place.

So in decoding a sequence, you'd do something like

```c++
...
  bool decode(const Node& node, T& sequence) {
    sequence.clear();

    // OK: empty sequence returned
    if (node.IsNull())
      return true;

    // OK: sequence with one element returned
    if (node.IsScalar()) {
      sequence.push_back(node.template as<typename T::value_type>());
      return true;
    }

    // At this point we require that the node is a sequence
    if (!node.IsSequence())
      return false;

    for (const auto& item : node) {
      sequence.push_back(item.template as<typename T::value_type>());
    }

    return true;
  }
...
```
Let me know what you think. I could incorporate this idea in this PR or make a new PR for it.

test/node/node_test.cpp
rwols: Is this request still relevant? Can you tell me what the missing tests are if any?

test/node/node_test.cpp
```diff
@@ -239,6 +267,272 @@ TEST(NodeTest, StdMap) {
...
+    EXPECT_EQ(*iter0, *iter2);
+    EXPECT_EQ(*iter0, *iter3);
+    ++iter0;
+    ++iter1;
+    ++iter2;
+    ++iter3;
+  }
+
+  // should fail -- a yaml sequence is not a yaml map.
+  EXPECT_THROW_REPRESENTATION_EXCEPTION((node.as<std::map<int, int>>()),
```
jbeder: Same with these guys.

test/node/node_test.cpp
```diff
@@ -239,6 +267,272 @@ TEST(NodeTest, StdMap) {
...
+  // should succeed -- an std::set is considered a yaml sequence
+  const auto primesAsSet = node.as<std::set<unsigned>>();
+  EXPECT_EQ(primesAsSet.size(), primesAsVec.size());
+
+  for (const auto prime : primesAsVec) {
+    EXPECT_NE(primesAsSet.find(prime), primesAsSet.end());
+  }
+}
+
+TEST(NodeTest, ConvolutedWayToRemoveDuplicates) {
```
jbeder: Maybe name this test "SequenceWithDuplicatesCanBeParsedAsSet"

test/node/node_test.cpp
```diff
@@ -239,6 +267,272 @@ TEST(NodeTest, StdMap) {
...
+
+  const std::vector<int> withDuplicates{1, 5, 3, 6, 4, 7, 5, 3, 3, 4, 5,
+                                        5, 3, 2, 7, 8, 9, 9, 9, 5, 2, 0};
+
+  const auto node = Node(withDuplicates);
+
+  // should succeed -- an std::set is considered a yaml sequence
+  const auto withoutDuplicatesAndSorted = node.as<std::set<int>>();
+
+  EXPECT_EQ(withoutDuplicatesAndSorted.size(), 10);
```
jbeder: testing::UnorderedElementsAreArray

test/node/node_test.cpp
jbeder: I think this is fine now.

include/yaml-cpp/node/convert.h
```diff
@@ -161,109 +170,222 @@ struct convert<bool> {
...
+#endif  // defined(__GNUC__)
+    }
+    return true;
+  }
+};
+
+// Helper class for converting something which is "set-like" to and from a Node.
+// A set is realized as a yaml sequence.
+template <class T>
+struct convert_as_set {
```
jbeder: The spec defines a set as the keys to a map with null values:

http://yaml.org/type/set.html

as you had before.

Although I'm not sure if I like that; but perhaps we should either do that or not implement set?

include/yaml-cpp/node/convert.h
```diff
@@ -161,109 +170,222 @@ struct convert<bool> {
...
 };
 
-// std::vector
-template <typename T>
-struct convert<std::vector<T>> {
-  static Node encode(const std::vector<T>& rhs) {
+// Helper class for converting something which is "multimap-like" to and from a
+// Node. A multimap is realized as a yaml sequence of key-value pairs.
+template <class T>
+struct convert_as_multimap {
```
jbeder: I think we should nix multimap; it's not in the spec, and there are various ways you could do it.

In particular, a sequence of key-value pairs is an ordered map, according to the spec:

http://yaml.org/type/omap.html

jbeder:
Also, regarding your last comment: I don't think a scalar should be allowed to be parsed as a sequence with a single node.

include/yaml-cpp/node/convert.h
rwols: I just checked what PyYAML does and it follows the spec literally:

```python
>>> import yaml
>>> aset = {2,3,5,7,11,13,13,13,13,13,13}
>>> yaml.dump(aset)
'!!set {2: null, 3: null, 5: null, 7: null, 11: null, 13: null}\n'
```
I don't like this either because:
1. It makes for an awkward syntax with the "?" symbol when you have complex keys
2. Redundant ": null" substrings

It's your choice though, let me know what you want so I can adjust the converters and tests accordingly. Me personally I vote for sets realized as sequences to maintain human readability.

include/yaml-cpp/node/convert.h
rwols: OK, I will remove (ordered) multimap for now.

rwols: Fixed in 	8dfc376

include/yaml-cpp/node/convert.h
```diff
@@ -161,109 +167,222 @@ struct convert<bool> {
...
-template <typename K, typename V>
-struct convert<std::map<K, V>> {
-  static Node encode(const std::map<K, V>& rhs) {
+// Helper class for converting something which is "sequence-like" to and from a
+// Node.
+template <class T>
+struct convert_as_sequence {
+  static Node encode(const T& sequence) {
+    Node node(NodeType::Sequence);
+    for (const auto& item : sequence)
```
rwols: Fixed in dc35076

test/node/node_test.cpp
rwols: Fixed in c34a8bf

test/node/node_test.cpp
rwols: Fixed in c34a8bf

test/node/node_test.cpp
rwols: Fixed in c34a8bf

test/node/node_test.cpp
```diff
@@ -239,6 +267,272 @@ TEST(NodeTest, StdMap) {
...
+}
+
+TEST(NodeTest, ConversionsBetweenSequences) {
+  const std::vector<int> primesAsVec{2, 3, 5, 7, 11, 13};
+
+  const auto node = Node(primesAsVec);
+
+  // should succeed -- yaml-cpp doesn't care what kind of underlying sequence
+  // type it was.
+  const auto primesAsDeque = node.as<std::deque<int>>();
```
rwols: I didn't use testing::ElementsAreArray because `std::array` screwed up that function. Anyway, fixed in c34a8bf

rwols:
> Can you split the emitter stuff off into a separate PR?

Fixed in 107b45a

rwols:
rebased to latest head & squashed

Issue #469, opened 2017-02-21
float is precision is not precise enough


rwols:
When I serialize the float `0.12345679` it gets serialized as `0.12345678`.

This fails: (include the header `boost/math/special_functions/next.hpp`)

```c++
float x = 0.12345679f;
float another_x;
{
    std::ofstream output("test.yml");
    YAML::Emitter emitter(output);
    emitter << YAML::BeginDoc 
            << YAML::BeginMap 
            << YAML::Key << "x" 
            << YAML::Value << x 
            << YAML::EndMap 
            << YAML::EndDoc;
}
{
    std::ifstream input("test.yml");
    const auto doc = YAML::Load(input);
    another_x = doc["x"].as<float>();
}
CHECK_TRUE(std::abs(boost::math::float_distance(x, another_x)) < 1);
```

rwols:
I'd like to note that `std::abs(boost::math::float_distance(x, another_x)) < 3` DOES succeed

jbeder:
Fixed by #649.

PR #470, opened 2017-02-22
Fix float precision

rwols:
Replaced

    std::numeric_limits<T>::digits10 + 1
with

    std::numeric_limits<T>::max_digits10

everywhere.

This fixes https://github.com/jbeder/yaml-cpp/issues/469

Understandably some tests will start failing now

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -76,7 +76,7 @@ struct convert<const char*> {
 
 template <std::size_t N>
 struct convert<const char[N]> {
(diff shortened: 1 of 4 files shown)


rwols:
https://github.com/jbeder/yaml-cpp/pull/470/commits/6ebfa37b3d7cc5f764b4bd0d34e405c0c5346c13 should fix the failing unit tests. I tried using

```c++
Emitter emitter;
emitter.SetFloatPrecision(3);
emitter.SetDoublePrecision(3);
```
but this only works for the very first float (or double) that gets emitted -- the floats that follow use the default precision. So I had to replace things like

```
1.01 --> 1.009999999998
```

Issue #404, opened 2016-08-11
noexcept is not supported by vs2013


xuboying:
```
#ifndef _MSC_VER
#define NOEXCEPT noexcept
#else
#define NOEXCEPT
#endif
```

https://stackoverflow.com/questions/18387640/how-to-deal-with-noexcept-in-visual-studio

SimplyKnownAsG:
looks like there is also a `_NOEXCEPT` macro within VS.

http://stackoverflow.com/a/22959202/910963

xuboying:
Thanks for the tip!

SGSSGene:
I don't think this is an issue anymore. Please reopen this ticket if this issue still exists.

PR #477, opened 2017-03-19
noexcept fix for visual studio

SimplyKnownAsG:
I ran the test suite, i neither know which tests were disabled nor why.

```
[----------] Global test environment tear-down
[==========] 899 tests from 12 test cases ran. (752 ms total)
[  PASSED  ] 899 tests.

  YOU HAVE 8 DISABLED TESTS
```

This addresses #404 (my other pull request had the wrong user, and also conflicted when defining `NOEXCEPT` with other projects).

### DIFF
--- a/include/yaml-cpp/exceptions.h
+++ b/include/yaml-cpp/exceptions.h
@@ -8,12 +8,16 @@
 #endif
 
 #ifndef NOEXCEPT
-#ifndef _MSC_VER
-#define NOEXCEPT noexcept
-#else
+#ifdef _MSC_VER
+#ifdef _NOEXCEPT
 #define NOEXCEPT _NOEXCEPT
-#endif
-#endif
+#else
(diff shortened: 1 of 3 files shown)


include/yaml-cpp/exceptions.h
```diff
@@ -7,6 +7,14 @@
 #pragma once
 #endif
 
+#ifndef NOEXCEPT
+#ifndef _MSC_VER
```
jbeder: Do you want to check the MSVC version number?

jbeder:
(Also, the 8 disabled tests are spec tests that have errors in the spec. They're there for documentation.)

SimplyKnownAsG:
sure, though I'm really not sure what version to check. I suppose the easiest thing would really be to check if `_NOEXECPT` was also defined. does that seem reasonable?

PR #485, opened 2017-04-02
Add support for emitting null and strings that are JSON-compatible

Ortham:
yaml-cpp currently only supports emitting null values as `~`, when they can also be (and JSON requires them to be) `null`. Also, JSON strings don't allow `\x` style escapes, and use UTF-16 surrogate pair escapes instead of `\U` escapes.

This PR adds emitter manipulators to support JSON-style null values and strings. When outputting JSON-style strings, yaml-cpp will escape only the code points it is required to by the YAML and JSON specs, and use two-character escape codes for the code points that have them.

(As an aside, I noticed that the existing code point escapes don't cover all the code points required by the YAML 1.2 spec section 5.1, but I didn't fix that as it's out of scope for this PR as JSON is fine with those that are currently unescaped.)

### DIFF
--- a/include/yaml-cpp/emittermanip.h
+++ b/include/yaml-cpp/emittermanip.h
@@ -19,6 +19,7 @@ enum EMITTER_MANIP {
   // output character set
   EmitNonAscii,
   EscapeNonAscii,
+  EscapeAsJson,
 
   // string manipulators
   // Auto, // duplicate
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -686,14 +686,27 @@ void Emitter::StartedScalar() { m_pState->StartedScalar(); }
 // *******************************************************************************************
 // overloads of Write
 
+StringEscaping::value GetStringEscapingStyle(const EMITTER_MANIP emitterManip) {
+  switch (emitterManip) {
+    case EscapeNonAscii:
+      return StringEscaping::NonAscii;
+    case EscapeAsJson:
+      return StringEscaping::JSON;
+    default:
+      return StringEscaping::None;
+      break;
+  }
+}
+
 Emitter& Emitter::Write(const std::string& str) {
   if (!good())
     return *this;
 
-  const bool escapeNonAscii = m_pState->GetOutputCharset() == EscapeNonAscii;
+  StringEscaping::value stringEscaping = GetStringEscapingStyle(m_pState->GetOutputCharset());
+
   const StringFormat::value strFormat =
       Utils::ComputeStringFormat(str, m_pState->GetStringFormat(),
-                                 m_pState->CurGroupFlowType(), escapeNonAscii);
+                                 m_pState->CurGroupFlowType(), stringEscaping == StringEscaping::NonAscii);
 
   if (strFormat == StringFormat::Literal)
     m_pState->SetMapKeyFormat(YAML::LongKey, FmtScope::Local);
@@ -708,7 +721,7 @@ Emitter& Emitter::Write(const std::string& str) {
       Utils::WriteSingleQuotedString(m_stream, str);
       break;
     case StringFormat::DoubleQuoted:
-      Utils::WriteDoubleQuotedString(m_stream, str, escapeNonAscii);
+      Utils::WriteDoubleQuotedString(m_stream, str, stringEscaping);
       break;
     case StringFormat::Literal:
       Utils::WriteLiteralString(m_stream, str,
@@ -814,8 +827,10 @@ Emitter& Emitter::Write(char ch) {
   if (!good())
     return *this;
 
+
+
   PrepareNode(EmitterNodeType::Scalar);
-  Utils::WriteChar(m_stream, ch);
+  Utils::WriteChar(m_stream, ch, GetStringEscapingStyle(m_pState->GetOutputCharset()));
   StartedScalar();
 
   return *this;
--- a/src/emitterutils.cpp
+++ b/src/emitterutils.cpp
@@ -218,20 +218,34 @@ bool IsValidLiteralScalar(const std::string& str, FlowType::value flowType,
   });
 }
 
-void WriteDoubleQuoteEscapeSequence(ostream_wrapper& out, int codePoint) {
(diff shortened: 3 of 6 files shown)


braindevices:
This PR is 3 years old. Any reason to NOT merge this? It might need to solve conflict now. But I think this one is quite useful when working with json.

src/emitterutils.cpp
```diff
@@ -225,20 +225,34 @@ bool IsValidLiteralScalar(const std::string& str, FlowType::value flowType,
...
+std::pair<uint16_t, uint16_t> EncodeUTF16SurrogatePair(int codePoint) {
+  const uint32_t leadOffset = 0xD800 - (0x10000 >> 10);
+
+  return {
+    leadOffset | (codePoint >> 10),
+    0xDC00 | (codePoint & 0x3FF),
+  };
+}
+
+void WriteDoubleQuoteEscapeSequence(ostream_wrapper& out, int codePoint, bool escapeAsJson) {
```
jbeder, line 237: All of these helper functions would probably read better with the StringEscaping enum, rather than a boolean.

include/yaml-cpp/emittermanip.h
```diff
@@ -36,6 +37,10 @@ enum EMITTER_MANIP {
   LongBool,       // yes, On
   ShortBool,      // y, t
 
+  // null manipulators
+  NullAsTilde,     // ~
```
jbeder, line 41: I think I prefer these as

TildeNull
VerboseNull (or maybe JsonNull?)

which match the bool manipulators all ending in Bool. Open to discussion, though.

src/emitter.cpp
```diff
@@ -678,10 +682,19 @@ Emitter& Emitter::Write(const std::string& str) {
   if (!good())
     return *this;
 
-  const bool escapeNonAscii = m_pState->GetOutputCharset() == EscapeNonAscii;
+  StringEscaping::value stringEscaping = StringEscaping::None;
```
jbeder, line 685: Might make sense to put this as a "constructor" to the enum rather than a function here. Then each case of the switch statement could just be a "return".

test/integration/emitter_test.cpp
```diff
@@ -642,6 +642,23 @@ TEST_F(EmitterTest, Null) {
   ExpectEmit("- ~\n- null value: ~\n  ~: null key");
 }
 
+TEST_F(EmitterTest, NullStyles) {
+  out << BeginSeq
```
jbeder, line 646: This test would read better as a "table" test, where you have a small data table that you then iterate over to test each case in succession.

See, e.g., https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

You could try this explicitly in GoogleTest:

https://chromium.googlesource.com/external/github.com/google/googletest/+/HEAD/googletest/docs/advanced.md#value_parameterized-tests

but I've never used that so I don't know how easy it is. A for loop would be fine also.

test/integration/load_node_test.cpp
```diff
@@ -226,9 +226,44 @@ TEST(NodeTest, IncompleteJson) {
   }
 }
 
-TEST(NodeTest, LoadTildeAsNull) {
-  Node node = Load("~");
+TEST(NodeTest, LoadNullAndTildeAsNull) {
+  Node node = Load("null");
```
jbeder, line 230: Ditto these about a table-driven test.

jbeder:
> This PR is 3 years old. Any reason to NOT merge this? It might need to solve conflict now. But I think this one is quite useful when working with json.

@braindevices No reason; I didn't notice it before. I reviewed the PR, but since it's so old it's possible the original author has dropped it. If they don't respond, would you pick it up?

Ortham:
I no longer use yaml-cpp to emit JSON, but I've rebased this PR and addressed most of the review comments.

test/integration/emitter_test.cpp
Ortham: I've used Google Test's parameterised testing support before, but I wrote the tests to fit the style of the surrounding tests. Can you point me to an existing example in the `yaml-cpp` tests so I can match whatever the new style is supposed to be?

src/emitterutils.cpp
Ortham: Done.

include/yaml-cpp/emittermanip.h
Ortham: Done (I went for `JsonNull`).

src/emitter.cpp
Ortham: Done.

test/integration/emitter_test.cpp
jbeder: Never mind, there are no existing examples and it's less clear what to do here than I had thought. This is fine.

test/integration/emitter_test.cpp
```diff
@@ -788,7 +805,44 @@ TEST_F(EmitterTest, Unicode) {
...
+
+TEST_F(EmitterTest, EscapedJsonString) {
+  out.SetStringFormat(DoubleQuoted);
+  out.SetOutputCharset(EscapeAsJson);
+  out << "\" \\ "
+    "\x01 \x02 \x03 \x04 \x05 \x06 \x07 \x08 \x09 \x0A \x0B \x0C \x0D \x0E \x0F "
+    "\x10 \x11 \x12 \x13 \x14 \x15 \x16 \x17 \x18 \x19 \x1A \x1B \x1C \x1D \x1E \x1F "
+    "\x24 \xC2\xA2 \xE2\x82\xAC \xF0\xA4\xAD\xA2";
+
+  ExpectEmit("\"\\\" \\\\ "
```
jbeder, line 819: These might be easier to read as a raw string (so you don't need all of the escaping).

Ortham: Good point, done.

Ortham:
> Sorry, this is probably frustrating, but: I just received PR #906, and I think I like that naming (and flexibility).
> 
> That would actually simplify this PR: it's just the EscapeAsJson manipulator now. Could you remove the null-related code and just keep the charset stuff?

I've removed the null formatting commits.

PR #486, opened 2017-04-02
fix some warnings in public headers

ibaned:
two variable name shadowing warnings
are fixed

### DIFF
--- a/include/yaml-cpp/node/detail/node.h
+++ b/include/yaml-cpp/node/detail/node.h
@@ -106,9 +106,9 @@ class node {
   node_iterator end() { return m_pRef->end(); }
 
   // sequence
-  void push_back(node& node_in, shared_memory_holder pMemory) {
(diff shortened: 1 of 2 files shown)


include/yaml-cpp/node/detail/node.h
```diff
@@ -106,9 +106,9 @@ class node {
   node_iterator end() { return m_pRef->end(); }
 
   // sequence
-  void push_back(node& node, shared_memory_holder pMemory) {
-    m_pRef->push_back(node, pMemory);
-    node.add_dependency(*this);
+  void push_back(node& node_in, shared_memory_holder pMemory) {
```
jbeder: call this "input", maybe?

ibaned: sure thing. done in commit 842f60c, added to the branch for this PR.

PR #490, opened 2017-05-01
fix shadow warnings with GCC 4.7.2

ibaned:


jbeder:
Thanks for the suggestion, but I prefer the code as-is. I don't think a local variable shadowing a member function is particularly confusing (and this is very common style).

Issue #478, opened 2017-03-20
Segmentation fault on null pointer when running tests built with Release or RelWithDebInfo build types


keith-bennett-gbg:
Note that, to debug this, I added `fsanitize=address -fno-omit-frame-pointer` to compiler and linker flags for the compiled targets. Gist of a diff you can apply: https://gist.github.com/keith-bennett-gbg/6ea43597acd55b2f338f34893f54401b#file-yaml-cpp-address-sanitizer-diff

```
$ git clone "https://github.com/jbeder/yaml-cpp.git"
$ cd yaml-cpp
$ cp "${HOME}/yaml-cpp.address-sanitizer.diff" ./
$ git checkout 0fdb1b910c1a31db80b7521446166d06cad10813 # HEAD as of creation of this issue.
$ git apply yaml-cpp.address-sanitizer.diff
$ mkdir build && cd build
$ cmake -DCMAKE_BUILD_TYPE="RelWithDebInfo" ../ && make -j$(nproc) && make test
-- The C compiler identification is GNU 6.2.0
-- The CXX compiler identification is GNU 6.2.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test FLAG_WEXTRA
-- Performing Test FLAG_WEXTRA - Success
-- Found PythonInterp: /usr/bin/python (found version "2.7.12") 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Configuring done
-- Generating done
-- Build files have been written to: /home/keithb/dev/yaml-cpp/build
```
cmake is successful...
build is successful...
running tests...
```
Running tests...
Test project /home/keithb/dev/yaml-cpp-second/build
    Start 1: yaml-test
[...]

Issue #493, opened 2017-05-10
Embedded version of gtest/gmock is buggy with recent compilers


rleigh-codelibre:
Fails on Ubuntu 15.04 with GCC 6.3.0 and FreeBSD 11 with clang 3.8.0.  Probably C++-11 by default for both?

```
% gdb test/run-tests
…
(gdb) run
…
[----------] 10 tests from EncodingTest
[ RUN      ] EncodingTest.UTF8_noBOM

Program received signal SIGSEGV, Segmentation fault.
testing::internal::FunctionMockerBase<void (YAML::Mark const&)>::InvokeWith(std::tuple<YAML::Mark const&> const&) (
    args=std::tuple containing = {...}, this=0x555555c1b490) at /data/code/yaml-cpp/test/gmock-1.7.0/include/gmock/gmock-spec-builders.h:1530
1530            this->UntypedInvokeWith(&args))->GetValueAndDelete();
(gdb) bt
#0  testing::internal::FunctionMockerBase<void (YAML::Mark const&)>::InvokeWith(std::tuple<YAML::Mark const&> const&) (
    args=std::tuple containing = {...}, this=0x555555c1b490) at /data/code/yaml-cpp/test/gmock-1.7.0/include/gmock/gmock-spec-builders.h:1530
#1  testing::internal::FunctionMocker<void (YAML::Mark const&)>::Invoke(YAML::Mark const&) (a1=..., this=0x555555c1b490)
    at /data/code/yaml-cpp/test/gmock-1.7.0/include/gmock/gmock-generated-function-mockers.h:97
#2  YAML::MockEventHandler::OnDocumentStart (this=this@entry=0x555555c1b488, gmock_a1=...) at /home/rleigh/code/yaml-cpp/test/mock_event_handler.h:10
#3  0x000055555586f25f in YAML::SingleDocParser::HandleDocument (this=this@entry=0x7fffffffdbb0, eventHandler=...)
    at /home/rleigh/code/yaml-cpp/src/singledocparser.cpp:32
#4  0x000055555585457d in YAML::Parser::HandleNextDocument (this=this@entry=0x7fffffffdd00, eventHandler=...)
    at /home/rleigh/code/yaml-cpp/src/parser.cpp:39
#5  0x00005555556255f5 in YAML::HandlerTest::Parse (
[...]

PR #494, opened 2017-05-10
Upgrade embedded version of gmock+gtest to 1.8.0

rleigh-codelibre:
See also: https://github.com/jbeder/yaml-cpp/issues/493

This commit upgrades gmock and gtest to 1.8.0, fixing the segfault noted in the issue.  It very much looks like gmock 1.7.0 is broken, and this defect is exposed when using reasonably contemporary compilers (both GCC and Clang).  Not bleeding edge; just those used in the latest stable releases of FreeBSD and Ubuntu, and doubtless others as well.


jbeder:
Looks like it fails on clang? Maybe worth splitting these up into to PRs so that they don't get squashed, in any case.

rleigh-codelibre:
Yes, I can split them apart.  The clang error is because Travis is using clang 3.4 with the *GCC 4.6* headers, and it's hitting some brokenness in those libstdc++ headers.  These are not a particularly great combination, and I'm not sure how realistic this scenario is for real-world usage.

I'm investigating if there's a way to work around this, but it may well be the case that the Travis test setup is broken for these type_trait templates.  There's no bug in either the yaml-cpp or googletest code that I can see.

rleigh-codelibre:
As a data point, on FreeBSD 10.3 with clang++ 3.4.1, this builds and tests perfectly.  It's definitely the unholy combination of clang++ with libstdc++ that's at fault here.

rleigh-codelibre:
The last commit is a total cheat, which drops the linux/clang combination from the matrix.  From what I can tell, and what's been reported, it's not a particularly viable setup.  Looks like people who want to test this are configuring travis to use their own docker images which are set up with a working compiler and standard library combination.  At work, we use FreeBSD for this purpose (this PR works perfectly there), but the MacOS X axis of the matrix is already providing good coverage here.

keith-bennett-gbg:
Also seems to resolve https://github.com/jbeder/yaml-cpp/issues/478

rleigh-codelibre:
For the clang with buggy libstdc++ issue, see also:

- https://groups.google.com/forum/#!topic/travis-ci/YvZuAVBNYMw
- http://www.itgo.me/a/x2917008569641495753/travis-ci-with-clang-3-4-and-c11/30925448
- http://redmine.gromacs.org/issues/1886

If you liked, I could look at using a more recent libstdc++ with travis?

rleigh-codelibre:
@jbeder Did you want any further changes making on this PR?

jbeder:
@rleigh-codelibre Sounds good. I'll merge this, but can you file an issue about that linux/clang issue, referencing this PR, and the links you've found? Not that we need to do any action on that right now, but just so we surface it.

rleigh-codelibre:
@jbeder Hope the above issue and cross-references are OK.

PR #517, opened 2017-07-25
Replace noexcept with macro compatible with VS

benapetr:
This way it's possible to build using older MSVC (<13) that don't
support this yet. Macro is undefined in each file where it is used so it
should stack well with other libs and sources.

### DIFF
--- a/src/exceptions.cpp
+++ b/src/exceptions.cpp
@@ -1,19 +1,31 @@
 #include "yaml-cpp/exceptions.h"
 
+// This is here for compatibility with older versions of Visual Studio
+// which don't support noexcept
+#ifdef _MSC_VER
+    #define YAML_CPP_NOEXCEPT _NOEXCEPT
+#else
+    #define YAML_CPP_NOEXCEPT noexcept
+#endif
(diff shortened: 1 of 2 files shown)


src/exceptions.cpp
```diff
@@ -1,19 +1,31 @@
...
+BadConversion::~BadConversion() YAML_CPP_NOEXCEPT {}
+BadDereference::~BadDereference() YAML_CPP_NOEXCEPT {}
+BadSubscript::~BadSubscript() YAML_CPP_NOEXCEPT {}
+BadPushback::~BadPushback() YAML_CPP_NOEXCEPT {}
+BadInsert::~BadInsert() YAML_CPP_NOEXCEPT {}
+EmitterException::~EmitterException() YAML_CPP_NOEXCEPT {}
+BadFile::~BadFile() YAML_CPP_NOEXCEPT {}
 }
+
+#undef YAML_CPP_NOEXCEPT
```
benapetr: This undef is probably not necessary as this file isn't supposed to be included

PR #520, opened 2017-07-28
Added a parameter to add more control of comment offset

rjk9w5:
I wanted this for generating config files that users would interact with on a regular basis. This makes it a little easier to look at as a user. The existing tests still work.

### DIFF
--- a/include/yaml-cpp/emittermanip.h
+++ b/include/yaml-cpp/emittermanip.h
@@ -113,11 +113,13 @@ inline _Tag SecondaryTag(const std::string content) {
 }
 
 struct _Comment {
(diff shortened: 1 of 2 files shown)


rjk9w5:
I added this for my own purposes but figured I would share it because sharing is caring. Super simple change that shouldn't effect any existing code.

Old:
```yaml
yaml: 
  language: cpp # the language field
  xdcx: 23423 # Some cryptic field name description 
              # that clears this up
```

New:
```yaml
yaml: 
  language: cpp     # the language field
  xdcx: 23423       # Some cryptic field name description 
                    # that clears this up
```

Issue #110, opened 2015-03-30
Put back some way of accessing anchor/alias string values


GoogleCodeExporter:
```
People seem to want this. The more stars it gets, the more likely I'll do it :)
```

Original issue reported on code.google.com by `jbe...@gmail.com` on 10 Jul 2011 at 4:30

danil-tolkachev:
Hello, @jbeder. Thank you for `yaml-cpp`. Do you plan to add this feauture?

rwols:
Bumping this

jbeder:
I have no plans to do this, but if you submit a PR, I'll look at its API and evaluate whether it's worth it.

rwols:
I'll look into this...

caryoscelus:
@rwols any progress on this?

rwols:
Hi @caryoscelus, I've looked into it and decided that it was going to be way more work than I expected. Feel free to submit a PR.

caryoscelus:
Ok, i'll look into it.

From what i've already seen, adding anchors to event-based parser (EventHandler) shouldn't be hard at all.

E.g.

```c++
virtual void OnAnchor(std::string const& anchor_name) {
  // do nothing by default to keep compatibility
}
```

Haven't looked into node-based parser yet.

perlpunk:
Would be nice to see this. In https://github.com/yaml/yaml-test-suite the format of the events includes the anchor name.
I'm generating a matrix http://matrix.yaml.io/ for several frameworks, and for yaml-cpp I have to replace the anchors with numbers to be able to compare the test results.

PR #530, opened 2017-09-11
Add optional OnAnchor method to EventHandler

caryoscelus:
ref #110


Slion:
I wonder why was this never merged? That should enable #687

Slion:
That did it, thanks for the quick turnaround.

PR #552, opened 2018-01-15
improve readme.md

dongzhuoer:
In the second step, I treat `source directory` as the `src` directory. 

Though my version seems a little verbose, it's at least unambiguous. I think it's important to make it clear whether the user should be.

Feel free to make modifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,11 @@ yaml-cpp uses [CMake](http://www.cmake.org) to support cross-platform building. 
 
 **Note:** If you don't use the provided installer for your platform, make sure that you add CMake's bin folder to your path.
 
-2. Navigate into the source directory, and type:
+2. clone this repository, navigate into the source directory, and make a `build` directory and navigate into it:
 
 ```
+git clone git@github.com:jbeder/yaml-cpp.git
+cd yaml-cpp
 mkdir build
 cd build
 ```


README.md
```diff
@@ -16,9 +16,11 @@ yaml-cpp uses [CMake](http://www.cmake.org) to support cross-platform building.
 
 **Note:** If you don't use the provided installer for your platform, make sure that you add CMake's bin folder to your path.
 
-2. Navigate into the source directory, and type:
+2. clone this repository, navigate into the source directory, and make a `build` directory and navigate into it:
```
jbeder: How about just inserting a new step of

"Clone the repository and navigate into the source directory"?

Also, what if someone just wants to download the tarball?

dongzhuoer: The key point is to make it clear that the use should be at the repository directory, rather than the `src` directory. Maybe show the output of `ls` is a better solution.

jbeder: I don't know; is the original really unclear?

dongzhuoer: Based on the fact that GitHub is a place to storage source code, you can see that the whole repository directory is the **source** directory. But inside it, we can say that `test` is the testing directory, `util` is the utility directory, `src` is the **source** directory. That's why I `cd` into `src`.
    Maybe you can use repository directory instead. But I think the following way will make it crystal clear. User don't have to struggle to guess the meaning of every word exactly the same as how you understand it.

```bash
mkdir build
cd build
ls
```

```
build            LICENSE    yaml-cpp-config.cmake.in
CMakeLists.txt   README.md  yaml-cpp-config-version.cmake.in
CONTRIBUTING.md  src        yaml-cpp.pc.cmake
include          test
install.txt      util
```

SimplyKnownAsG:
Bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb

PR #556, opened 2018-02-01
Fix paths in pkg-config file

jamiesnape:
`pkg-config` should be giving absolute paths rather than relative paths.

Broken by https://github.com/jbeder/yaml-cpp/commit/dc9c750efd0114e5ebf736f5639aaf3df15c4bab and in 0.6.0. If it would be possible to cut a 0.6.1 release with this fix in it, it would be very much appreciated.

Before:

```
$ cat yaml-cpp.pc
libdir=lib
includedir=include

Name: Yaml-cpp
Description: A YAML parser and emitter for C++
Version: 0.6.0
Requires:
Libs: -L${libdir} -lyaml-cpp
Cflags: -I${includedir}
$ pkg-config --cflags yaml-cpp 
-Iinclude
$pkg-config --libs yaml-cpp 
-Llib -lyaml-cpp
```

After (for example with install prefix `/usr/local`):
```
$ cat yaml-cpp.pc 
prefix=/usr/local
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: Yaml-cpp
Description: A YAML parser and emitter for C++
Version: 0.6.0
Requires:
Libs: -L${libdir} -lyaml-cpp
Cflags: -I${includedir}
$ pkg-config --cflags yaml-cpp 
-I/usr/local/include
$ pkg-config --libs yaml-cpp 
-L/usr/local/lib -lyaml-cpp
```

PTAL @jbeder.

### DIFF
--- a/yaml-cpp.pc.cmake
+++ b/yaml-cpp.pc.cmake
@@ -1,5 +1,7 @@
-libdir=@LIB_INSTALL_DIR@
-includedir=@INCLUDE_INSTALL_ROOT_DIR@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=${prefix}
+includedir=${prefix}/@INCLUDE_INSTALL_ROOT_DIR@
+libdir=${exec_prefix}/@LIB_INSTALL_DIR@
 
 Name: Yaml-cpp
 Description: A YAML parser and emitter for C++


jbeder:
Is there a way to automatically test stuff like this? It would be nice to avoid having these CMake changes breaking things.

jamiesnape:
Something like:
```bash
make install
mkdir /tmp/hello_world
cd /tmp/hello_world
cat << 'EOF' > hello_world.cpp
#include <yaml-cpp/yaml.h>
int main() {
  YAML::Node node = YAML::Node("Hello, World!");
  return 0;
}
EOF
$CXX hello_world.cpp -Wall -std=c++11 $(pkg-config --cflags yaml-cpp) $(pkg-config --libs yaml-cpp) -o hello_world_pkg_config
./hello_world_pkg_config
```
Though possibly that would still compile (with a warning from `ld`) if the install prefix were `/usr/` or `/usr/local`.


To test `cmake` you could add something like:
```bash
make install
mkdir /tmp/hello_world
cd /tmp/hello_world
cat << 'EOF' > hello_world.cpp
#include <yaml-cpp/yaml.h>
int main() {
  YAML::Node node = YAML::Node("Hello, World!");
  return 0;
}
EOF
cat << 'EOF' > CMakeLists.txt
cmake_minimum_required(VERSION 2.6)
project(hello_world)
find_package(yaml-cpp REQUIRED)
add_executable(hello_world_cmake hello_world.cpp)
target_link_libraries(hello_world_cmake yaml-cpp)
EOF
cmake .
make
./hello_world_cmake
```

jbeder:
I mean automatically; how do I write that test?

jamiesnape:
The quick way would be paste the above into a file `my_bash_script`, make it executable, and modify .travis.yml:

```yaml
script:
  - make
  - test/run-tests
  - make install
  - ./my_bash_script
```

PR #557, opened 2018-02-04
Update cmake configuration to be able to build both shared and static…

hobbes1069:
… libraries at the same time.

This has only been tested on Fedora linux and probably needs some tweaking for other platforms.


jbeder:
Couple high-level questions:

1) Why do you want to build both static and shared libraries at the same time?
2) Can you find a way to test this on other platforms? I don't have the bandwidth to do so, but don't want to accept a patch that doesn't work cross-platform.

hobbes1069:
The shared library is for all of the packages within the distro to be compiled against, but because yaml-cpp updates frequently break API/ABI, we need something more stable for those developing their own applications. If they use the static library then we don't immediately break their applications with a distro package update.

hobbes1069:
It should work for any linux distro so perhaps just adding a `if(LINUX)` conditional requirement to it would be sufficient.

PR #559, opened 2018-02-17
fix up static, so works as DLL

dand-oss:
Windows has problems exporting static data members.
Recommend using a function for statics instead as below

https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/

Handling class static data members 
Eliminate the static data –
   This can be done by making a static getter function with a local static variable

### DIFF
--- a/include/yaml-cpp/node/detail/node_data.h
+++ b/include/yaml-cpp/node/detail/node_data.h
@@ -81,7 +81,7 @@ class YAML_CPP_API node_data {
                     shared_memory_holder pMemory);
 
  public:
-  static std::string empty_scalar;
+  static std::string& empty_scalar();
 
  private:
   void compute_seq_size() const;
--- a/include/yaml-cpp/node/detail/node_data.h
+++ b/include/yaml-cpp/node/detail/node_data.h
@@ -81,7 +81,7 @@ class YAML_CPP_API node_data {
                     shared_memory_holder pMemory);
 
  public:
-  static std::string& empty_scalar();
+  static const std::string& empty_scalar();
 
  private:
   void compute_seq_size() const;
--- a/src/node_data.cpp
+++ b/src/node_data.cpp
@@ -13,7 +13,10 @@
 namespace YAML {
 namespace detail {
 
-std::string node_data::empty_scalar;
+std::string& node_data::empty_scalar(){
+    static std::string _es;
+    return _es ;
+}
 
 node_data::node_data()
     : m_isDefined(false),
--- a/src/node_data.cpp
+++ b/src/node_data.cpp
@@ -15,7 +15,7 @@ namespace detail {
 
 const std::string& node_data::empty_scalar() {
     static const std::string svalue;
-    return svalue ;
+    return svalue;
 }
 
 node_data::node_data()
--- a/src/node_data.cpp
+++ b/src/node_data.cpp
@@ -13,9 +13,9 @@
 namespace YAML {
 namespace detail {
 
-std::string& node_data::empty_scalar(){
-    static std::string _es;
-    return _es ;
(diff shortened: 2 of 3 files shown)


include/yaml-cpp/node/detail/node_data.h
```diff
@@ -81,7 +81,7 @@ class YAML_CPP_API node_data {
                     shared_memory_holder pMemory);
 
  public:
-  static std::string empty_scalar;
+  static std::string& empty_scalar();
```
jbeder: return a const ref

src/node_data.cpp
```diff
@@ -13,7 +13,10 @@
 namespace YAML {
 namespace detail {
 
-std::string node_data::empty_scalar;
+std::string& node_data::empty_scalar(){
```
jbeder: make sure to run clang-format; it would put a space before the {

src/node_data.cpp
```diff
@@ -13,7 +13,10 @@
...
+    static std::string _es;
```
jbeder: name this something like "empty_scalar" or "scalar" or "value"

jbeder: also, it should be declared as const, since it'll never be modified

sergiud:
Exporting all symbols is really not a good idea. This practically inhibits many optimizations. The DLL fix is really simple, see https://github.com/jbeder/yaml-cpp/issues/461#issuecomment-359411541.

jeanga:
There is an apparent need by many yaml-cpp users to have this issue (and her sisters) fixed once and for all.
Any plan to merge this change?

jbeder:
@jeanga I'm waiting on changes requested to the PR.

src/node_data.cpp
dand-oss: fixed

dand-oss: fixed

include/yaml-cpp/node/detail/node_data.h
dand-oss: fixed

jbeder:
@dand-oss Did you actually make any changes? I don't see a difference in this commit.

RickoLacharity:
Looks like all the changes are in now

src/node_data.cpp
```diff
@@ -13,7 +13,10 @@
 namespace YAML {
 namespace detail {
 
-std::string node_data::empty_scalar;
+const std::string& node_data::empty_scalar() {
+    static const std::string svalue;
+    return svalue ;
```
jbeder: extra space before the semicolon?

clang-format should help you out here.

dand-oss: ok, space removed.  
Can we add clang-format to the tests?

jbeder: Thanks. How would we do that?

src/node_data.cpp
shenlebantongying: This line is for search engine:

undefined symbol: _ZN4YAML6detail9node_data12empty_scalarB5cxx11E

This change at least changed this much bug reports.

https://bugzilla.opensuse.org/show_bug.cgi?id=1178331
https://bugzilla.opensuse.org/show_bug.cgi?id=1160171
https://bugzilla.redhat.com/show_bug.cgi?id=1773357
https://github.com/calamares/calamares/issues/1312#issuecomment-587258287
https://aur.archlinux.org/packages/huggle/#comment-741229
https://stackoverflow.com/questions/52713571/mongod-symbol-lookup-error-mongod-undefined-symbol-zn4yaml6detail9node-data
https://bugzilla.opensuse.org/show_bug.cgi?id=1178332

Issue #505, opened 2017-06-22
Certain constructs used in the CMake build files are deprecated


tambry:
Using CMake 3.9.0-rc3, one can see the following deprecation warnings:
```
CMake Deprecation Warning at 3rdparty/yaml-cpp/CMakeLists.txt:9 (cmake_policy):
  The OLD behavior for policy CMP0012 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


CMake Deprecation Warning at 3rdparty/yaml-cpp/CMakeLists.txt:13 (cmake_policy):
  The OLD behavior for policy CMP0015 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
```

timrae:
bump, still getting these warnings

tambry:
@timrae See #563, which fixes this.

timrae:
Nice :)

PR #563, opened 2018-02-27
Improvements to CMake buildsystem

winterheart:
Various improvements to CMake buildsystem:

* Raised minimal CMake version to 3.1 and added in-built c++11 compiler checks (#377). Since yaml-cpp requires decent compilers that supports c++11, cmake 2.6 requirement becomes useless for old systems.
* Remove cmake_policies (#505), those policies are deprecated or already defined to NEW behavior.
* Externalize GTest subproject (#539), now shared libgmock and libgtest are excluded from installation step.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,26 +1,34 @@
 include(ExternalProject)
 
+if(MSVC)
+    # MS Visual Studio expects lib prefix on static libraries,
+    # but CMake compiles them without prefix
+    # See https://gitlab.kitware.com/cmake/cmake/issues/17338
+    set(CMAKE_STATIC_LIBRARY_PREFIX "")
+endif()
+
 ExternalProject_Add(
 	googletest_project
 	SOURCE_DIR "${CMAKE_SOURCE_DIR}/test/gtest-1.8.0"
 	INSTALL_DIR "${CMAKE_BINARY_DIR}/prefix"
-	CMAKE_ARGS -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR> -DBUILD_GMOCK=ON
+	CMAKE_ARGS
+		-DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+		-DBUILD_GMOCK=ON
+		-Dgtest_force_shared_crt=ON
 )
 
 add_library(gmock UNKNOWN IMPORTED)
 set_target_properties(gmock PROPERTIES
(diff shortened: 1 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -1,26 +1,31 @@
...
+ExternalProject_Add(
+	googletest_project
+	SOURCE_DIR "${CMAKE_SOURCE_DIR}/test/gtest-1.8.0"
+	INSTALL_DIR "${CMAKE_BINARY_DIR}/prefix"
+	CMAKE_ARGS -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR> -DBUILD_GMOCK=ON
+)
+
+add_library(gmock UNKNOWN IMPORTED)
+set_target_properties(gmock PROPERTIES
+	IMPORTED_LOCATION ${PROJECT_BINARY_DIR}/prefix/lib/libgmock.a
```
jbeder: Does this work on Windows?

test/CMakeLists.txt
```diff
@@ -1,26 +1,31 @@
...
+)
+
+find_package(Threads)
```
jbeder: What's this package?

test/CMakeLists.txt
```diff
@@ -36,9 +41,15 @@ add_executable(run-tests
...
+add_dependencies(run-tests googletest_project)
+
 set_target_properties(run-tests PROPERTIES
   COMPILE_FLAGS "${yaml_c_flags} ${yaml_cxx_flags} ${yaml_test_flags}"
 )
-target_link_libraries(run-tests yaml-cpp gmock)
+target_link_libraries(run-tests
+	yaml-cpp
+	gmock
+	${CMAKE_THREAD_LIBS_INIT})
```
jbeder: ... and this?

test/CMakeLists.txt
winterheart: You're right, it will fail. Fixed in next commit.

test/CMakeLists.txt
winterheart: On Linux Googletest requires pthread library linking. Here we find this library (optional, since for Windows we don't need them), and next ${CMAKE_THREAD_LIBS_INIT} is added as -lpthread linker option.

noahdesu:
I'd like to see this merged to get rid of deprecation warnings with newer versions of cmake.

gocarlos:
is there anything that still needs to be addressed?

jbeder:
FYI this broke the build when used as a git submodule (#612). I rolled this change back; if you want to try to resubmit a PR fixing the issue, please let me know!

PR #567, opened 2018-03-10
additional unittest for aliases

rhaschke:
I'm having trouble with accessing values in a yaml file like this:
```yaml
A: &DEFAULT
  str: string
  float: 3.1415
  int: 42
B: *DEFAULT
```
For some reason, in my application, accessing values `B[key]` fails, while `A[key]` succeeds. For this reason, I thought there might be an issue in yaml-cpp and implemented appropriate unittests with a minimal example. However, the unittests work as expected. Now idea, how to proceed now...
Nevertheless, the unittests might be interesting to you anyway. If haven't found particular tests for aliases/references.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -41,4 +41,5 @@ set_target_properties(run-tests PROPERTIES
 )
 target_link_libraries(run-tests yaml-cpp gmock)
 
-add_test(yaml-test ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/run-tests)
+add_test(NAME yaml-test COMMAND run-tests
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
--- a/test/integration/load_node_test.cpp
+++ b/test/integration/load_node_test.cpp
@@ -78,6 +78,36 @@ TEST(LoadNodeTest, IterateMap) {
   EXPECT_EQ(3, i);
 }
 
+TEST(LoadNodeTest, AliasAccess) {
+  const Node doc = Load("{A: &DEFAULT {str: string, int: 42, float: 3.1415}, B: *DEFAULT}");
+  const Node& A = doc["A"];
+  const Node& B = doc["B"];
+
+  // A and B have the same content
+  ASSERT_TRUE(A);
+  ASSERT_TRUE(A.IsMap());
+  ASSERT_TRUE(B);
+  ASSERT_TRUE(B.IsMap());
+
+  // A and B have the same content
+  std::map<std::string, std::string> values = {{"str", "string"}, {"float", "3.1415"}, {"int", "42"}};
+  for (YAML::const_iterator it = A.begin(); it != A.end(); ++it) {
+    const std::string& key = it->first.as<std::string>();
+    SCOPED_TRACE("key " + key);
+    const Node& a = A[key];
+    const Node& b = B[key];
+    EXPECT_TRUE(a);
+    EXPECT_TRUE(b);
+    EXPECT_TRUE(a.IsScalar());
+    EXPECT_TRUE(b.IsScalar());
+
+    // a and b should be identical
+    EXPECT_EQ(a.as<std::string>(), b.as<std::string>());
+    // ... and have the values given in the map
+    EXPECT_EQ(a.as<std::string>(), values[key]);
+  }
+}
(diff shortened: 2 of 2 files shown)


test/integration/load_node_test.cpp
```diff
@@ -78,6 +79,45 @@ TEST(LoadNodeTest, IterateMap) {
...
+    EXPECT_STREQ(a.as<std::string>().c_str(), values[key].c_str());
+  }
+}
+
+TEST(LoadNodeTest, AliasAccess) {
+  alias_access(Load("{A: &DEFAULT {str: string, int: 42, float: 3.1415}, B: *DEFAULT}"));
+}
+
+TEST(LoadNodeTest, AliasAccessStream) {
+  std::ifstream input("integration/alias.yaml");
```
jbeder: No need to have this separate file-based loading; it causes various problems like worrying about working directories, etc.

Instead, if you're worried about flow vs. block, then you can keep the alias_access function and have two tests; otherwise, one test is fine.

test/integration/load_node_test.cpp
```diff
@@ -78,6 +79,45 @@ TEST(LoadNodeTest, IterateMap) {
...
+    SCOPED_TRACE("key " + key);
+    Node a = A[key];
+    Node b = B[key];
+    EXPECT_TRUE(a);
+    EXPECT_TRUE(b);
+    EXPECT_TRUE(a.IsScalar());
+    EXPECT_TRUE(b.IsScalar());
+
+    // a and b should be identical
+    EXPECT_STREQ(a.as<std::string>().c_str(), b.as<std::string>().c_str());
```
jbeder: You can just expect that they're equal; you don't need to convert to c_str.

test/integration/load_node_test.cpp
```diff
@@ -78,6 +79,45 @@ TEST(LoadNodeTest, IterateMap) {
   EXPECT_EQ(3, i);
 }
 
+static void alias_access(Node&& doc) {
```
jbeder: If you do keep this function (see below):

- change its name to camel case
- put it in an anonymous namespace instead of being static
- make it take a const ref instead of an rvalue reference
- call it something that indicates it does some assertions, like VerifyAliasAccess

test/integration/load_node_test.cpp
rhaschke: Feel free to to adapt and use only what you need.
While tracking down the issue, I tried  variation along many directions. One was to load from an input...

jbeder: Do you mean you're abandoning this PR?

rhaschke: If you don't think, it's useful, yes.

jbeder: I do think it's useful, and I added some comments before approval.

test/CMakeLists.txt
```diff
@@ -41,4 +41,5 @@ set_target_properties(run-tests PROPERTIES
 )
 target_link_libraries(run-tests yaml-cpp gmock)
 
-add_test(yaml-test ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/run-tests)
+add_test(NAME yaml-test COMMAND run-tests
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```
jbeder: This shouldn't be necessary since you're not using files, right?

test/integration/load_node_test.cpp
```diff
@@ -78,6 +78,36 @@ TEST(LoadNodeTest, IterateMap) {
...
+  const Node& A = doc["A"];
+  const Node& B = doc["B"];
+
+  // A and B have the same content
+  ASSERT_TRUE(A);
+  ASSERT_TRUE(A.IsMap());
+  ASSERT_TRUE(B);
+  ASSERT_TRUE(B.IsMap());
+
+  // A and B have the same content
```
jbeder: I think A and B are actually the same node, so you can just check

ASSERT_TRUE(A.is(B));

or

ASSERT_TRUE(A == B);

(which is the same thing).

rhaschke:
Dear @jbeder,

please feel free to modify the pull request to your liking and merge the result. I myself abandoned the use of this library due to https://github.com/jbeder/yaml-cpp/pull/568.

Issue #208, opened 2015-03-30
Improve the syntax of the new API to prevent unintuitive behavior


GoogleCodeExporter:
```
One of the nice things about the new API is that it lets you easily modify 
nodes, e.g.:

foo["x"] = 5;

However, this currently causes the following behavior (see Issue 207):

Node a;
a = foo{"x"];
// ...
a = foo["y'];  // <-- this line is the same as foo["x"] = foo["y"], which 
modifies the document

This is unintuitive. It would be nice to support the former behavior but 
prevent the latter.
```

Original issue reported on code.google.com by `jbe...@gmail.com` on 7 Jun 2013 at 9:41

bobkocisko:
For anyone else who may have come here looking for a workaround so that you can assign node variables to one another while parsing a document and be sure that it won't modify any of the nodes in the process:

```
YAML::Node currentNode  = YAML::Load( "{ a: 1, b: 2, c: 3}" );

// currentNode = currentNode["a"]; // Would modify the document
currentNode.reset(currentNode["a"]); // Switches to the "a" node
```

So basically, when parsing use the .reset(otherNode) method to assign a node variable to another node.

However, there is one caveat:  If the variable already is invalid (`!currentNode.IsDefined()`) then the reset method will not change it.  So you have to be sure never to reset a variable to an undefined node if you plan to reset it in the future to a defined node.

P.S.  Does anyone know if there are any plans to change the reset method so the m_isValid field would be reset as well?

lollix89:
That's helpful @bobkocisko, i could not find the reset method mentioned anywhere!! How would you switch back to the root node once you are done iterating through the keys?

bobkocisko:
If I understand your question correctly, I have done something like this to switch back to the root node
```
YAML::Node currentNode;

YAML::Node rootNode = YAML::Load( "{ a: 1, b: 2, c: 3}" );
currentNode.reset(rootNode);
currentNode.reset(currentNode["a"]); // Switches to the "a" node
currentNode.reset(rootNode); // Switch back to the root node
```

PR #568, opened 2018-03-11
failing unittest: multiple assignment to Node variable with references in document

rhaschke:
I narrowed down my issue already mentioned in #567:
The failure is actually related to the assignment to the very same Node variable `value`.
If different Node variables are used each time, it works.
If no alias is involved, it works as well.

Please have a look to fix this issue. I already wasted a full day to narrow down that issue. Didn't look for reasons yet.

If I missed something and this kind of usage is illegal, please explain what's wrong.

### DIFF
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -239,11 +239,12 @@ inline void Node::Assign(char* rhs) {
 }
 
 inline Node& Node::operator=(const Node& rhs) {
-  if (!m_isValid || !rhs.m_isValid)
-    throw InvalidNode();
-  if (is(rhs))
-    return *this;
-  AssignNode(rhs);
+  AssignData(rhs);
+  return *this;
+}
+
+inline Node& Node::operator=(const NodeAlias& rhs) {
+  AssignNode(rhs.m_node);
   return *this;
 }
 
--- a/include/yaml-cpp/node/node.h
+++ b/include/yaml-cpp/node/node.h
@@ -26,6 +26,7 @@ struct iterator_value;
 }  // namespace YAML
 
 namespace YAML {
+class YAML_CPP_API NodeAlias;
 class YAML_CPP_API Node {
  public:
   friend class NodeBuilder;
@@ -81,6 +82,7 @@ class YAML_CPP_API Node {
   template <typename T>
   Node& operator=(const T& rhs);
   Node& operator=(const Node& rhs);
+  Node& operator=(const NodeAlias& rhs);
   void reset(const Node& rhs = Node());
 
   // size/iterator
@@ -132,6 +134,14 @@ class YAML_CPP_API Node {
   bool m_isValid;
   mutable detail::shared_memory_holder m_pMemory;
   mutable detail::node* m_pNode;
+};
+
+class NodeAlias {
+  friend class Node;
+public:
+  NodeAlias(const Node& node) : m_node(node) {}
+private:
+  const Node& m_node;
 };
 
 YAML_CPP_API bool operator==(const Node& lhs, const Node& rhs);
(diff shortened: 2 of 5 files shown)


jbeder:
This is WAI, although it's definitely counterintuitive. I'm sorry you ran into trouble.

This is a consequence of nodes being treated as identity, so subsequent assignment is actually mutating.

I've pondered some ways to improve the API to disallow this, but haven't taken any action. See #208

rhaschke:
IMHO it's not acceptable to keep an non-intuitive API. Obviously, I wasn't the only one yet, who tapped into this pitfall.
According to #208, I understand that you actually want to differentiate between assignment of a node `value` and assignment of a node `alias`. Why not making it explicit as follows:
``` c++
Node a;
Node b;
a = 5;  // value of a should be changed
a = b;  // a.AssignData(b)
a = NodeAlias(b); // a.AssignNode(b)
```
The following commit realizes this behaviour and thus resolves the ambiguity. However, obviously it breaks the semantics of the existing API.

There is a third assignment method: `a.reset(b)`. What is the difference of this w.r.t. the others? I don't see a need for a third semantics yet.
Finally, why do you need the `m_isValid` flag? Isn't the node automatically invalid, if `m_pNode == nullptr`?

rhaschke:
@jbeder Obviously my new commits to this branch are not considered by github anymore, probably because you closed this PR. Do you mind to reopen it?

jbeder:
I'll reopen this PR, but I'm not sure what you're trying to do; are you offering a patch that changes the API? Or are you trying to check in a failing test that highlights the aspect of the API that's at issue?

(Also, please consider your language when you write, "IMHO it's not acceptable to keep an non-intuitive API." This is an open source library maintained by one person (me!) in my spare time (which I don't have much of!). It's mean-spirited to declare that it's "not acceptable" for me not to devote significant amounts of time into change the API.)

rhaschke:
> Please consider your language.

I didn't want to offend you or your work on this project. I just stated that we shouldn't accept the current situation and with my recent commits, which became visible after re-opening this PR, I already suggested a potential solution that fixes the issue. I am myself maintainer of several actively used github projects and I really appreciate your work and your short reaction times!

Coming back to the discussion of the issue: Did you already had a look at my API changes? I'm sure I missed some details in the implementation - as I said I'm not very familiar with your source code. However, all the unittests pass and the semantic ambiguity of node and node alias is explicitly resolved. What do you think of this approach? I'm open for discussions on it.
Please, also shortly answer my questions in https://github.com/jbeder/yaml-cpp/pull/568#issuecomment-373937051 so that I can gain a better understanding of your source.

jbeder:
No worries, thank you for the apology.

To the point at hand: let me try to summarize your API change.

First change `operator =` so that

```
Node x = ...;
Node y = ...;
y = x;
```

just makes the *variable* `y` the same as `x`, but changes no internal state.

Then, add a new function `NodeAlias`, so that when you write

```
Node x = ...;
Node y = ...;
y = NodeAlias(x);
```

it *will* change internal state, so that `y` becomes an alias of `x`.

Is this summary correct?

If so, I think I do like the idea behind this change, although it is a breaking API change, so I'd like to think about it a little.

My instinct, however, is to actually remove the overload for `NodeAlias` (now that you're splitting it) and turn it into a function on `Node`, e.g., `SetAlias`. That seems clearer to me, and I didn't think of that before you suggested this patch.

What do you think?

rhaschke:
Yes, I think this is essentially what this PR suggests. However, NodeAlias is not a function, but a class constructor. 
Actually you already had two internal functions with these two different semantics: `AssignData` and `AssignNode` and the two assignment operators (for Node / NodeAlias as rhs) simply relay to them. Personally, I like the assingment version more than the function version, because it is more obvious what is assigned where.

jbeder:
OK, let me think about these options.

(As for the function v. class constructor, I called it a function because of the syntax (which is all users see). I believe we'd rather have a free function that delegates to a constructor rather than making it a real constructor, as you have now; but no need to change this PR yet, since I'm not sure what syntax we want.)

rhaschke:
May I ask my original questions again.
- What is the semantics of `Node::reset(Node)` in contrast to `Node::AssignNode()`?
- Why do you need the `m_isValid `flag? Can't you simply test for `m_pNode != nullptr`?

Issue #574, opened 2018-04-05
The library fails to compile in Visual studio 2017 with std++17


tankiJong:
Hi friends,

I was using the library with my project. Since I switch the project to C++17, error message pop up:
`1>error C4996: 'std::iterator<_Category,_Ty,_Diff,_Pointer,_Reference>::value_type': warning STL4015: The std::iterator class template (used as a base class to provide typedefs) is deprecated in C++17. (The <iterator> header is NOT deprecated.) The C++ Standard has never required user-defined iterators to derive from std::iterator. To fix this warning, stop deriving from std::iterator and start providing publicly accessible typedefs named iterator_category, value_type, difference_type, pointer, and reference. Note that value_type is required to be non-const, even for constant iterators. You can define _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
`
[there is a related post about this.](https://docs.microsoft.com/en-us/cpp/cpp-conformance-improvements-2017#annex_d)
Since it's error fix one, I have already done a [patch commit](https://github.com/tankiJong/yaml-cpp/commit/eb910ee9163935af3824bec870cebbd38fbc172d).

Could anyone help to review the commit? I can open a PR if it makes sense.

jbeder:
I'm happy to review the commit if you submit a PR.

PR #575, opened 2018-04-07
Update the iterator implementation for C++17

tankiJong:
Issue related: #574 

### Problem solved
std does not need user to derive `std::iterator` when they try to implement there own iterator, in C++17, this will cause compiler error unless `_SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING` is defined in the code, which supress the warning.

### What I did
I got rid of the base class for `iterator_base`, and declared several type defs, according to the standard requirement.

### compatibality
I only ran the test on my own machine(win10, VS 2017), all passed with several tests skipped(Which I think is configured with intention?). Since this is an platform indepentent issue, I did not test it on other platform and I expect it should be fine on other platform. I will be very glad to fix any problem if one shows up.

Thank you very much :D

### DIFF
--- a/include/yaml-cpp/node/detail/iterator.h
+++ b/include/yaml-cpp/node/detail/iterator.h
@@ -19,8 +19,7 @@ namespace detail {
 struct iterator_value;
 
 template <typename V>
-class iterator_base : public std::iterator<std::forward_iterator_tag, V,
-                                           std::ptrdiff_t, V*, V> {
+class iterator_base {
 
  private:
   template <typename>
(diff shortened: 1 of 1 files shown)


include/yaml-cpp/node/detail/iterator.h
```diff
@@ -37,8 +36,11 @@ class iterator_base : public std::iterator<std::forward_iterator_tag, V,
   };
 
  public:
-  typedef typename iterator_base::value_type value_type;
-
+  using iterator_category = std::forward_iterator_tag;
+  using value_type = V;
+  using difference_type = std::ptrdiff_t;
+  using pointer = V * ;
```
jbeder: Make sure to run clang-format; I think it would change this line.

include/yaml-cpp/node/detail/iterator.h
```diff
@@ -37,8 +36,11 @@ class iterator_base : public std::iterator<std::forward_iterator_tag, V,
...
+  using reference = V;
  public:
```
jbeder: Add a newline before this next section.

tankiJong:
Ping @jbeder
I just updated the format. If it's good, I can squash the commits. Thank you!

PR #580, opened 2018-04-29
Fix clang error: a space is required between '>>'

leolchat:
Full error: a space is required between consecutive right angle brackets (use '> >')


leolchat:
Actually, the issue is that the library doesn't have -std=c++11 in the package config file.

The issue here might be rare enough (old compiler like clang from macos + code not needing c++11 flag to compile but dependent on yaml-cpp).
Do you guys want a PR to fix the package config file?

PR #582, opened 2018-05-02
remove items from sequence

StarJiao:
based on pull request 471 [pullrequest 471](https://github.com/jbeder/yaml-cpp/pull/471)
allow users remove items from sequence and keep the type


util/demo.cpp
```diff
@@ -0,0 +1,45 @@
+#include "yaml-cpp/yaml.h"
```
jbeder: Instead of a new demo executable, unit tests are sufficient.

test/node/node_test.cpp
```diff
@@ -106,6 +130,14 @@ TEST(NodeTest, RemoveUnassignedNode) {
   EXPECT_EQ(0, node.size());
 }
 
+TEST(NodeTest, RemoveUnassignedNodeFromMap) {
```
jbeder: Similarly, please add a tests that verifies that if you have a map, where one of the keys is an integer, you can remove it and it's still a map.

include/yaml-cpp/node/detail/impl.h
```diff
@@ -56,6 +56,39 @@ struct get_idx<Key, typename std::enable_if<std::is_signed<Key>::value>::type> {
   }
 };
 
+template <typename Key, typename Enable = void>
+struct remove_idx {
+  static char remove(std::vector<node*>& /* sequence */, const Key& /* key */) {
```
jbeder: No need to comment out the variables.

include/yaml-cpp/node/detail/impl.h
```diff
@@ -129,23 +162,29 @@ inline node& node_data::get(const Key& key, shared_memory_holder pMemory) {
...
-
-  for (node_map::iterator it = m_map.begin(); it != m_map.end(); ++it) {
-    if (it->first->equals(key, pMemory)) {
-      m_map.erase(it);
-      return true;
-    }
-  }
+	if (m_type == NodeType::Sequence) {
+		char result = remove_idx<Key>::remove(m_sequence, key);
+		if (result == 0) return false;
```
jbeder: Add curly braces around all "if" statements, even if they're one line.

(The codebase isn't quite consistent there, but I'm trying to enforce it for new code.)

include/yaml-cpp/node/detail/impl.h
jbeder: Why does this return a 'char'? It looks like in all implementations, it either returns 0 or 1.

include/yaml-cpp/node/detail/impl.h
```diff
@@ -129,23 +162,29 @@ inline node& node_data::get(const Key& key, shared_memory_holder pMemory) {
...
+		if (result == 1) return true;
+		convert_to_map(pMemory);
```
jbeder: Can this ever happen? (See above comment.)

include/yaml-cpp/node/detail/impl.h
```diff
@@ -129,23 +162,29 @@ inline node& node_data::get(const Key& key, shared_memory_holder pMemory) {
...
+	}
+	if (m_type == NodeType::Map) {
+
```
jbeder: Extra newline.

StarJiao:
@jbeder Thank you so much and I'm honoured by your reviewing .To be honest, this is the first time to create a pull request and there is many rules i need learn.  I am not familiar with C++ . Haha. Thanks again and I will fix all these mistake by your review.

jbeder:
No problem! Thanks for contributing!

StarJiao:
@jbeder Done

PR #583, opened 2018-05-02
Use target-based include_directories to enable modern CMake usage

stotko:
This PR adds support for modern target-based linking for CMake 2.8.12 and newer where the include directories are automatically propagated without further commands. The old fashioned way of linking (with `(target_)include_directories`) is still supported across all CMake versions for users relying on it.


timrae:
Could you please make a new release including this commit? It feels so wrong manually using the variable.

PR #585, opened 2018-05-10
Allow customization via inheritance of Node

### LINKED ISSUE #584: Make customization through inheritance easier
I have a particular use case where it would be very useful to implement a custom subclass of `Node`. However, this is currently very difficult, or maybe impossible. At a minimum there are two things that would be extremely helpful:

* Change `private` members of `Node` to `protected` in order to facilitate inheritance
* Make `NodeBuilder` part of the public API to allow it to be inherited. In order for this to work, its `private` members would also need to become `protected`.

Is there any chance these changes are acceptable? If so, I may have a few spare cycles to submit a PR.

drdavella:
This PR makes it possible for clients to subclass `Node`. It includes the following changes:

* change `private` members of `Node` to `protected` to support inheritance
* make `NodeBuilder` part of the public API to facilitate creation of custom nodes
* change `private` members of `NodeBuilder` to `protected` to support inheritance

No functionality was added/changed/deleted for this PR.

This resolves #584


drdavella:
Hi, just a gentle poke. I'm hoping this PR is simple and reasonable enough to be accepted 😃.

include/yaml-cpp/node/node.h
```diff
@@ -128,7 +128,7 @@ class YAML_CPP_API Node {
   void AssignData(const Node& rhs);
   void AssignNode(const Node& rhs);
 
- private:
+ protected:
```
drdavella: Rather than making these attributes `protected` (and those below in `nodebuilder.h`), it might be a better idea to add `protected` getters (which of course I'm willing to do, if the general thrust of this PR is reasonable to you).

jbeder:
I commented on the Issue: Sorry, I don't think inheritance is the right way to go for customization. If there are particular things you'd like to do that you can't now, I'd rather address them.

drdavella:
@jbeder Sorry I missed the comment on the issue.

I'll try to spare you the nitty gritty details, but basically I was hoping to use your library as the basis for the C++ implementation of the [Advanced Scientific Data Standard (ASDF)](https://asdf-standard.readthedocs.io/en/latest/index.html) (Python reference implementation [here](https://github.com/spacetelescope/asdf)).

ASDF uses YAML as its metadata representation, but the files also contain binary data blocks with numerical array data. I need each node in the YAML tree to have a reference to the top-level node so that it can read data blocks if necessary when calling `.as`. I was able to achieve this with the modification that I proposed in this PR. Unless you are willing to support references to the top of the tree within `yaml-cpp` itself, I'm going to have to settle for a significantly more awkward and less useable approach.

I understand that you're under no obligation to accept PRs, and I appreciate the work you have done for the community with this project. However, I thought my proposed change was rather modest and reasonable, and I'm not sure why allowing clients to take advantage of inheritance for their own purposes should be so objectionable.

jbeder:
Thanks for the explanation!

First, the high-level - allowing inheritance is dangerous for an API (it expands the surface area quite a bit), so my preference in general is to restrict that, unless the API is explicitly intended to be for inheritance.

To the specifics - this is a really interesting use case; without knowing more about what the `.as()` function actually does, my instinct is that either:

(a) you can add a reference to the root node (see anchors/aliases) in the nodes that need it
(b) you can make your own function that has signature `(YAML::Node node, YAML::Node root) -> some-type)` and call that instead of `.as()`

drdavella:
> unless the API is explicitly intended to be for inheritance

This is what `protected` indicates, does it not?

My current usage is something like the following:
```c++
Asdf::AsdfFile asdf("example.asdf");
Asdf::Node tree = asdf.get_tree();
auto array = tree["my_data"].as<Asdf::NDArray<int>>();
```

The reference to the top-level file object is passed through to every node so that the memory mapped binary block is accessible when accessing nodes that represent arrays.

I can't use `(a)` because that would require modification to the file format, which would require changes to the ASDF standard. I guess I'll have to go with `(b)` and come up with some other API.

jbeder:
There's more to design an API for inheritance than using `protected`; you need to make sure that no matter what subclasses do, all invariants are always preserved.

I think going with (b) is probably best.

drdavella:
Okay, fair enough. Thanks for the discussion and for implementing a very useful library.

Issue #624, opened 2018-09-04
Option to disable installing


VelocityRa:
The following line
https://github.com/jbeder/yaml-cpp/blob/d0da14404ef7cc960ef01bbf97775ead08d3c477/CMakeLists.txt#L325-L330

results in the following error:
```
CMake Error at 3rdparty/yaml-cpp/CMakeLists.txt:325 (file):
  file RELATIVE_PATH must be passed a full path to the directory:
  install/CMake
```
every time a cmake file in my project changes. I need to do a complete cache clean for it to be resolved (and then it happens again after I change something). This happens on Windows only and for 2 different projects that I'm aware of.

One option would be to optionally disable yaml-cpp installation, or maybe provide an alternate code path if the path passed to that line is absolute.

jbeder:
I'm not sure I understand this bug.

1) You said "This happens on Windows only and for 2 different projects that I'm aware of." Does this mean "Windows" and "2 different non-Windows projects"? Or "Windows for 2 different projects", and doesn't happen for "Windows and a third project"?

2) Does this happen when you try installing? Or building?

In general, it's better to reduce the number of options; what else can we do to fix this problem for you?

VelocityRa:
- I know two cross-platform projects that use yaml-cpp.
- Both of these projects experience the issue.
- It only happens on Windows
- It happens when [building](https://github.com/Vita3K/Vita3K/blob/be58d2856caeb09de3c709db195e3f8036d7554e/src/external/CMakeLists.txt#L109-L112), we don't install it anywhere.

>In general, it's better to reduce the number of options; what else can we do to fix this problem for you?

I'm not sure I even understand the problem, so I'm probably not the person to ask this, but I _think_ that (as the title says), making yaml-cpp installation optional with something like a `YAML_NO_INSTALL ` CMake option, would mean that the erroneous line above is not executed, nullifying the issue.

kukkamario:
CMAKE_INSTALL_PREFIX is relative (in this case usually just "install") which causes "${CMAKE_INSTALL_PREFIX}/${INCLUDE_INSTALL_ROOT_DIR}" to be a relative path. file(RELATIVE_PATH) only works with absolute paths so it fails with an error.

[...]

PR #625, opened 2018-09-05
Make install target generation optional (#624)

kukkamario:
- Add YAML_CPP_INSTALL option which defaults to ON
  - Allows disabling install target generation
- Fix configuring when CMAKE_INSTALL_PREFIX is relative path
  - Fixes issue #624


VelocityRa:
Any reason this isn't merged yet?

jbeder:
Isn't the option redundant? Can't you just not run "make install"?

kukkamario:
Having optional install is useful when yaml-cpp is included in another project by using (git submodule +) `add_subdirectory`. Because there is always only one install target, running the `make install` for the main project will also install yaml-cpp which isn't always desirable (eg. when yaml-cpp is only internal dependency and the users of the main project don't need yaml-cpp headers).

VelocityRa:
Having to deal with this bug is infuriating, is there any particular reason this is still not merged?

Such a CMake option is present in many (most?) libraries that support CMake, I really don't understand what the problem is.

stotko:
I'm not sure whether this is the right solution. This looks more like a workaround (see [my comment here](https://github.com/jbeder/yaml-cpp/issues/624#issuecomment-422742512)).

kukkamario:
Relative CMAKE_INSTALL_PREFIX shouldn't occur normally so I removed commit that implemented workaround for that.

VelocityRa:
Bump...

PR #629, opened 2018-09-28
Update README.md

traderggl:
Build works if you navigate into yaml-cpp-master NOT yaml-cpp-master/src which is implied from instructions.


jbeder:
Ah, I see the confusion. However, `yaml-cpp-master` is not necessarily the name of the top-level directory. Typically, we say `source` to indicate the directory that you check out the source code to, so I think I'll leave as-is.

I appreciate the consideration.

Issue #605, opened 2018-07-04
CMAKE_CXX_FLAGS_xxx broken with Intel on MSVC


maartendb:
The CMakeLists.txt overwrites the CMAKE_CXX_FLAGS_xxx (line 174) when using Intel on MSVC. As a result, the Debug configuration will be using /MD instead of /MDd and cause linking errors with binaries compiled with /MDd.

These two solutions will work:
- append instead of overwrite the flags (e.g set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG -g")
- put these four set(CMAKE_CXX_FLAGS_xxxx ....) lines inside an if(NOT WIN32) block.

jbeder:
Probably the first?

maartendb:
I think the first is better, because it allows passing these flags with additional  compiler options to cmake from the outside, for all platforms. The second would result in MSVC-with-intel matching the current behaviour of MSVC-with-native.

SGSSGene:
Looks like #635 fixed this. Please reopen this ticket or a new one if any issues persist.

PR #635, opened 2018-10-23
Don't stomp on build flags

joelfrederico:
Let CMake handle the default optimizations for various configurations. We don't need to override them. In fact, overriding them makes it impossible for users to override them themselves.


jbeder:
See #605, I think; probably just extend the existing ones, not remove these blocks entirely?

joelfrederico:
Ah here I’ll lay out my analysis and the full best practices as I know them.

- They’re mostly redundant. For some reason, they don’t fully optimize in release mode, -O2 instead of -O3. So they conflict with the default flags (they are appended to the default flags) and also prevent end user customization.
- These flags are intended for the end user. If you want to change the default flags, you should use CMAKE_\<LANG>\_FLAGS_\<CONFIG>_INIT. (
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG_INIT.html)
- AFAICT, these flags have existed for 8 years according to git blame. If there was a reason for them before, I would be surprised if it is still true.
- Best practices would be to leave these flags alone entirely. Your project may be included via add_subdirectory and the flags should reflect the master project. Cmake normally handles this itself. If you override Cmake’s way of handling them, then users have to figure out your unique system and make it work the way they intend. Generally speaking, it would be really odd to have different compiler flags for different sections of a project, and in some cases this can lead to bugs.
- If you do need to override, say if -O3 runs into a compiler bug and that’s the only workaround, best practices would be to do string substitution on a variable. Users may change these variables and stomping on variables removes tools from users.
- As they stand now, they don’t work with FetchContent on Windows with clang-cl.exe.

jbeder:
Thanks for the explanation. You're probably right that these no longer are needed. I vaguely remember adding them, but I'm not sure why.

PR #644, opened 2018-11-23
Export only YAML_CPP_API-tagged symbols on Linux.

RAOF:
Mark classes tagged with YAML_CPP_API with visibility("default"), to explicitly export them, and then hide everything else via `-fvisibility=hidden`.

This reduces the number of exported symbols by more than 50%, which has a small application start-up time and memory useage benefit. It also makes it more feasible to audit the set of exported symbols in automated ABI compatibility systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,14 @@ if(WIN32)
 	endif()
 endif()
 
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+	set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+	if (POLICY CMP0063)
+		cmake_policy(SET CMP0063 NEW)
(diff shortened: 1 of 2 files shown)


RAOF:
In addition to running the unit-tests I've also built all of the packages in Ubuntu which depend on yaml-cpp against the it. All of them (at least the ones which don't fail for other reasons, like not using C++11) build fine against this, so those symbols are really unused.

Zingam:
@RAOF For Windows the classes shouldn't be tagged with YAML_CPP_API but the public methods. Will that conflict with Linux in any way?

vmiheer:
#958 seems related and especially I think letting cmake handle this is good idea. https://github.com/jbeder/yaml-cpp/pull/958#issuecomment-759639237

Issue #272, opened 2015-03-30
unable to round trip floating point numbers


GoogleCodeExporter:
```
The code used to control the precision of floating point numbers is too 
restrictive, the limit should be defined more like this

(2 + std::numeric_limits<float>::digits * 3010/10000)

see

http://www2.open-std.org/JTC1/SC22/WG21/docs/papers/2005/n1822.pdf

not sure if the default should be changed, for consistency I would
```

Original issue reported on code.google.com by `Kevin.J....@gmail.com` on 30 Jan 2015 at 1:57

Attachments:
- [yaml-cpp.precision.diff](https://storage.googleapis.com/google-code-attachments/yaml-cpp/issue-272/comment-0/yaml-cpp.precision.diff)

KevinJW:
Looks like #470 might also address this.

Kevin

jbeder:
Fixed by #649.

Issue #197, opened 2015-03-30
allow for full precision floating point serialization


GoogleCodeExporter:
```
What steps will reproduce the problem?
1.output a floating point number such as 1 + eps

What version of the product are you using? On what operating system?
0.3.0, affects 0.5.0 too

What is the expected output? What do you see instead?
A number that when parsed yields the same value that was converted, instead a 
different number is yielded.

See this or many other threads for detail:
http://stackoverflow.com/questions/4738768/printing-double-without-losing-precis
ion

The executive summary is that digits10 is truncating the fractional digits a 
floating point number has and the round-trip output-input will be different in 
many cases.  You should instead clamp at digits10+1 and arguably set the 
defaults to be fully preserving as well to minimize surprises.  Patch attached 
that does this.  Alternatively use/implement David Gay's algorithm to convert 
to the shortest string which yields the same value could be used to keep yaml 
files small, preserve readability and what the computer actually had in memory 
at time of serialization.

http://www.ampl.com/REFS/rounding.pdf
David Gay's implementation:
http://www.netlib.org/fp/dtoa.c
http://www.netlib.org/fp/g_fmt.c

Apparently it is used all over the place (python, numerous web-browsers and 
popular databases) so it seems this is the defacto way to do it.

You may also want to use C/C++ strtod on input, which again likely borrows much 
of David's code as iostreams don't handle numerous cases, in particular 
denormed values (you likely already handle nan/inf)

http://www.cplusplus.com/reference/cstdlib/strtod/
http://www.cplusplus.com/reference/string/stod/

```

Original issue reported on code.google.com by `nev...@gmail.com` on 10 Apr 2013 at 1:06

Attachments:
- [yaml-cpp-precision-fix.patch](https://storage.googleapis.com/google-code-attachments/yaml-cpp/issue-197/comment-0/yaml-cpp-precision-fix.patch)

jbeder:
Fixed by #649.

PR #649, opened 2018-12-11
Bugfix: fixing float precision

### LINKED ISSUE #229: reading max double, throws bad conversion
```
What steps will reproduce the problem?

#include <fstream>
#include <iostream>
#include <yaml-cpp/yaml.h>
using namespace std;

int main()
{
  double doubleVal;
  float floatVal;
  ofstream outYaml;
  YAML::Node writer;
  YAML::Node reader;

  // write
  writer["float"] = numeric_limits<float>::max();
  writer["double"] = numeric_limits<double>::max();
  outYaml.open("test.yaml");
  outYaml << writer;
  outYaml.close();

  // read
  reader = YAML::LoadFile("test.yaml");
  if (reader["float"])
  {
    floatVal = reader["float"].as<float>();
    cout << "float " << floatVal << endl;
  }
  if (reader["double"])
  {
    doubleVal = reader["double"].as<double>(); // throws bad conversion
    cout << "double " << doubleVal << endl;
  }
  return 0;
}


What is the expected output? What do you see instead?
reading max float is ok, but reading max double throws bad conversion

What version of the product are you using? On what operating system?
Using yaml-cpp 0.5.1 on Ubuntu 12.04 64-bit

Please provide any additional information below.


```

Original issue reported on code.google.com by `lucky13...@gmail.com` on 13 Dec 2013 at 7:36

SGSSGene:
This fixes the round trip of floats/doubles into YAML::Node and back.

This is very similar to #470 (why is this not merged?). The big difference to #470 is that this PR changes the  unit tests that are not testing floating point capabilities (or do it in a wrong way) to "stable floating point numbers" that can be represented exact in memory.

This also fixes PR #469. (Same goes for #272 and #197)

This probably also fixes #229

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -93,7 +93,7 @@ struct convert<_Null> {
   struct convert<type> {                                                 \
     static Node encode(const type& rhs) {                                \
       std::stringstream stream;                                          \
-      stream.precision(std::numeric_limits<type>::digits10 + 1);         \
+      stream.precision(std::numeric_limits<type>::max_digits10);         \
       stream << rhs;                                                     \
       return Node(stream.str());                                         \
     }                                                                    \
--- a/src/emitterstate.cpp
+++ b/src/emitterstate.cpp
@@ -24,8 +24,8 @@ EmitterState::EmitterState()
   m_seqFmt.set(Block);
   m_mapFmt.set(Block);
   m_mapKeyFmt.set(Auto);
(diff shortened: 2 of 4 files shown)


SGSSGene:
@jbeder any clue what is broken with the CI? it didn't even run the test code

jbeder:
No idea. Looks like it might be a network failure. I restarted it.

SGSSGene:
Thanks for restarting it. Phat is the process to get this PR accepted?

SGSSGene:
Very cool, thank you

PR #656, opened 2018-12-28
Better error messages

### LINKED ISSUE #423: Exception for missing value is non-descriptive
Using the example given in the tutorial: 

``` c++
YAML::Node config = YAML::LoadFile("config.yaml");

if (config["lastLogin"]) {
  std::cout << "Last logged in: " << config["lastLogin"].as<DateTime>() << "\n";
}

const std::string username = config["username"].as<std::string>();
const std::string password = config["password"].as<std::string>();
login(username, password);
config["lastLogin"] = getCurrentDateTime();

std::ofstream fout("config.yaml");
fout << config;
```

and the following YAML file:

``` yaml
usernaem: "user" # typo in key
password: "password
```

generates the exception 

> terminate called after throwing an instance of 'YAML::InvalidNode'
>  what():  invalid node; this may result from using a map iterator as a sequence iterator, or vice-versa

Given any half-complex setup, this is impossible to debug without pulling up GDB, so all code is filled with manual error-checking for every field.

My suggestion given the current API would be to just add an stl-style `YAML::Node at(T key)`, which throws on failure, allowing the error to contain the key (at least in the case of string/int), and the Mark() for the parent node.

bedapisl:
Fixes #423.
Exceptions `InvalidNode` and `BadSubscript` print the first invalid key, if it is printable.
For example given `config.yaml`:
```
first:
    second: 1
    third: 2
```
this code:
```
YAML::Node config = YAML::LoadFile("config.yaml");
int x = config["first"]["second"]["missing_key"].as<int>();
```
prints exception
```
operator[] call on a scalar (key: 'missing_key')
```
and this code:
```
const YAML::Node config = YAML::LoadFile("config.yaml");
auto n = config["first"]["missing_key"].as<int>();
```
prints exception:
```
invalid node - first invalid key: 'missing_key'
```


aftersought:
@bedapisl I wonder if something similar could be done to `TypedBadConversion` exception for the missing keys (see the question [here](https://stackoverflow.com/questions/54208964/add-missing-key-to-exception-message-when-failing-to-convert)).

bedapisl:
@aftersought I don't know how to do it for `TypedBadConversion`. But if you make the config `const`, then your code should throw `InvalidNode` exception which  (with this pull request) prints the missing key. See the second example in my previous comment.

aftersought:
@bedapisl Fair enough, thank you. 

Can only reiterate that this is an important commit as exceptions should fully explain the problem, at least in the typical use cases (such as misspelled key in config file).

aftersought:
@bedapisl One more thing -- apparently this is still not verbose enough for the nested configs. For example:

```yaml
students:
- name: Mike
  age: 20
- name: Ann
  age: 21
- Name: Pete
  ag: 28
```

Trying to parse the third node (`student["age"].as<int>()`) under your implementation will throw `what():  invalid node - first invalid key: 'age'`, however this is ambiguous as there are many potential places where the user could make a typo in the key `age`. Any ideas how we could support the user here? Ultimately what the user expects is the line in the config file where the parent node (`student`) resides (in line with tinyxml and their ilk).

bedapisl:
I agree that in this case the error message is not good enough. But in my opinion getting better error message here would require much bigger changes than I did.

bedapisl:
Hello @jbeder, can you review this?

include/yaml-cpp/node/node.h
```diff
@@ -130,6 +132,7 @@ class YAML_CPP_API Node {
 
  private:
   bool m_isValid;
+  std::string m_invalidKey;
```
jbeder: Add a comment, that this is the string representation of an invalid key, if this node is invalid.

bedapisl:
Thanks for the review, tests added.

include/yaml-cpp/node/impl.h
```diff
@@ -368,24 +374,35 @@ inline typename to_value_t<T>::return_type to_value(const T& t) {
 }
 }
 
+template<typename Key>
+std::string key_to_string(const Key& key) {
+  std::stringstream ss;
+  if (is_streamable<std::stringstream, Key>::value) {
```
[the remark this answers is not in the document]
bedapisl: Yes, you were right, it didn't work correctly.

include/yaml-cpp/node/node.h
bedapisl: Done.

jbeder: Style nit: Please capitalize the first letter of the comment, and put a period at the end.

test/integration/error_messages_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+  try {                                                            \
+    statement;                                                     \
+  } catch (const exception_type& e) {                              \
+    EXPECT_EQ(e.msg, message);                                     \
+  }
+
+namespace YAML {
+namespace {
+
+TEST(ErrorMessageTest, Ex9_1_BadSubscriptErrorMessage) {
```
jbeder: Inline the test's YAML for these specific feature tests rather than reusing the spec. (It's harder to understand when you have to go dig for these.)

include/yaml-cpp/exceptions.h
```diff
@@ -114,6 +114,35 @@ inline const std::string KEY_NOT_FOUND_WITH_KEY(
...
+  stream << BAD_SUBSCRIPT << " (key: '" << key << "')";
+  return stream.str();
+}
+
+inline const std::string INVALID_NODE_WITH_KEY(const std::string& key) {
+  std::stringstream stream;
+  if (key.empty()) {
+    return INVALID_NODE;
+  }
+  stream << "invalid node - first invalid key: '" << key << "'";
```
jbeder: I'd use a semicolon instead of dash, and double-quote inside the string

"invalid node; first invalid key: \"" << key << "\"";

(ditto for the quotes in the above examples)

include/yaml-cpp/node/impl.h
```diff
@@ -368,24 +374,45 @@ inline typename to_value_t<T>::return_type to_value(const T& t) {
 }
 }
 
+template<typename Key, bool Streamable>
+struct key_to_string_impl {
```
jbeder: This is more of a streamable_to_string, right? I think it should go in the traits file instead, and rename.

include/yaml-cpp/traits.h
```diff
@@ -100,4 +103,17 @@ template <class Cond, class T = void>
 struct disable_if : public disable_if_c<Cond::value, T> {};
 }
 
+template <typename S, typename T>
+class is_streamable {
+  template <typename SS, typename TT>
```
jbeder: Flip these around, so the public methods are before the private ones.

bedapisl: I changed it to struct and made everything public, because `is_streamable.test` must be declared before `is_streamable.value`.

include/yaml-cpp/traits.h
```diff
@@ -104,16 +106,30 @@ struct disable_if : public disable_if_c<Cond::value, T> {};
...
+    std::stringstream ss;
+    ss << key;
+    return ss.str();
+  }
+};
+
+template<typename Key>
+struct streamable_to_string<Key, false> {
+  static std::string impl(const Key&) {
+    return std::string();
```
jbeder: A thought: maybe write "<unprintable>" here rather than an empty string?

test/specexamples.h
```diff
@@ -843,9 +843,4 @@ const char *ex8_22 =
     " - nested\n"
     "mapping: !!map\n"
     " foo: bar\n";
-
-const char *ex9_1 =
```
jbeder: Accidental deletion?

include/yaml-cpp/traits.h
bedapisl: Yeah, that looks better.

bedapisl:
Hello @jbeder, would you look at this again please? I was getting some warnings, that members of Node are initialized in wrong order, so I adjusted the constructors of Node. I can remove these changes if you don't like them, they are not necessary at all.

PR #659, opened 2019-01-05
Fix GCC-types flags appended on Clang-cl

pent0:
Clang-cl actually uses MSVC compile flags, not GCC-style compile flags.

Some GCC compile flags got mixed in clang-cl and makes yaml-cpp not linkable in both Debug and Release mode.

This PR strengthens the check, not appending the GCC flags when we are at CL mode.


pent0:
Thanks for merging :)

PR #668, opened 2019-02-06
Introduce functor detail::node::less to sort set of nodes*

c4pQ:
It allows to keep order of nodes at output.

### DIFF
--- a/include/yaml-cpp/node/detail/node.h
+++ b/include/yaml-cpp/node/detail/node.h
@@ -13,10 +13,16 @@
 #include "yaml-cpp/node/ptr.h"
 #include "yaml-cpp/node/type.h"
 #include <set>
+#include <atomic>
 
 namespace YAML {
 namespace detail {
 class node {
+ private:
+  struct less {
+    bool operator ()(const node* l, const node* r) {return l->m_index < r->m_index;}
+  };
+
  public:
   node() : m_pRef(new node_ref), m_dependencies{} {}
   node(const node&) = delete;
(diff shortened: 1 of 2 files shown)


c4pQ:
@jbeder could you please say something about this PR?

include/yaml-cpp/node/detail/node.h
```diff
@@ -160,8 +166,10 @@ class node {
 
  private:
   shared_node_ref m_pRef;
-  typedef std::set<node*> nodes;
+  typedef std::set<node*, less> nodes;
   nodes m_dependencies;
+  size_t m_index;
+  static size_t m_amount;
```
jbeder: This isn't threadsafe.

c4pQ: How about this?

c4pQ: Almost an anniversary.

de-vri-es:
This would be very nice to have merged in mainline. Any change this could be accepted and released?

include/yaml-cpp/node/detail/node.h
de-vri-es: To be fair, `std::set` also isn't thread-safe.

c4pQ: de-vri-es, that's true. Any ideas?

de-vri-es: I don't think it makes sense to try to make `Node` thread safe. I think it is better to just document the fact that it isn't and require the user to do synchronization. Doing it inside `Node` would add a lot of complexity and hurt performance, also in cases where it isn't needed.

c4pQ:
Rebased

de-vri-es:
Nice, thanks for the merge :)

PR #674, opened 2019-02-13
Fixed building with Ninja on Windows

Zingam:
...and updated gtest


test/CMakeLists.txt
```diff
@@ -7,20 +7,26 @@ if(MSVC)
...
 		-Dgtest_force_shared_crt=ON
+	BUILD_BYPRODUCTS
+		"<INSTALL_DIR>/lib/${CMAKE_STATIC_LIBRARY_PREFIX}gmock${CMAKE_DEBUG_POSTFIX}${CMAKE_STATIC_LIBRARY_SUFFIX}"
 )
 
 add_library(gmock UNKNOWN IMPORTED)
 set_target_properties(gmock PROPERTIES
     IMPORTED_LOCATION
-    ${PROJECT_BINARY_DIR}/test/prefix/lib/${CMAKE_STATIC_LIBRARY_PREFIX}gmock${CMAKE_STATIC_LIBRARY_SUFFIX}
+		${PROJECT_BINARY_DIR}/test/prefix/lib/${CMAKE_STATIC_LIBRARY_PREFIX}gmock${CMAKE_DEBUG_POSTFIX}${CMAKE_STATIC_LIBRARY_SUFFIX}
```
jbeder: This seems surprising. Why does it need this postfix? Why debug, given that this line isn't specific to debug?

Zingam: Gtest sets the postfix.

Zingam: For my projects I use automatic download of gtest and compilation at a configuration time but I am not sure if you'd like that.
Also the configuration and the build fail in VS2019 without the above changes.

Zingam:
Gtest fails to compile:
`/home/travis/build/Zingam/yaml-cpp/test/gtest-1.8.1/googletest/src/gtest-typed-test.cc:109:5: error: ‘Abort’ is not a member of ‘testing::internal::posix’

Zingam:
Looks like we need a C++11 compiler: https://github.com/google/googletest/issues/1804
Should I up the gcc version?

test/CMakeLists.txt
jbeder: (I prefer to vendor the dep here.)

So gtest sets the postfix, but only for debug and ninja builds?

jbeder:
Yes, please feel free to upgrade the compiler so that it compiles. It looks like googletest uses 4.9. That's fine with me.

PR #686, opened 2019-03-11
Add more compile time diagnostics for g++ and clang++

TedLyngmo:
Add compilation flags: -Wshadow -Weffc++ -pedantic -pedantic-errors

* Delete implicit copy & move constructors & assignment operators
  in classes with pointer data members.
* An exception to the above: Add default copy & move constructors &
  assignment operators for the Binary class.
* Convert boolean RegEx operators to binary operators.
* Initialize all members in all classes in ctors.
* Let default ctor delegate to the converting ctor in
  Binary and RegEx
* Don't change any tests except regex_test (as a result of the change
  to binary operators).

Note: [bugzilla 1544675](https://bugzilla.redhat.com/show_bug.cgi?id=1544675) makes -Weffc++ report a false positive in "include/yaml-cpp/node/impl.h".


src/contrib/graphbuilderadapter.h
```diff
@@ -50,8 +54,8 @@ class GraphBuilderAdapter : public EventHandler {
   struct ContainerFrame {
     ContainerFrame(void* pSequence)
         : pContainer(pSequence), pPrevKeyNode(&sequenceMarker) {}
-    ContainerFrame(void* pMap, void* pPrevKeyNode)
-        : pContainer(pMap), pPrevKeyNode(pPrevKeyNode) {}
+    ContainerFrame(void* pMap, void* PrevKeyNode)
```
jbeder: Typo? Looks like you lost the 'p' prefix.

src/directives.cpp
```diff
@@ -1,11 +1,11 @@
 #include "directives.h"
 
 namespace YAML {
-Directives::Directives() {
-  // version
-  version.isDefault = true;
-  version.major = 1;
-  version.minor = 2;
+Directives::Directives() : version{true, 1, 2}, tags{} {
+  /* Version:
```
jbeder: Why this comment?

src/exp.h
```diff
@@ -33,45 +33,45 @@ inline const RegEx& Tab() {
   return e;
 }
 inline const RegEx& Blank() {
-  static const RegEx e = Space() || Tab();
+  static const RegEx e = Space() | Tab();
```
jbeder: Why switch to overload | instead of ||?

src/regex_yaml.cpp
```diff
@@ -2,17 +2,19 @@
...
+RegEx::RegEx() : RegEx(REGEX_EMPTY) {}
 
-RegEx::RegEx(char ch) : m_op(REGEX_MATCH), m_a(ch) {}
+RegEx::RegEx(char ch) : m_op(REGEX_MATCH), m_a(ch), m_z(0), m_params{} {}
 
-RegEx::RegEx(char a, char z) : m_op(REGEX_RANGE), m_a(a), m_z(z) {}
+RegEx::RegEx(char a, char z) : m_op(REGEX_RANGE), m_a(a), m_z(z), m_params{} {}
 
-RegEx::RegEx(const std::string& str, REGEX_OP op) : m_op(op) {
+RegEx::RegEx(const std::string& str, REGEX_OP op) : m_op(op), m_a(0), m_z(0), m_params(str.begin(), str.end()) {
```
jbeder: The single-arg constructors should really be marked explicit (can you do this?), so this change will have to be reverted.

src/singledocparser.cpp
```diff
@@ -166,10 +168,10 @@ void SingleDocParser::HandleBlockSequence(EventHandler& eventHandler) {
 
     // check for null
     if (!m_scanner.empty()) {
-      const Token& token = m_scanner.peek();
-      if (token.type == Token::BLOCK_ENTRY ||
-          token.type == Token::BLOCK_SEQ_END) {
-        eventHandler.OnNull(token.mark, NullAnchor);
+      const Token& tnullcheck = m_scanner.peek();
```
jbeder: Why change the name?

src/emitterstate.cpp
```diff
@@ -6,27 +6,32 @@
...
-  m_preCommentIndent.set(2);
-  m_postCommentIndent.set(1);
-  m_seqFmt.set(Block);
-  m_mapFmt.set(Block);
-  m_mapKeyFmt.set(Auto);
-  m_floatPrecision.set(std::numeric_limits<float>::max_digits10);
-  m_doublePrecision.set(std::numeric_limits<double>::max_digits10);
-}
+      m_docCount(0)
+{}
```
jbeder: I think clang-format will put the {} on the same line as the last initializer. Can you verify? (Here and elsewhere in this PR.)

include/yaml-cpp/node/impl.h
```diff
@@ -15,7 +15,7 @@
 #include <string>
 
 namespace YAML {
-inline Node::Node() : m_isValid(true), m_pNode(NULL) {}
+inline Node::Node() : m_isValid(true), m_pMemory(nullptr), m_pNode(NULL) {}
```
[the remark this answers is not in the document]
TedLyngmo: Ok, will do!

include/yaml-cpp/contrib/anchordict.h
```diff
@@ -32,7 +32,7 @@ class AnchorDict {
   T Get(anchor_t anchor) const { return m_data[anchor - 1]; }
 
  private:
-  std::vector<T> m_data;
+  std::vector<T> m_data{};
```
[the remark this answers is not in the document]
TedLyngmo: Ok, I wasn't sure if adding a ctor would be prefered, but now I know :-)

include/yaml-cpp/node/impl.h
```diff
@@ -202,6 +202,15 @@ inline Node& Node::operator=(const T& rhs) {
   return *this;
 }
 
+inline Node& Node::operator=(const Node& rhs) {
```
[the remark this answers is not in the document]
TedLyngmo: I placed it right after the copy assignment template when trying to fix the -Weffc++ false positive and grew to like it there (even though I didn't include a fix for the false positive, it became ugly). I just thought it looked nice to have them together.

src/contrib/graphbuilderadapter.h
TedLyngmo: I did that to get rid of the -Wshadow message which I admit is strange since this isn't really a shadowing situation.

src/directives.cpp
TedLyngmo: Oups, it was for myself, will remove.

src/exp.h
TedLyngmo: Overloading the boolean operators generates -Weffc++ warnings since both operands will be evaluated in those overloads. I also though these operators behaved more like when combining bit patterns than boolean operators.

src/regex_yaml.cpp
TedLyngmo: Sure, will fix.

src/singledocparser.cpp
TedLyngmo: I was to get rid of a -Wshadow message.

src/emitterstate.cpp
TedLyngmo: Sorry, missed that. Will do.

include/yaml-cpp/noncopyable.h
```diff
@@ -13,12 +13,10 @@ namespace YAML {
 // this is basically boost::noncopyable
 class YAML_CPP_API noncopyable {
```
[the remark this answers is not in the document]
TedLyngmo: Would you like deleted/defaulted copy/move ctors/operators for classes holding objects like `std::unique_ptr<...>` where copying is already prohibited while moving is ok?

TedLyngmo:
Hmm, it looks like `clang-format` ruined the tests.

include/yaml-cpp/contrib/anchordict.h
```diff
@@ -22,6 +22,7 @@ namespace YAML {
 template <class T>
 class AnchorDict {
  public:
+  AnchorDict() : m_data{} {}
```
jbeder: Make sure to run clang-format; I think there's an extra space here.

src/setting.h
```diff
@@ -62,9 +62,12 @@ inline std::unique_ptr<SettingChangeBase> Setting<T>::set(const T& value) {
   return pChange;
 }
 
-class SettingChanges : private noncopyable {
+class SettingChanges {
  public:
   SettingChanges() : m_settingChanges{} {}
+  SettingChanges(const SettingChanges&) = delete;
+  SettingChanges(SettingChanges&& rhs) : m_settingChanges(std::move(rhs.m_settingChanges)) {}
```
jbeder: "= default" should work here

include/yaml-cpp/node/impl.h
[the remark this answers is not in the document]
jbeder: OK, fine by me

src/contrib/graphbuilderadapter.h
jbeder: I see. Then just update the local variable to something else that doesn't start with a capital letter.

src/exp.h
jbeder: OK.

src/singledocparser.cpp
jbeder: OK. How about "nextToken"?

jbeder:
> Hmm, it looks like `clang-format` ruined the tests.

Just run it on the files you changed, not on all files. And if it totally changes an entire file, then just run update the section of code that you changed.

include/yaml-cpp/contrib/anchordict.h
TedLyngmo: `clang-format` didn't seem to change that.

PR #690, opened 2019-03-15
unique_ptr deleted incomplete types

Dekken:
Windows MSVC is complaining about "Directives" and "Scanner"


Dekken:
https://ci.appveyor.com/project/Dekken/statick#L217

include/yaml-cpp/scanner.h
```diff
@@ -15,9 +15,9 @@
 #include <stack>
 #include <string>
 
-#include "ptr_vector.h"
-#include "stream.h"
-#include "token.h"
+#include "yaml-cpp/ptr_vector.h"
```
jbeder: Why these changes?

include/yaml-cpp/parser.h
```diff
@@ -11,6 +11,8 @@
 #include <memory>
 
 #include "yaml-cpp/dll.h"
+#include "yaml-cpp/directives.h"
```
jbeder: Is the idea that these are missing includes? What errors or warnings are you getting? I don't see the word "unique_ptr" in the link you sent.

Dekken: wrong link https://ci.appveyor.com/project/Dekken/statick/builds/23102012

include/yaml-cpp/scanner.h
Dekken: consistency

Dekken:
this is quite an annoyance

include/yaml-cpp/scanner.h
jbeder: These files aren't in the set of public headers; they're just in the src directory. That's why they're included as relative names. I'm surprised it still compiles - that's a bug. In any case, please revert these and just leave the additions you need to make msvc happy.

jbeder:
> this is quite an annoyance

By the way, if you're interested in helping further, it'd be cool to have an automated test for Windows.

include/yaml-cpp/scanner.h
Dekken: These files (ptr_vector/etc) have been moved to the include directory as they're required by scanner.h (at least)

so I can replace 

```#include "yaml-cpp/ptr_vector.h"``` with ```#include "yaml-cpp/ptr_vector.h"``` for current directory lookup but I don't see the difference, with the first you know that it's in the include directory

Comment (jbeder) [reply include/yaml-cpp/scanner.h]: Oh, wow, I totally missed that you moved these files to the public directory.

Hm. I'd prefer if we didn't extend the public API here. It looks like the error is just for the move constructor and assignment, right? If so, I don't think the Parser actually needs to be moveable, so you could just change those to be = delete instead of = default.

If we actually do need to move these, then you can just move the definition of the move constructor/assignment to the src file (still writing = default).

**Comment by Dekken** [comment]
I think it's actually the destructor that's the isssue

there's a struct forwarding then unique_ptr - and from my googling it looks like the destructor must be visible for MSVC

**Comment by jbeder** [comment]
The destructor's defined in the [src file](https://github.com/jbeder/yaml-cpp/blob/0d5c57150cebb430fcc216d8066dbb28aeaafe8e/src/parser.cpp#L18).

The error message you linked to earlier says:

```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\memory:2082:21: error: invalid application of 'sizeof' to an incomplete type 'YAML::Scanner'
                static_assert(0 < sizeof (_Ty),
                                  ^~~~~~~~~~~~
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\memory:2296:4: note: in instantiation of member function 'std::default_delete<YAML::Scanner>::operator()' requested here
                        this->get_deleter()(get());
                        ^
p\\include\yaml-cpp/parser.h:33:3: note: in instantiation of member function 'std::unique_ptr<YAML::Scanner, std::default_delete<YAML::Scanner> >::~unique_ptr' requested here
  Parser(Parser&&) = default;
  ^
p\\include\yaml-cpp/parser.h:18:7: note: forward declaration of 'YAML::Scanner'
class Scanner;
      ^
```

Note the 

```
std::unique_ptr<...>::~unique_ptr' requested here
 Parser(Parser&&) = default;
  ^
```

indicating the move constructor. There's no corresponding error for the destructor.

**Comment by Dekken** [comment]
https://ci.appveyor.com/project/Dekken/maiken#L212

**Comment by Dekken** [comment]
OK! I think it's a DLLEXPORT issue

**Comment by Dekken** [comment]
https://ci.appveyor.com/project/Dekken/maiken#L212

**Comment by Dekken** [comment]
this also works 

```
$ git diff
diff --git a/include/yaml-cpp/parser.h b/include/yaml-cpp/parser.h
index e9fba0c..d3a5755 100644
--- a/include/yaml-cpp/parser.h
+++ b/include/yaml-cpp/parser.h
@@ -11,12 +11,12 @@
 #include <memory>

 #include "yaml-cpp/dll.h"
+#include "../../src/directives.h"
+#include "../../src/scanner.h"

 namespace YAML {
 class EventHandler;
 class Node;
-class Scanner;
-struct Directives;
 struct Token;
```

jbeder:
Huh, if it's a dll issue, I have no idea. The only thing that I can see from a C++ standards issue is the move constructor and assignment. It's legal to forward declare a type and then have a unique_ptr of that type, as long as you don't try to call the destructor of the unique_ptr while the type is still incomplete. That's only happening in the move functions.

Please look for a solution without including any private headers.

Dekken:
unique_ptr needs access to the destructor so the call is implicit 

I don't see any way outside the two proposed

jbeder:
unique_ptr only needs access to the destructor during *its* destructor. That only gets called from:

a) the destructor of the Parser class (whose definition is in the .cpp file)
b) the move constructor (whose definition in the .h file)
c) the move assignment operator (whose definition in the .h file)

Do you see another place the unique_ptr destructor gets called?

Dekken:
https://stackoverflow.com/a/40383492/795574 I'm not an expert on what's "legal", I just make things work

Dekken:
https://www.google.com/search?q=use+of+undefined+type+forward+declaration+unique_ptr

jbeder:
That Stack Overflow answer says exactly what I'm saying: move the call to ~unique_ptr to the .cpp file.

PR #694, opened 2019-03-24
moved move/copyop to cpp for MSVC

Dekken:
Replacing https://github.com/jbeder/yaml-cpp/pull/690


Dekken:
my project [here](https://github.com/Dekken/maiken) builds this for Linux/Mac/Windows, perhaps we can trigger it for commits here

jbeder:
I just deleted the move constructor/assignment in bd7f8c60c82614bb0bd1c526db2cbc39dac02fec, so that should obviate this PR.

Dekken:
works thanks!

PR #697, opened 2019-03-26
make use of convert::decode_optional

cishisedis:
so that custom YAML::convert\<T>, where T has no default constructor, can implement 'decode_optional' instead of 'decode'.


jbeder:
I commented on the issue, but I think the upshot is that we’ll wait for std::expected.

PR #698, opened 2019-04-02
Autoconf support

till-s:
Hi - many thanks for yaml-cpp.

Not being a big friend of cmake and the fact that it didn't build both, static- and dynamic libraries out of the box inspired me to add a few lines of gnu-autoconf (which of course also has drawbacks).

If you like you can merge my branch which will enable users to choose between cmake or autoconf...


jbeder:
Sorry, I can't afford to maintain another build system. Thanks for the effort, though!

PR #699, opened 2019-04-05
meson build system support

lukas-weber:
I am using yaml-cpp on systems with horribly outdated software and without privileges. This means I have to install a lot of my dependencies locally.

Meson is a build system with growing support (GNOME, systemd, Xorg, Wayland, Mesa, …) that aims to be easier to use than CMake. But the real reason I use it for my projects is the fact that it can automatically download and build third party dependencies provided they are meson projects. That’s why I went ahead and wrote a meson.build file for yaml-cpp for myself. :) The tests seem to be working and you can even see the fallback dependency feature in action for GTest.

It didn’t require touching the code at all so I thought maybe I make a pull request. But if you are not interested in adding new build systems, that’s fine. ;)

I tried generating the cmake package information from meson as well, but failed on getting it completely right.

Edit:
The commands to build this are
```
meson . build
cd build
ninja
ninja install
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 build/
 /tags
+subprojects/
+!subprojects/*.wrap
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,134 @@
+project('yaml-cpp', 'cpp',
+  version : '0.6.2',
+  meson_version : '>=0.46.0',
+  license : 'MIT')
+
+sources = files([
+  'src/binary.cpp',
+  'src/convert.cpp',
+  'src/directives.cpp',
+  'src/emit.cpp',
+  'src/emitfromevents.cpp',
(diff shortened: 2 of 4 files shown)


jbeder:
Thanks for the effort, but it's too hard to maintain more build systems.

eli-schwartz:
Randomly came across this.

@lukas-weber, Did you ever consider submitting this to https://github.com/mesonbuild/wrapdb/ as a community-maintained build system specifically for subproject use?

lukas-weber:
In principle, I would not mind, but I no longer use yaml-cpp in any of my projects so I wouldn’t have a straightforward way to check if it still works or will work in the future. If it is useful to you and you have the resources to maintain it yourself, feel free to adopt it.

PR #710, opened 2019-06-25
Added ability to emit folded strings

### LINKED ISSUE #405: How do I output a folded version of this?
out << YAML::Literal << "bla bla";

what is the syntax for the folded version of this?

Lemongrass3110:
Fixes #405


Lemongrass3110:
@jbeder any help on fixing your CI tests? I do not really understand what is wrong with it. It somehow duplicates the output on your test it seems.

jbeder:
I’m not sure I get what you’re trying to do here. Folded strings only make sense when the author manually truncates to some number of characters per line (or so). If you just depend on the new lines in the string, it won’t really be different from literals.

Issue #265, opened 2015-03-30
YAML::Node has no comparison operator


GoogleCodeExporter:
```
It isn't possible to use YAML::Node in std::set or std::map due to a lack of 
any comparison operator.

I have a scenario where I need deserialisation of a particular node to always 
return the same instance of the deserialised class. YAML::Node implements the 
equality operator, which means I can easily check if two nodes are actually 
identical, but it has no comparison operator, so I can push them into an 
std::map.

I attempted to write my own comparison operator for the std::map, but I can't 
see any publicly accessible members or accessors to do it.

Furthermore, attempts to use the comparison operator actually work, but it 
looks like the node is implicitly converted to a bool via 
YAML_CPP_OPERATOR_BOOL(). This means the comparison operator always returns 
false.

Is there any sensible way of comparison? I'd settle for internal pointer 
comparison, but it may be more sensible to compare by location in YAML tree or 
something similar.
```

Original issue reported on code.google.com by `sho...@gmail.com` on 13 Nov 2014 at 1:51

dwightguth:
For those of you stumbling upon this later, one nasty hack you can do to get around this is to store a pointer to the value you want as a scalar integer in a node whose key is a complex yaml key. This will allow you to have a "map" whose keys are yaml nodes. Of course, it would still be vastly better if std::hash and operator< were implemented properly instead of the former being unimplementable and the latter being functionally incorrect.

PR #715, opened 2019-07-16
Make nodes hashable

FredrikAppelros:
Implement a specialization of `std::hash<YAML::Node>` so that nodes can be used in unordered associative containers.

This would resolve parts of what is mentioned in #265.

### DIFF
--- a/include/yaml-cpp/node/detail/node.h
+++ b/include/yaml-cpp/node/detail/node.h
@@ -18,6 +18,8 @@ namespace YAML {
 namespace detail {
 class node {
  public:
+  friend struct std::hash<node>;
+
   node() : m_pRef(new node_ref), m_dependencies{} {}
   node(const node&) = delete;
   node& operator=(const node&) = delete;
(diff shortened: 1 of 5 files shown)


jbeder (commented):
What are the semantics of this hash?

FredrikAppelros:
As you can see from the implementation it basically boils down to looking at the location of the node in memory. If two nodes point to the same memory location then they produce the same hash digest.

The idea here is that, as mentioned in #265, there are situations in which you may want to be able to use the nodes themselves as indices into a data structure when processing the graph that the YAML represents.

jbeder:
So there'll still be times when the nodes are the 'same', but don't hash to
the same value?

FredrikAppelros:
I guess that depends on what your definition of "same" is. If I have understood it correctly this is based on the same principle as the existing equality operator that is used for nodes as can be seen [here](https://github.com/jbeder/yaml-cpp/blob/master/include/yaml-cpp/node/detail/node.h#L25).

FredrikAppelros:
Just a friendly ping to see if we could resume this discussion. :)

FredrikAppelros:
Is there anything else that I can do to move this along?

TedLyngmo:
If I put two nodes, created separately, with the same content, would they get the same hash?

The `is()` member function does not test for equality so it's not a great fit.

PR #717, opened 2019-08-03
Make sure output of NaN, Infinity and -Infinity is identical on all platforms

Ezphares:
Since the stream output of these special floating point values are implementation defined, .nan, .inf and -.inf were emitted without dots on some platforms.


Ezphares:
You are right of course, the nan-check was written a bit too quickly. I have fixed up the check and added a unit test for infinities and nan

Ezphares:
Updated with std::isnan and added unit test for doubles as well. I did not do long doubles, as the emmitter currently do not seem to be specialized for them anyway. I can add the tests if needed.

PR #711, opened 2019-07-01
Travis CI: Discover Python syntax errors and undefined names

cclauss:
test/gtest-1.8.0 is not Python 3 compatible https://travis-ci.org/jbeder/yaml-cpp/jobs/552724169#L183

PR #718, opened 2019-08-04
Modernize Python 2 code to prepare for Python 3

cclauss:
As requested in #711 this PR is the result of running __futurize --stage1 -w .__ on the codebase.

https://github.com/PythonCharmers/python-future


jbeder:
Ohhh, I see what you're saying. What are these python scripts for in googlemock? Do they even run when you're running the tests?

Issue #722, opened 2019-08-12
Bazel Support


zaucy:
I would love to see support for [Bazel](https://bazel.build/) in this repository. Would this be something that interests the maintainers here?

I made a fork with the necessary bazel build files over [here](https://github.com/bazelregistry/yaml-cpp), but it would be great to have support from the official repository. I'd be willing to help with improvements or any issues that come up in relation to bazel.

jbeder:
Yep, I'd be happy to put a Bazel file in yaml-cpp! PR welcome :)

PR #724, opened 2019-08-18
Bazel Support (#722)

zaucy:
closes https://github.com/jbeder/yaml-cpp/issues/722

Example of how someone might consume yaml-cpp with bazel

```python
cc_binary(
    name = "example",
    srcs = ["example.cc"],
    deps = ["@com_github_jbeder_yaml_cpp//:yaml-cpp"],
)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 build/
 /tags
+/bazel-*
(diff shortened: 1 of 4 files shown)


Namburger:
this is soooo awesome, thanks a ton

PR #731, opened 2019-09-05
changed cmake version from 3.x to 2.8 for centos 7

kurt-cb:
I did not see a good reason to require make 3.x
builds fine on lentos 7 with this change.


jbeder:
I think the goal was to keep this up to date. 3.1 was from 2016.

PR #734, opened 2019-09-10
CMake: Prefix options with "YAML" and hide platform-specific options

asmaloney:
When including yaml-cpp as a subproject, some option names can conflict with other projects.

(1) Make sure the yaml-cpp options are prefixed with YAML
(2) Hide platform-specific options when possible to avoid cluttering the cmake option list

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,16 +24,20 @@ option(YAML_CPP_INSTALL "Enable generation of install target" ON)
 # --> General
 # see http://www.cmake.org/cmake/help/cmake2.6docs.html#variable:BUILD_SHARED_LIBS
 #     http://www.cmake.org/cmake/help/cmake2.6docs.html#command:add_library
-option(BUILD_SHARED_LIBS "Build Shared Libraries" OFF)
+option(YAML_BUILD_SHARED_LIBS "Build Shared Libraries" OFF)
 
 # --> Apple
-option(APPLE_UNIVERSAL_BIN "Apple: Build universal binary" OFF)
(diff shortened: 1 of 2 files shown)


jbeder:
Can you update the documentation for BUILD_SHARED_LIBS at the same time?

asmaloney:
Ah yes - sorry about that. Done.

traversaro:
Are you sure that now the library builds correctly, according to the new  `YAML_BUILD_SHARED_LIBS` variable? As `add_library` is called without type specifier, it still listen to the `BUILD_SHARED_LIBS` CMake variable (see https://cmake.org/cmake/help/latest/command/add_library.html and https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS ), that is now an option still observed by the project, but not documented.

asmaloney:
Thanks @traversaro !

PR #737 should fix it.

jbeder:
This makes me wonder if this is the right decision. The documentation:

https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

Suggests that projects should just add BUILD_SHARED_LIBS as an option.

asmaloney:
The problem is that this setting is project-wide, not specific to the target. So if you want some shared libs and some linked statically it has to be target-specific. If anything it should be used at the application level, not the library level like this.

I think the docs in some ways reflect the "old" cmake, not the newer target-based approach.

jbeder:
OK, makes sense.

PR #741, opened 2019-09-21
Refactor CMake to use more modern paradigms

slurps-mad-rips:
Remove 2.6-isms
Remove 2.8-isms
Bump CMake minimum version to 3.4

Disable some options when used as a subdirectory

use `CONFIGURE_DEPENDS` with `file(GLOB)` when possible

Backport CMake 3.15's MSVC_RUNTIME_LIBRARY setting. This simplifies a lot of the logic regarding /M(T|D)[d]
Set all compile options as generator expressions.
Set all find-package files to be installed to the correct file.

Remove `export(PACKAGE)`, as this has been deprecated.
Remove fat binary support
Remove manual setting of iPhone settings. These should be set by parent
projects.
Remove use of ExternalProject for a local use
Conditionally remove format target unless clang-format is found


CMakeLists.txt
```diff
@@ -1,367 +1,165 @@
...
+set(not-msvc $<NOT:$<CXX_COMPILER_ID:MSVC>>)
+
+if (NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
+  set(CMAKE_MSVC_RUNTIME_LIBRARY
+    MultiThreaded$<$<CONFIG:Debug>:Debug>$<${build-shared}:DLL>)
+endif()
+
+set(contrib-pattern "src/contrib/*.cpp")
+set(src-pattern "src/*.cpp")
+if (CMAKE_VERSION VERSION_GREATER 3.12)
```
jbeder: What is this conditional for?

CMakeLists.txt
```diff
@@ -1,367 +1,165 @@
...
+# http://msdn.microsoft.com/en-us/library/3c594ae3.aspx
+set(msvc-rt $<TARGET_PROPERTY:MSVC_RUNTIME_LIBRARY>)
+
+set(mtd-static $<STREQUAL:${msvc-rt},MultiThreadedDebug>)
+set(mt-static $<STREQUAL:${msvc-rt},MultiThreaded>)
+
+set(mtd-dll $<STREQUAL:${msvc-rt},MultiThreadedDebugDLL>)
+set(mt-dll $<STREQUAL:${msvc-rt},MultiThreadedDLL>)
+
+set(backport $<VERSION_LESS:${CMAKE_VERSION},3.15>)
```
jbeder: Same here.

CMakeLists.txt
```diff
@@ -1,367 +1,165 @@
...
+
+add_library(yaml-cpp ${yaml-cpp-type})
+add_library(yaml::yaml ALIAS yaml-cpp)
```
jbeder: What does this mean?

slurps-mad-rips: When `yaml-cpp` is used via `add_subdirectory`, users can link against `yaml::yaml` and anyone who might pull in their project will not be able to modify the properties of it. ALIAS libraries are explained in detail [here][1]

[1]: https://cmake.org/cmake/help/latest/command/add_library.html#alias-libraries

CMakeLists.txt
slurps-mad-rips: The current cmake files have you using `file(GLOB)`, which means to add or remove files you would need to regenerate the project. What this does is that for anyone using a more recent version of CMake, we insert `CONFIGURE_DEPENDS` into the arguments (tbqh, the `VERSION_GREATER` should be against 3.11. Slight typo which I can fix) which will check if file lists have changed. This means anyone forking or adding to the project as they might will have an automatic step to regenerate the project. This can help with a variety of build tools that *don't* start with "X" and end in "Code", but it also allows tools like Ninja and Make to give developers a better experience when actively working on a project.

CMakeLists.txt
jbeder: So is the idea that the yaml-cpp target is only internally visible, and yaml::yaml is externally visible, and immutable?

If so, is yaml-cpp really only internally visible, and if so, what makes it so?

CMakeLists.txt
slurps-mad-rips: This is being done because CMake 3.15 adds support for setting `CMAKE_MSVC_RUNTIME_LIBRARY`. We don't want to step on their toes for that, thus we only enable this setting if someone is using pre-3.15. Otherwise we could have `target_compile_options` and cmake generation conflict and cause link time errors. All the logic regarding the `backport` variable is just done to make sure that it gets eliminated early on newer CMake versions.

CMakeLists.txt
jbeder: Makes sense; this one sounds good.

CMakeLists.txt
jbeder: Gotcha. Can you rename this variable, something like add_msvc_runtime_option (or whatever, something that says what it does)?

CMakeLists.txt
slurps-mad-rips: `yaml-cpp` isn't internally visible if used in an `add_subdirectory` form, but *can* be partially internally visible if its an `IMPORTED` library, and not marked as `GLOBAL` (which is the default in `find_package`)

IIRC I fixed (or have a todo to fix) the `install(TARGETS ... EXPORT ...)`  `find_package` file to make sure `yaml::yaml` is available. This means that the actual exported target name doesn't really matter, as an ALIAS is used instead, and in the case of `find_package`, one wouldn't be able to modify it. Doing it this way allows people to swap between an installed version, or one they pull in manually, without having upstream projects having to check or care.

The scoping of libraries in CMake is weird, and we unfortunately can't change Kitware's minds on how to fix them because that would require breaking backwards compatibility. 

It's also hella confusing, so I'm trying to reduce that while giving upstream users the option.

CMakeLists.txt
slurps-mad-rips: Sure! I'll update the names in a bit and do a push when I have a moment later today.

slurps-mad-rips: Names have been updated! Let me know if it meets your needs :)

jbeder:
It looks you have a conflict also, before I merge this - I released a version yesterday, so you'll have to update the version tag.

slurps-mad-rips:
> It looks you have a conflict also, before I merge this - I released a version yesterday, so you'll have to update the version tag.

Working on it now. Should we bump the version as well? Otherwise folks will get the outdated CMake build if they pull the 0.6.3 version.

jbeder:
You mean, bump it again to 0.6.4? I want to let this bake for a bit in case there are any bugs.

slurps-mad-rips:
Yeah, that's what I meant. And that's totally fine as well. :)

jbeder:
Can you take a look at

https://github.com/jbeder/yaml-cpp/issues/745

jbeder:
One more :)

Can you look at #748?

jbeder:
Could you review #760? It seems reasonable, but I want to make sure it doesn't revert a goal that this PR had.

slurps-mad-rips:
Ack! This is a typo and the fix for it is way simpler than what that PR does (and it's literally "I forgot to put a space in). I'm sorry for all the problems 😬

jbeder:
Could you also take a look at #773 and #776? It looks like people are having trouble with Windows builds.

jirutka:
This PR caused tons of problems. :(

jbeder:
@jirutka Sorry for any trouble - could you help fix whatever problems you're running into?

PR #754, opened 2019-10-02
Modernize: Use "using" instead of "typedef"

asmaloney:


include/yaml-cpp/node/detail/bool_type.h
```diff
@@ -13,7 +13,7 @@ struct unspecified_bool {
   struct NOT_ALLOWED;
   static void true_value(NOT_ALLOWED*) {}
 };
-typedef void (*unspecified_bool_type)(unspecified_bool::NOT_ALLOWED*);
```
jbeder: If you’re interested, in a follow up, this pattern can be replaced with an explicit bool operator

asmaloney: I don't fully understand what the `YAML_CPP_OPERATOR_BOOL` macro is actually doing & I've never seen this pattern before.

Is it replaceable by this?

```cpp
explicit operator bool() const { return IsDefined(); }
```

jbeder: Yep, that's right. It's an old pattern from before explicit operators showed up.

Issue #759, opened 2019-10-03
Build fails on CMAKE version 3.10.2 


azubieta:
When trying to build the latest commit from master I got the following cmake error:
```
CMake Error at test/CMakeLists.txt:23 (add_executable):
  add_executable called with incorrect number of arguments
```

Which points to this line: https://github.com/jbeder/yaml-cpp/blame/master/test/CMakeLists.txt#L23

swartik:
Glad I'm not the only one to encounter this. According to the cmake documentation, add_executable requires at least one source in versions 3.0–3.10. Starting with version 3.11.4, sources are optional.

fyazici:
should be solved by #763 
 can you ( @azubieta and @swartik ) please confirm?

swartik:
It works now with cmake version 3.10. Thanks!

azubieta:
Thanks !

Issue #757, opened 2019-10-03
CMake build configuration minimum version


fyazici:
It appears that the main CMakeLists.txt file utilizes `add_library` with omitted sources. This feature is not supported in cmake 3.10.? and below. Noticed the issue while configuring on a machine with Ubuntu 16.04 cmake 3.10.2. I suggest that the minimum version statement be updated to at least 3.11 so that cmake fails in a more understable way.

jbeder:
3.11 is too recent, I think (Feb 2018). Can you instead update it to work
on earlier CMakes? (I don't understand what "omitted sources" means.)

fyazici:
I did not dig into the file any deeper than finding the culprit but I will look and see what I can do (though I am not very experienced in CMake) in about a day from now.

ashpekvs:
3.10 cmake fails to build the lib on Ubuntu 18.04.3 x64 LTS
issues with pthreads lib (can't find it)

3.15.4 worked like a charm!
built 100%

fyazici:
solved by #763

PR #760, opened 2019-10-03
add_library & and_executable omitted sources added back

fyazici:
for cmake 3.10.x and below compatibility. related to issue #757 (by me) and #759 (by another user)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,9 @@ set(msvc-rt-mt-dll $<STREQUAL:${msvc-rt},MultiThreadedDLL>)
 
 set(backport-msvc-runtime $<VERSION_LESS:${CMAKE_VERSION},3.15>)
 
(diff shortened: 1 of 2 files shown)


slurps-mad-rips:
Hi apologies about this. I forgot to add a space between the target name and the closing parentheses. This is technically how its supposed to be done. In other words:

```cmake
add_library(yaml-cpp ${yaml-cpp-type} "")
```

would also work. That said, before this gets merged @jbeder, either the put an empty string approach should be taken, or the `target_sources` calls need to be removed. I'd prefer to keep the `target_sources` calls in, but that's a personal preference for "readability" (even if `add_library(name "")` isn't as readable.

fyazici:
@slurps-mad-rips no worries. I will try your suggestion on my 16.04 machine just in case. I am closing this PR for now. If that works as expected I will open a new PR with your solution.

PR #762, opened 2019-10-04
Modernize: Use range-based for loops for readability

asmaloney:


include/yaml-cpp/node/detail/node.h
```diff
@@ -42,9 +42,8 @@ class node {
       return;
 
     m_pRef->mark_defined();
-    for (nodes::iterator it = m_dependencies.begin();
-         it != m_dependencies.end(); ++it)
-      (*it)->mark_defined();
+    for (auto dependency : m_dependencies)
```
jbeder: Use the actual type here rather than auto. (And throughout where you've used auto.)

In general, I only like auto when

(a) it avoids stutter, like "auto *foo = new Foo()", or
(b) it saves typing a ridiculous type, like some_crazy::iterator

See

https://google.github.io/styleguide/cppguide.html#Type_deduction

for more details.

asmaloney: I replaced all but one - an anonymous struct in convert.cpp @line 64.

PR #769, opened 2019-10-11
Add Android NDK complie supported.

wuruofan:
Usage: ./ndk_make.sh [clean]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 build/
 /tags
 /bazel-*
+
+libs/
+obj/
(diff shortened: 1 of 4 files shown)


jbeder:
Can this be generated from CMake with some flags?

wuruofan:
I think it can and it is the next thing I want to do.

But actually I have less experience in using CMake compiling Android native library without Android Studio's help.😂
No matter which method is used, the `ndk-bundle` is needed.

jbeder:
OK, I think I'll close this then. I don't want to maintain lots of different build systems.

PR #772, opened 2019-10-14
CMake: Add Android dynamic library cross compile support.

wuruofan:
Add cross compile for Android instruments to README.md.


jbeder:
It surprises me that this would be the way to handle Android builds; we don't, e.g., have a special option for iOS builds, but it works fine.

Do you have CMake documentation showing that this is the right way to do it?

wuruofan:
OK.

https://cmake.org/cmake/help/v3.7/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-a-standalone-toolchain

The link above is from cmake.org, but Android official doesnot recommend this way.

https://developer.android.com/ndk/guides/cmake.html#variables

You can see the `COMMAND LINE` tab.

I have tried to write my own toolchain.cmake to contain all necessary arguments as default, but it doesn't make much sense. Android Studio use gradle to compile, actually the arguments is configured just like what I do, so that's why I decide just write some instruments in README.

c4pQ:
@slurps-mad-rips I think, you could help here.

slurps-mad-rips:
To be honest, I'm a little confused why you're copying the libc++ shared library. Also I'd prefer to refer to CMake's position on Android support. The NDK's android toolchain file is... not great if I'm being honest, but I'm not going to help google fix it when they have engineers paid way more than I am :v

wuruofan:
> To be honest, I'm a little confused why you're copying the libc++ shared library. Also I'd prefer to refer to CMake's position on Android support. The NDK's android toolchain file is... not great if I'm being honest, but I'm not going to help google fix it when they have engineers paid way more than I am :v

I do copy the libc++ shared library just because I wanna use yaml-cpp as a dynamic library, and since NDK r18, libc++ is the only STL on Android, which is NOT a system library. You can see more information from this [page](https://developer.android.com/ndk/guides/cpp-support) from Android official.

slurps-mad-rips:
> I do copy the libc++ shared library just because I wanna use yaml-cpp as a dynamic library, and since NDK r18, libc++ is the only STL on Android, which is NOT a system library. You can see more information from this [page](https://developer.android.com/ndk/guides/cpp-support) from Android official.

If that's the case, the copying should be done in your root CMakeLists.txt file, not within `yaml-cpp`, as this could cause a clash with anyone who wants their `libc++_shared.so` in a different location. Your patch also doesn't take into account if someone statically links `libc++` to a dynamically linked `yaml-cpp`. This is also something that should be taken care of in your toolchain file.

wuruofan:
> > I do copy the libc++ shared library just because I wanna use yaml-cpp as a dynamic library, and since NDK r18, libc++ is the only STL on Android, which is NOT a system library. You can see more information from this [page](https://developer.android.com/ndk/guides/cpp-support) from Android official.
> 
> If that's the case, the copying should be done in your root CMakeLists.txt file, not within `yaml-cpp`, as this could cause a clash with anyone who wants their `libc++_shared.so` in a different location. Your patch also doesn't take into account if someone statically links `libc++` to a dynamically linked `yaml-cpp`. This is also something that should be taken care of in your toolchain file.

I wrote clearly in README, the work is only `Cross Compile Dynamic Library for Android`, and I don't think someone else prefer to statically link `libc++` to their library on Android developing work.

Anyway, you can do copy or not, it doesn't matter.

I do just because I think it will be more convenience for Android APP developers, when they do compile, they will get all libraries they need (without libc++_shared.so, libyaml-cpp.so won't work). They can just simply copy libraries to their Android Studio project without modify their existed CMakeList.

I have tried to combine NDK's `android.toolchain.cmake` work with `yaml-cpp`, but it is not simple and  elegant at all, so I just add some lines in README.

If you get better ways to make `yaml-cpp` work on Android, please let me know.

slurps-mad-rips:
> I wrote clearly in README, the work is only Cross Compile Dynamic Library for Android, and I don't think someone else prefer to statically link libc++ to their library on Android developing work.

You cannot assume this isn't an already existing workflow

> Anyway, you can do copy or not, it doesn't matter.

It does, in fact, matter.

> I do just because I think it will be more convenience for Android APP developers, when they do compile, they will get all libraries they need (without libc++_shared.so, libyaml-cpp.so won't work). They can just simply copy libraries to their Android Studio project without modify their existed CMakeList.

It might seem convenient for you, but this does not mean it will be convenient for everyone. If someone wants to build their own `libc++_shared.so` for android, your copying could potentially overwrite their custom version and no one would know unless they inspected `yaml-cpp`'s `CMakeLists.txt`

> If you get better ways to make yaml-cpp work on Android, please let me know, and I will close this PR.

I suggest you read CMake's own [documentation][1] on properly interacting with the NDK. Setting the `CMAKE_ANDROID_STL_TYPE` to `c++_shared` should do the correct thing. I would personally *not* rely on the NDK's toolchain file, and instead simply set the various parameters on the commandline as specified in the linked documentation.

[1]: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-the-ndk

wuruofan:
I had read the CMake document about ndk, and I had tried to do set command line parameters the documents said to do so, but it does not work as expected, something must be in toolchain file in my situation, I wrote one, but soon realized that as I must use ndk-bundle to compile why not use official toolchain, it better than mine, maybe you can do better than it.
I'll close this PR and wish to see your work.

PR #777, opened 2019-10-15
Show test output on CI

carmeli-tamir:
When checking my previous PR I wasn't sure if the tests indeed execute, so I ran them locally. The tests do run but don't show the output. 
This PR makes the CI show the tests output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_script:
   - cd ..
 script:
   - cmake --build build
-  - cmake --build build --target test
+  - cmake --build build --target check
 matrix:
   exclude:
   - os: linux
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -46,4 +46,4 @@ if (NOT DEFINED CMAKE_CXX_STANDARD)
 endif()
 
 
-add_test(yaml-cpp::test yaml-cpp-tests)
+add_custom_target(check COMMAND yaml-cpp-tests)


jbeder:
Is there an alternate way of making CMake print the output? It seems like making a custom command isn't what's intended.

carmeli-tamir:
There might be a way to show it only on failure, but I think it's beneficial to see it on success too.
I based the PR on this stackoverflow thread https://stackoverflow.com/questions/5709914/using-cmake-how-do-i-get-verbose-output-from-ctest
Where some of the answers suggest this. I understand that this feels "unnatural" but I didn't find a better way.

carmeli-tamir:
Also checked in the add_test documentation https://cmake.org/cmake/help/v3.2/command/add_test.html , doesn't seem that it supports this feature.

jbeder:
I like the "output on failure". If someone wants to see the output on
success, they can just run it locally, I think; if the tests pass, there
should be no need to investigate further.

carmeli-tamir:
OK, changed to show only on failure.
This requires a new custom target (but leaves the add_test directive).
References:
https://stackoverflow.com/questions/36726545/replace-ctest-command-with-ctest-output-on-failure-permanently-for-a-specifi

https://stackoverflow.com/questions/16158091/cmake-setting-an-environmental-variable-for-ctest-or-otherwise-getting-failed

jbeder:
Can you just set 'CTEST_OUTPUT_ON_FAILURE=1' in the travis command?

jbeder:
(In other words: that SO answer was from 2013. We're using a much later CMake now. Maybe they made this easier?)

PR #782, opened 2019-10-31
Adapt label postfix to fix build with Visual Studio

dAmihl:
The inline IF:ELSE syntax doesn't seem to work with Visual Studio.
I don't know CMake enough to fix it, so here's a quick fix to keep the postfix but make it build properly.


MrScriptMaster:

PR #788, opened 2019-11-17
Removed dynamic allocation

arttet:
Hello.

I have removed dynamic allocation from `TokenNames`.
Now this code works faster.

### DIFF
--- a/src/token.h
+++ b/src/token.h
@@ -13,7 +13,7 @@
 #include <vector>
 
 namespace YAML {
-const std::string TokenNames[] = {
+constexpr char const* TokenNames[] = {
     "DIRECTIVE",        "DOC_START",      "DOC_END",       "BLOCK_SEQ_START",
     "BLOCK_MAP_START",  "BLOCK_SEQ_END",  "BLOCK_MAP_END", "BLOCK_ENTRY",
     "FLOW_SEQ_START",   "FLOW_MAP_START", "FLOW_SEQ_END",  "FLOW_MAP_END",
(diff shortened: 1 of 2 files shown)


jbeder:
It looks like the allocation only happens at startup, and it looks small. Do you expect any runtime difference? Have you measured any difference?

arttet:
I had counted a number of times a function new calls. It is the large number (about 100 in VS 2017).
Now it doesn't do anything at run-time (no overhead at run-time).

hyperxor:
@jbeder Using constexpr looks good in this case, we know all values of `TokenNames` on compile time.

@arttet Artyom, what's the tool do you use to measure the number of `new` calls? Something inside VS?

arttet:
This is my example.

```
#include <new>
#include <cstdio>

int number_of_allocs = 0;

void* operator new(std::size_t size) {
    ++number_of_allocs;
    void *p = malloc(size);
    if (!p) throw std::bad_alloc();
    return p;
}
void operator delete(void* ptr) throw() { free(ptr); }

#include <yaml-cpp/yaml.h>

int main(int const argc, char const * const argv[])
{
    printf("%i", number_of_allocs); // <--- 126 in VS 2017

    YAML::Node config = YAML::LoadFile(argv[1]);
    return EXIT_SUCCESS;
}
```

[I read it on stackoverflow.](https://stackoverflow.com/questions/9927856/how-to-use-operator-new-to-count-number-of-times-of-dynamic-memory-allocation)

hyperxor:
@arttet I have checked it on Ubuntu with gcc version 7.4.0:
Current code base:
```
6
```
After your change:
```
0
```

jbeder:
Here's the thing - replacing `std::string` with `const char *` makes the code less clear. For a handful of small `new` calls at startup (they're tiny strings!), I'm not sure it matters.

(Note also you've got an unrelated CMake change in this PR.)

Because of this, I'm going to close this PR.

PR #791, opened 2019-11-23
Small code improvements in Parser + test

hyperxor:
I have made small improvements in `Parser` class implementation and one test

### DIFF
--- a/test/parser_test.cpp
+++ b/test/parser_test.cpp
@@ -1,0 +1,20 @@
+#include "yaml-cpp/parser.h"
+#include "mock_event_handler.h"
+#include "gtest/gtest.h"
+
+using YAML::Parser;
+using YAML::MockEventHandler;
+
+TEST(ParserTest, Empty) {
+    Parser parser;
+
+    EXPECT_FALSE(parser);
+
+    MockEventHandler handler;
+    EXPECT_FALSE(parser.HandleNextDocument(handler));
+    
+    std::ostringstream oss;
+    parser.PrintTokens(oss);
+
+    EXPECT_EQ(oss.str(), "");
+}
--- a/test/parser_test.cpp
+++ b/test/parser_test.cpp
@@ -12,10 +12,5 @@ TEST(ParserTest, Empty) {
     EXPECT_FALSE(parser);
 
     StrictMock<MockEventHandler> handler;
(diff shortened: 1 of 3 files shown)


test/parser_test.cpp
```diff
@@ -0,0 +1,20 @@
...
+TEST(ParserTest, Empty) {
+    Parser parser;
+
+    EXPECT_FALSE(parser);
+
+    MockEventHandler handler;
+    EXPECT_FALSE(parser.HandleNextDocument(handler));
+    
+    std::ostringstream oss;
+    parser.PrintTokens(oss);
```
jbeder: Maybe instead of checking the token list, verify that there are no interactions with the MockEventHandler?

hyperxor: @jbeder Yes, good proposal! I will do it. I have some doubts about doing it _instead_. I would add it as an addition to current checks. What do you think?

jbeder: The reason I'd suggest *instead* is that Parser::PrintTokens is a debugging function (it's actually unused, and probably should just be deleted...), so it's not good to rely on its output in a test.

hyperxor: @jbeder I have done it

test/parser_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+using YAML::MockEventHandler;
+using ::testing::StrictMock;
+
+TEST(ParserTest, Empty) {
+    Parser parser;
+
+    EXPECT_FALSE(parser);
+
+    StrictMock<MockEventHandler> handler;
+    parser.HandleNextDocument(handler);
```
jbeder: Why not keep the expect false?

test/parser_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+    
+    std::ostringstream oss;
```
jbeder: Remove this, as discussed?

hyperxor: @jbeder sorry, did wrong change, now everything looks as discussed

test/parser_test.cpp
hyperxor: @jbeder I did wrong commit, sorry, now everything is good

Issue #519, opened 2017-07-28
The assert can lead to dos.


zhouat:
The assert [code](https://github.com/jbeder/yaml-cpp/blob/master/src/scanner.cpp#L32) can lead to a dos attack.

Trigger:
```
cat << EOF > file_poc
> !2
> EOF

build/util/parse file_poc
```
crash:
```
parse: /yaml/yaml-cpp/src/scanner.cpp:32: YAML::Token &YAML::Scanner::peek(): Assertion `!m_tokens.empty()' failed.
Aborted

```

carnil:
This issue has been assigned CVE-2017-11692

Justinzobel:
@jbeder can we please see an update on this one also?

hobbes1069:
Ping

jbeder:
Is there a PR you'd like me to review?

carnil:
The issue is still present in abf941b20d21342cd207df0f8ffe09f41a4d3042 FWIW (seen with a Debug build)

jbeder:
Fixed by c9460110e072df84b7dee3eb651f2ec5df75fb18

PR #806, opened 2020-01-20
Handle bad input with a ParseException not an assert()

AlanGriffiths:
Handle bad input with a ParseException not an assert() (Fixes: #519 aka CVE-2017-11692)

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-11692

### DIFF
--- a/include/yaml-cpp/exceptions.h
+++ b/include/yaml-cpp/exceptions.h
@@ -52,6 +52,7 @@ const char* const INVALID_HEX = "bad character found while scanning hex number";
 const char* const INVALID_UNICODE = "invalid unicode: ";
 const char* const INVALID_ESCAPE = "unknown escape character: ";
 const char* const UNKNOWN_TOKEN = "unknown token";
+const char* const NO_MORE_TOKENS = "no more tokens to be read";
 const char* const DOC_IN_SCALAR = "illegal document indicator in scalar";
 const char* const EOF_IN_SCALAR = "illegal EOF in scalar";
 const char* const CHAR_IN_SCALAR = "illegal character in scalar";
--- a/src/scanner.cpp
+++ b/src/scanner.cpp
@@ -34,9 +34,10 @@ void Scanner::pop() {
 
 Token& Scanner::peek() {
   EnsureTokensInQueue();
-  assert(!m_tokens.empty());  // should we be asserting here? I mean, we really
-                              // just be checking
-                              // if it's empty before peeking.
+
+  // If there are no more tokens, we cannon fulfil our post-condition
+  if (m_tokens.empty())
+      throw ParserException(INPUT.mark(), ErrorMsg::NO_MORE_TOKENS);
 
 #if 0
 		static Token *pLast = 0;
--- a/test/parser_test.cpp
+++ b/test/parser_test.cpp
@@ -1,9 +1,11 @@
 #include "yaml-cpp/parser.h"
+#include "yaml-cpp/exceptions.h"
 #include "mock_event_handler.h"
 #include "gtest/gtest.h"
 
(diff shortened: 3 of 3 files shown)


jbeder:
I believe this should actually be valid YAML, so it's preferable to make it parse rather than throw an exception. (Although I suppose an exception is better than an assert?)

jbeder:
Both:

https://codebeautify.org/yaml-validator
http://www.yamllint.com/

validate your example (for reference, below)

```
!2
```

jbeder:
Hmm, but

https://yaml-online-parser.appspot.com/

does not like it, and gives the error:

```
ERROR:

could not determine a constructor for the tag '!2'
  in "<unicode string>", line 1, column 1:
    !2
    ^
```

jbeder:
OK, so I'm going to think about whether this is valid YAML or not; but the desired behavior is:

(a) if it's valid, let's parse it
(b) if it's invalid, let's give a proper error message (not having to do with "tokens")

jbeder:
OK, I believe it's valid to parse, but not to represent. So we shouldn't throw an exception at all, since yaml-cpp is permissive about representation. Here are the relevant lines from the spec:

```
[207] l-bare-document ::= s-l+block-node(-1,block-in)
[196] s-l+block-node(n,c) ::= s-l+block-in-block(n,c)
                            | s-l+flow-in-block(n)
[197] s-l+flow-in-block(n) ::= s-separate(n+1,flow-out)
                               ns-flow-node(n+1,flow-out) s-l-comments
[161] ns-flow-node(n,c) ::= c-ns-alias-node
                          | ns-flow-content(n,c)
                          | ( c-ns-properties(n,c)
                              ( ( s-separate(n,c) ns-flow-content(n,c) )      
                                | e-scalar ) )
[96] c-ns-properties(n,c) ::= ( c-ns-tag-property    
                                ( s-separate(n,c) c-ns-anchor-property )? )
                            | ( c-ns-anchor-property
                                ( s-separate(n,c) c-ns-tag-property )? )

[97] c-ns-tag-property ::= c-verbatim-tag | c-ns-shorthand-tag | c-non-specific-tag
[99] c-ns-shorthand-tag ::= c-tag-handle ns-tag-char+
[89] c-tag-handle ::= c-named-tag-handle | c-secondary-tag-handle | c-primary-tag-handle
[90] c-primary-tag-handle ::= “!”
[105] e-scalar ::= /* Empty */
[80] s-separate(n,c) ::= c = block-out ⇒ s-separate-lines(n)
                         c = block-in  ⇒ s-separate-lines(n)
                         c = flow-out  ⇒ s-separate-lines(n)
                         c = flow-in   ⇒ s-separate-lines(n)
                         c = block-key ⇒ s-separate-in-line
                         c = flow-key  ⇒ s-separate-in-line
[81] s-separate-lines(n) ::= ( s-l-comments s-flow-line-prefix(n) )
                           | s-separate-in-line
[66] s-separate-in-line ::= s-white+ \| /* Start of line */
```

So, I think the document:

```
!2
```

tokenizes as:

```
s-separate-in-line c-primary-tag-handle "2" e-scalar
```

which parses to

```
primary-tag("2") e-scalar
```

In other words, an empty scalar (i.e., null) with a primary tag of 2. If that tag isn't registered, then this'll fail in representation, but we should still parse like that.

jbeder:
Fixed by c9460110e072df84b7dee3eb651f2ec5df75fb18

AlanGriffiths:
> Fixed by [c946011](https://github.com/jbeder/yaml-cpp/commit/c9460110e072df84b7dee3eb651f2ec5df75fb18)

Works for me. Thanks!

Issue #809, opened 2020-01-21
Make move constructor and assignment operator noexcept


TedLyngmo:
Classes containing a `YAML::Node` will not get their move constructors and assignment operators created with the `noexcept(true)` specifier - which in turn makes the classes containing a `YAML::Node` failing to pass `clang-tidy` linting or compiling with `clang++`. Example to reproduce:
```
#include <yaml-cpp/yaml.h>
  
class Foo {
private:
    Foo(const Foo&) = default;
    Foo(Foo&&) noexcept = default; // won't help
    Foo& operator=(const Foo&) = default;
    Foo& operator=(Foo&&) noexcept = default; // won't help
private:
    YAML::Node node;
};

int main() {}
```
```
% clang++ -I ~/git/yaml-cpp/include -L ~/git/yaml-cpp/build -lyaml-cpp -o yaml3-clang yaml3.cpp
yaml3.cpp:6:5: error: exception specification of explicitly defaulted move constructor does not match the calculated one
    Foo(Foo&&) noexcept = default; // won't help
    ^

yaml3.cpp:8:10: error: exception specification of explicitly defaulted move assignment operator does not match the calculated one
    Foo& operator=(Foo&&) noexcept = default; // won't help
         ^
2 errors generated.
```

Issue #721, opened 2019-08-11
YAML::node cannot deal with std::swap


darkangel-ua:
YAML::Node n1;
YAML::Node n2;

n1["id"] = "1";
n2["id"] = "2";

std::swap(n1, n2);

Now n1 == n2 !? Is it by design?

PR #810, opened 2020-01-22
Make Node move constructor and assignment operator noexcept (#809)

TedLyngmo:
Move constructor:
```
 * m_isValid    (bool)                 exchange(rhs.m_isValid, true)
 * m_invalidKey (std::string)          std::move()
 * m_pMemory    (shared_memory_holder) std::move()
 * m_pNode      (node*)                exchange(rhs.m_pNode, nullptr)
```
 This leaves the moved-from Node as if it was just default constructed.

Move assignment:

* A sanity test is performed to check if it's a valid move, and if not: `*this` is returned (with an added assert() for debugging).

* A temporary Node is move constructed (using the new move constructor), leaving the moved-from Node as if it was just default constructed.

* If this->m_pNode == nullptr, the same operation as AssignNode would do is done and `*this` is returned.

* if temporary.m_pNode == nullptr:
  m_pNode->set_null()
  swap(*this, temporary)
  return *this;

 Otherwise the merge step that `AssignNode` would do if both `m_pNodes` are not
 nullptr is done, using a new member function, `AssignNodeDetail()`.

### DIFF
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -12,11 +12,29 @@
 #include "yaml-cpp/node/detail/node.h"
 #include "yaml-cpp/node/iterator.h"
 #include "yaml-cpp/node/node.h"
+#include "yaml-cpp/noexcept.h"
+#include <cassert>
 #include <sstream>
 #include <string>
+#include <utility>
 
 namespace YAML {
-inline Node::Node()
+namespace detail {
+#if __cplusplus >= 201402L
+using ::std::exchange;
+#else
+template<class T, class U = T>
+T exchange(T& obj, U&& new_value) {
+  T old_value = std::move(obj);
+  obj = std::forward<U>(new_value);
+  return old_value;
+}
+#endif
+}  // namespace detail
+}  // namespace YAML
+
+namespace YAML {
+inline Node::Node() YAML_CPP_NOEXCEPT
     : m_isValid(true), m_invalidKey{}, m_pMemory(nullptr), m_pNode(nullptr) {}
 
 inline Node::Node(NodeType::value type)
@@ -43,6 +61,13 @@ inline Node::Node(const detail::iterator_value& rhs)
       m_pNode(rhs.m_pNode) {}
 
 inline Node::Node(const Node&) = default;
+
+inline Node::Node(Node&& rhs) YAML_CPP_NOEXCEPT
+    : m_isValid(detail::exchange(rhs.m_isValid, true)),
+      m_invalidKey(std::move(rhs.m_invalidKey)),
+      m_pMemory(std::move(rhs.m_pMemory)),
+      m_pNode(detail::exchange(rhs.m_pNode, nullptr)) {
+}
 
 inline Node::Node(Zombie)
     : m_isValid(false), m_invalidKey{}, m_pMemory{}, m_pNode(nullptr) {}
@@ -208,6 +233,13 @@ inline void Node::SetStyle(EmitterStyle::value style) {
 }
 
 // assignment
+inline bool Node::CheckValidMove(const Node& rhs) const YAML_CPP_NOEXCEPT {
+  // Note: Both m_pNode's can be nullptr.
+  return
+    m_isValid && rhs.m_isValid &&
+    (!m_pNode || !rhs.m_pNode || !m_pNode->is(*rhs.m_pNode));
+}
+
 inline bool Node::is(const Node& rhs) const {
   if (!m_isValid || !rhs.m_isValid)
     throw InvalidNode(m_invalidKey);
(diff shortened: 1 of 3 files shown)


include/yaml-cpp/node/impl.h
```diff
@@ -209,6 +240,19 @@ inline Node& Node::operator=(const Node& rhs) {
...
 }
 
+inline Node& Node::operator=(Node&& rhs) YAML_CPP_NOEXCEPT {
+  if (is(rhs))
+    return *this;
+  Node tmp(std::move(rhs));
+  try {
+    AssignNode(tmp);
+  }
+  catch(...) {
```
jbeder: I think this is a little unclear. The problem seems to be that AssignNode throws InvalidNodeException? Maybe this needs to be refactored something like

CheckValid()

which does the if (isValid)/throw check

EnsureNodeExists

which does the rest, and then AssignNode can just be inlined into this operator= and avoid throwing.

jbeder (changes requested):

include/yaml-cpp/node/impl.h
TedLyngmo: Yes, in order to be able to make it `noexcept` I was trying to come up with something, but couldn't figure out how to do it without calling some function that could throw. Now that I'm looking at it, I see that I missed the throw in `is()`. I'll try to make something out of your suggestions.

TedLyngmo:
This one probably needs reading through once or twice more. I'll be testing it at work tomorrow. For some reason I'm not catching as many linting problems at home.

TedLyngmo:
Making move assignment `noexcept` was trickier than I anticipated and I hope I haven't messed it up completely. :-)

jpan127:
Hi, what's the status on this PR?

It seems like it is not only making the move special functions noexcept, but actually defining them for the first time.  It would be great if this library supported moves.

TedLyngmo:
> Hi, what's the status on this PR?

Howdy and thanks for showing interest in this one!

> It seems like it is not only making the move special functions noexcept, but actually defining them for the first time. It would be great if this library supported moves.

It's ongoing (even though it's not been updated for a some time). I'm on a rebase mission to prepare for this change actually, so I've put it on hold for now, but I will need to face it soon. I have a library with classes carrying a `YAML::Node` which makes moving my class objects potentially throwing, which will not be accepted in the long run, so something needs to be done.

I _think_ (@jbeder will have to pitch in) that the move constructor in this PR will never destroy any existing functionality, so that part of the PR could be separated and submitted on its own so that the complicated move assignment operator would be the only one left. Perhaps that'd be a way to take a step forward?

I've also throught about making a lot of test cases for things that this PR (or something like it) could break and deliver those long before this PR -  but the thing is that don't have that deep knowledge of the inner workings of this lib that I wish I had, so the test cases I come up with is stuff that I sort of stumbe on while trying out new approaches and fix anyway. I'd be good if someone else wrote them. If you have ideas for tests etc., please do PR's to this branch in my fork and we can make an effort together.

TedLyngmo:
Another PR would be one that makes a deliberate breaking change so that moving really means moving. There are one or two test cases (regarding Aliases) that refuses that - but I could make an outline for such a change. It'd break those test cases and perhaps programs making use of the functionality though.

As it is, it _feels_ (for lack of better words) that any program using the copy magic when moving relies on side effects that should not be trusted in the first place, but I could be wrong. I'd like such a PR better because it'd make moving a `YAML::Node` comprehensible and more in sync with the rest of the move semantics we're used to.

rr-mark:
I think this PR will resolve https://github.com/jbeder/yaml-cpp/issues/1209 and https://github.com/jbeder/yaml-cpp/issues/721. I think there are a number of assumptions made by std:: algorithms make about how move constructors behave which are not met by YAML::Nodes at present.

I briefly looked into adding `default` move constructors and move assignment, but this breaks a couple of unit tests (`NodeTest.SimpleAlias` and `NodeTest.TempMapVariableAlias`), and more generally leads to a difference in behaviour between copy assignment and move assignment, which probably has many downstream consequences.

SGSSGene:
@TedLyngmo Thanks you for working on this!
Please send a ping when this PR is ready for reviewing!

TedLyngmo:
> @TedLyngmo Thanks you for working on this! Please send a ping when this PR is ready for reviewing!

Thanks @SGSSGene! I haven't used yaml is some years now but it may soon be needed again so I thought of revisiting this old PR since I'm pretty sure potentially throwing move semantics won't be accepted in this project. I'm not sure I'll have the time or energy to finalize it though. :-/

I've at least put my AI agent to do a thorough code review on it. That's not something I had available 6 years ago :-)

Edit: Seems like there are some holes in my idea that probably prevents it from landing. A deliberate breaking change so that moving really means moving is probably the only realistic option.

SGSSGene:
I will close this PR for now. If anyone wants to pick this up or try a new approach, please feel free to do so.

PR #815, opened 2020-02-08
Fix plenty of clang warnings `declaration shadows a variable`: using class enum instead of enum

hyperxor:
I have got a lot warnings from clang compiler:
```
yaml-cpp/src/emitterstate.h:28:24: warning: declaration shadows a variable in namespace 'YAML' [-Wshadow]
  enum value { NoType, Flow, Block };
                       ^
```
and my proposal is to use enum classes.

### DIFF
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -677,7 +677,7 @@ Emitter& Emitter::Write(const std::string& str) {
     return *this;
 
   const bool escapeNonAscii = m_pState->GetOutputCharset() == EscapeNonAscii;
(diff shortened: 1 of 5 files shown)


hyperxor:
@jbeder Hello, unfortunatelly, this change breaks API, and shouldn't be done in minor version.
I run clang-format and push the following changes. In general, I think this change makes API clear and better, but I'm not hurry and it can be merged to master later.

rickyjames35:
I see this got closed by never made it into a release. Any chance we can get this in?

PR #817, opened 2020-02-10
Fix storing inf and NaN

T0ny0:
Tests that I'v added in commit fail without this fix.

### DIFF
--- a/test/node/node_test.cpp
+++ b/test/node/node_test.cpp
@@ -422,10 +422,64 @@ TEST(NodeTest, FloatingPrecisionFloat) {
   EXPECT_EQ(x, node.as<float>());
 }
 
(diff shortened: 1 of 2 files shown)


test/node/node_test.cpp
```diff
@@ -422,12 +422,66 @@ TEST(NodeTest, FloatingPrecisionFloat) {
...
   const double x = 0.123456789;
   Node node = Node(x);
   EXPECT_EQ(x, node.as<double>());
 }
 
+TEST(NodeTest, FloatingPrecisionPositiveInfinityDouble) {
+  if (!std::numeric_limits<double>::has_infinity) {
+    return;
+  }
+  const double x = -std::numeric_limits<double>::infinity();
```
jbeder: Should be +inf, right?

T0ny0: fixed

PR #820, opened 2020-02-13
Fix size() after removing node from sequence

JeffWegher:
I found a bug with sequence nodes wherein size() does not decrease when removing by index.

It was tricky because if node.size() had not been called prior to removing the node by index then the m_seqSize variable would never have been increased to begin with, so the first call to node.size() would then be correct. I added unit tests which call node.size() prior to calling node.remove(idx) which fail without the fix and pass with it.

The symptom of this was in a for loop like this:

```c++
for (size_t i=0; i < seqNode.size(); ++i) {
    if (seqNode[i].Scalar() == someValue) {
        seqNode.remove(i--);
    }
}
```
If there was more than one element removed in that loop, seqNode would convert to Map node!

### DIFF
--- a/include/yaml-cpp/node/detail/impl.h
+++ b/include/yaml-cpp/node/detail/impl.h
@@ -58,7 +58,9 @@ struct get_idx<Key, typename std::enable_if<std::is_signed<Key>::value>::type> {
 
 template <typename Key, typename Enable = void>
 struct remove_idx {
-  static bool remove(std::vector<node*>&, const Key&) { return false; }
+  static bool remove(std::vector<node*>&, const Key&, std::size_t*) {
+    return false;
+  }
 };
 
 template <typename Key>
@@ -66,11 +68,14 @@ struct remove_idx<
     Key, typename std::enable_if<std::is_unsigned<Key>::value &&
                                  !std::is_same<Key, bool>::value>::type> {
 
-  static bool remove(std::vector<node*>& sequence, const Key& key) {
+  static bool remove(std::vector<node*>& sequence, const Key& key,
+                     std::size_t* seqSize) {
     if (key >= sequence.size()) {
       return false;
     } else {
+      bool defined = sequence[key]->is_defined();
       sequence.erase(sequence.begin() + key);
+      *seqSize -= static_cast<size_t>(defined && *seqSize > key);
       return true;
     }
   }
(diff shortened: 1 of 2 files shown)


include/yaml-cpp/node/detail/impl.h
```diff
@@ -58,19 +58,24 @@ struct get_idx<Key, typename std::enable_if<std::is_signed<Key>::value>::type> {
...
 
-  static bool remove(std::vector<node*>& sequence, const Key& key) {
+  static bool remove(std::vector<node*>& sequence, const Key& key,
+                     std::size_t* seqSize) {
     if (key >= sequence.size()) {
       return false;
     } else {
+      bool defined = sequence[key]->is_defined();
       sequence.erase(sequence.begin() + key);
+      *seqSize -= static_cast<size_t>(defined && *seqSize > key);
```
jbeder: Prefer this as an "if" statement.

include/yaml-cpp/node/detail/impl.h
```diff
@@ -58,19 +58,24 @@ struct get_idx<Key, typename std::enable_if<std::is_signed<Key>::value>::type> {
 
 template <typename Key, typename Enable = void>
 struct remove_idx {
-  static bool remove(std::vector<node*>&, const Key&) { return false; }
+  static bool remove(std::vector<node*>&, const Key&, std::size_t*) {
```
jbeder: Pass the std::size_t as a reference, not a pointer.

include/yaml-cpp/node/detail/impl.h
TedLyngmo: Assuming `seqSize` will be a `std::size_t&` instead of a `std::size_t*`, this should make it clear:
```
      sequence.erase(std::next(sequence.begin(), key));
      seqSize = std::distance(sequence.begin(), sequence.end());
```

include/yaml-cpp/node/detail/impl.h
JeffWegher: That was my preference as well, I've just been on too many projects following the google style guide and I get told to change my references to pointers for output parameters :laughing:

include/yaml-cpp/node/detail/impl.h
JeffWegher: I've changed it to an if-statement, and also removed the condition of it being defined. I realized that if seqSize was already greater than the key/index, then the node at one point must have been defined and it's only a pathological case where someone specifically set the Node = undefined.

Issue #823, opened 2020-02-15
ambiguous call to overloaded function - Windows VS 2019 (latest)


Dekken:
```
p\\include\yaml-cpp/node/convert.h(161): note: while trying to match the argument list '(const unsigned __int64)'
p\\include\yaml-cpp/node/convert.h(163): error C2668: 'signbit': ambiguous call to overloaded function
C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.18362.0\\ucrt\corecrt_math.h(305): note: could be 'bool signbit(float) throw()'
C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.18362.0\\ucrt\corecrt_math.h(310): note: or       'bool signbit(double) throw()'
C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.18362.0\\ucrt\corecrt_math.h(315): note: or       'bool signbit(long double) throw()'

Dekken:
from: https://github.com/jbeder/yaml-cpp/pull/817

msuiche:
Yes. I do have the same problem here too.

TedLyngmo:
Reproducible example: https://godbolt.org/z/HQmy3y
Have you filed a bug report on MSVC?
It's supposed to do `std:.signbit(static_cast<double>(IntArg))` in these cases.

barcharcraz:
this isn't a quick fix, because we're not sure if the standard's mandate of integer overloads for the floating point related functions in cmath is intended.

I would strongly suggest not calling those overloads in libyaml

nihalar:
> this isn't a quick fix, because we're not sure if the standard's mandate of integer overloads for the floating point related functions in cmath is intended.
> 
> I would strongly suggest not calling those overloads in libyaml

How to fix it? I am facing the same problem.

TedLyngmo:
@barcharcraz The intent of paragraph [[`cmath.syn/2`]](https://eel.is/c++draft/cmath.syn#2) in the C++20 draft and in the C++17 standard seems pretty clear to me, but I see that the issue I raised is now tagged `LWG` so I suppose you'll propose a change in the upcoming standard to be applied backwards to affect C++17 too.

My "_quick fix(™)_" remark was meant to acknowledge that it may require more work than I anticipate to get this done even though it looks fairly straightforward.

nihalar:
[...]

PR #824, opened 2020-02-15
enable_if floating point to hide call to std::signbit 

Dekken:
with unsupportd parameters


Dekken:
fix for https://github.com/jbeder/yaml-cpp/issues/823

Dekken:
:+1:

PR #828, opened 2020-02-19
Upgrade to gtest 1.10

TedLyngmo:
This makes all warnings when building test cases go away.


TedLyngmo:
Ok, that was far from what I got locally.

PR #834, opened 2020-03-12
Update README.md

alanyee:
Add easy instructions for using the library without having to build from source


jbeder:
I don't think we need distribution-specific instructions.

Issue #623, opened 2018-09-04
Included gtest is too old to compile with MSVC 15.7 / VS 2017.15.7 onwards


staffantj:
Attempting to compile the embedded gtest fails under MSVC 15.7 (2017.15.7.x).

```
[CustomBuild] d:\buildagent1\work\dd2dd3ecd59185a2\test\gtest-1.8.0\googletest\include\gtest\internal\gtest-port.h(996, 0): error C2220: warning treated as error - no 'object' file generated [D:\BuildAgent1\work\dd2dd3ecd59185a2\v140\test\googletest_project-prefix\src\googletest_project-build\googlemock\gmock.vcxproj]
[11:20:16][CustomBuild] d:\buildagent1\work\dd2dd3ecd59185a2\test\gtest-1.8.0\googletest\include\gtest\internal\gtest-port.h(996, 0): warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning. [D:\BuildAgent1\work\dd2dd3ecd59185a2\v140\test\googletest_project-prefix\src\googletest_project-build\googlemock\gmock.vcxproj]
[11:20:16][CustomBuild]   c:\vs15\vc\tools\msvc\14.14.26428\include\tuple(1163): note: see declaration of 'std::tr1'
[11:20:16]
```

This was fixed in gtest-1.8.1 over a year ago [](https://github.com/google/googletest/commit/77380cddf77133b98a16b5427ac732648233de29)

staffantj:
[](https://github.com/google/googletest/commit/77380cddf77133b98a16b5427ac732648233de29)

staffantj:
https://github.com/google/googletest/commit/77380cddf77133b98a16b5427ac732648233de29

sigmasoldi3r:
Any directions to fix this?

I'm using also MSVC 17 and I've got 24 errors (C1083) related to googletest.

peterchen-cp:
I got the current head (b87c76a2ef4f974e2739a2d410ee5e36122cfc00) to build with the following changes:

 - for MSVC (15.9.6), install VS2015 toolset (v1400)
 - replace content of test/gtest-1.8.0/  with the content of gtest 1.8.1 (01148677a937eae1c0fd1e488ca81445531ea368)
[...]

PR #839, opened 2020-03-27
Solve compilation warnings and errors in VS2019

rdeterre:
Hi,

These changes solve warnings and error that show up when using Visual Studio 2019 with the C++ standard set to C++ 17.

The first commit updates the version of Google test in the "test" directory from v1.8.0 to v1.10.0 (related ticket: #623).

The second commit removes the use of `std::iterator` which is deprecated in C++ 17.  This was addressed before by #574 and #575, but `std::iterator` is used again the current master branch.

I tested everything with CMake here, but I'm not very familiar with Bazel, so I'm not sure if anything needs to be updated there.

Cheers,

Romain

### DIFF
--- a/include/yaml-cpp/node/detail/node_iterator.h
+++ b/include/yaml-cpp/node/detail/node_iterator.h
@@ -52,10 +52,7 @@ struct node_iterator_type<const V> {
 };
 
 template <typename V>
-class node_iterator_base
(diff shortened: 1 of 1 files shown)


jbeder:
Could you separate into a PR for Google test and the rest?

rdeterre:
No problem, this PR now only contains the change to remove the use of std::iterator.

PR #840, opened 2020-03-27
Update Google Test to v1.10.0

rdeterre:
This commit updates the version of Google Test from 1.8 to 1.10. This solves issues for compilation under Visual Studio 2019.

I tested everything with CMake here, but I'm not very familiar with Bazel, so I'm not sure if anything needs to be updated there.


jbeder:
I just noticed that this breaks on Linux/gcc:

The build for this commit:

https://travis-ci.org/github/jbeder/yaml-cpp/builds/668233704

with the log for Linux/gcc:

https://travis-ci.org/github/jbeder/yaml-cpp/jobs/668233706

I'm not sure how I didn't notice the failure on the PR. I'd roll back, but there are several dependent PRs that have already been merged, so it'll be a mess, and we may as well fix-forward.

Could you take a look?

jbeder:
I think this was just an old version of gcc. Fixed by updating to 4.9 in #845.

PR #841, opened 2020-03-31
Add vcpkg installation instructions

PhoebeHui:
yaml-cpp is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build yaml-cpp, ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for yaml-cpp and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,17 @@ cmake [-G generator] [-DYAML_BUILD_SHARED_LIBS=ON|OFF] ..
 
 5. To clean up, just remove the `build` directory.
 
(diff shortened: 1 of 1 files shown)


jbeder:
Thanks, but I think it's better to have this documentation outside of yaml-cpp.

PR #847, opened 2020-04-16
trigger installation through add_subdirectory from another CMakeLists.txt

deflinhec:
fix installation while building with add_subdirectory from another CMakeLists.txt.

As #741 mentioned. I'm not quite understand, why installation should be disable while used as a subdirectory?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,13 +13,11 @@ find_program(YAML_CPP_CLANG_FORMAT_EXE NAMES clang-format)
 option(YAML_CPP_BUILD_CONTRIB "Enable yaml-cpp contrib in library" ON)
 option(YAML_CPP_BUILD_TOOLS "Enable parse tools" ON)
 option(YAML_BUILD_SHARED_LIBS "Build yaml-cpp shared library" ${BUILD_SHARED_LIBS})
+option(YAML_CPP_INSTALL "Enable generation of yaml-cpp install targets" ON)
 
 cmake_dependent_option(YAML_CPP_BUILD_TESTS
   "Enable yaml-cpp tests" ON
   "BUILD_TESTING;CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
-cmake_dependent_option(YAML_CPP_INSTALL
-  "Enable generation of yaml-cpp install targets" ON
-  "CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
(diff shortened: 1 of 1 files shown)


jbeder:
I don't understand this PR. Could you rename the title so it describes what it does, and then fill in the body with the reason for the change?

deflinhec:
YAML_CPP_INSTALL should be optional rather than being disable while used as subdirectory.

jbeder:
Can you edit the title accordingly (which turns into the first line of the commit)?

Looking at the change, this appears to do something to the option YAML_CPP_INSTALL - it changes it from a dependent option to a non-dependent option? Anyways, that should be somewhere in your explanation.

CMakeLists.txt
```diff
@@ -13,13 +13,11 @@ find_program(YAML_CPP_CLANG_FORMAT_EXE NAMES clang-format)
 option(YAML_CPP_BUILD_CONTRIB "Enable yaml-cpp contrib in library" ON)
 option(YAML_CPP_BUILD_TOOLS "Enable parse tools" ON)
 option(YAML_BUILD_SHARED_LIBS "Build yaml-cpp shared library" ${BUILD_SHARED_LIBS})
+option(YAML_CPP_INSTALL "Enable generation of yaml-cpp install targets" ON)
```
madebr, line 16: This will always install yaml-cpp, also when included as a subproject, which is often undesired.
I would change the default to only install when yaml-cpp is the main project.

So something as follows:

Add this to the top of the cmake script:
```cmake
set(YAML_CPP_MAIN_PROJECT ON)
if(DEFINED PROJECT_SOURCE_DIR
    set(YAML_CPP_MAIN_PROJECT OFF)
endif()

# 3.5 is actually available almost everywhere, but this a good minimum
cmake_minimum_required(VERSION 3.4)
project(YAML_CPP VERSION 0.6.3 LANGUAGES CXX)
```

And then change this line to:
```suggestion
option(YAML_CPP_INSTALL "Enable generation of yaml-cpp install targets" ${YAML_CPP_MAIN_PROJECT})
```

This combines the best of both world:
- yaml-cpp is not installed (by default) when included as a subproject
- if it is really wanted, yaml-cpp can be installed by enabling the `YAML_CPP_INSTALL` cache definition.

SpaceIm:
I need this modification for conan-center-index, but PR seems to be stuck, so I've opened https://github.com/jbeder/yaml-cpp/pull/1007

jbeder:
Obviated by #1007.

PR #849, opened 2020-04-23
Add qtyaml.h

MartinDelille:
I'm a [Qtian](https://www.qt.io) and wondered if you would accept this wrapper found in [this gist](https://gist.github.com/brcha/d392b2fe5f1e427cc8a6)

@brcha I set you as the commit author for information.

### DIFF
--- a/include/yaml-cpp/qtyaml.h
+++ b/include/yaml-cpp/qtyaml.h
@@ -1,0 +1,167 @@
+/* ***** BEGIN LICENSE BLOCK *****
+ * Version: MPL 2.0
+ *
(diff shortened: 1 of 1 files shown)


jbeder:
I think this is better either in the QT project, or in a separate project.

MartinDelille:
Yes indeed: cloning the gist as a gitmodule is fine for me. Would you mind if I reference this gist in the README.md for other Qt users?

jbeder:
Sure; add a new section "related something something" (with a note that
these aren't explicitly supported).

PR #850, opened 2020-04-23
Include wiki content into a doc folder

MartinDelille:
I was reading the documentation and wonder if you would mind about including it into the main repository (it's always better to have the documentation sticked to the code).

I managed to do it by preserving the history thanks to [this SO answer](https://stackoverflow.com/a/1684694/118125).

Outside simplifying documentation contribution you could later enable a Github pages for the docs folder (I'll be happy to help you out if you want).


MartinDelille:
Maybe we should now redirect the wiki content to the docs folder in order to avoid to maintain duplicate content.

jbeder:
How?

sei-mwd:
This stackoverflow article seems to discuss this in depth.
https://stackoverflow.com/questions/31969868/how-to-store-github-wiki-as-part-of-source

MartinDelille:
A simple solution would be to replace the current wiki page content to the repository page.

For example, we could change to content of this page: https://github.com/jbeder/yaml-cpp/wiki/Tutorial

to

*The wiki has been disabled. The content is now available here: https://github.com/jbeder/yaml-cpp/blob/master/docs/Tutorial.md*

PR #855, opened 2020-04-27
implement convert for stl containers allowing for custom allocators

ibtaylor:
This extends `convert` for `list`, `vector`, and `map` to be useful to users venturing astray from the default stl container parameters.

For example, I ran into this while decoding into a vector with a custom allocator for alignment.
see: [Using STL Containers with Eigen](https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html)

For testing, I attempted to create the simplest allocator I could, named `CustomAllocator` which is just an adapter to `malloc` and `free`.

Similar issues could crop up in the usage of `std::string` rather than `std::basic_string`, but that seemed a lot more invasive and so I did not complete work on that.

### DIFF
--- a/test/node/node_test.cpp
+++ b/test/node/node_test.cpp
@@ -11,6 +11,38 @@
 
 #include <sstream>
 
+namespace {
+
+// malloc/free based allocator just for testing custom allocators on stl containers
+template <class T>
+class CustomAllocator : public std::allocator<T> {
+  public:
+    typedef std::size_t     size_type;
+    typedef std::ptrdiff_t  difference_type;
+    typedef T*              pointer;
+    typedef const T*        const_pointer;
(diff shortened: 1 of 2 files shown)


test/node/node_test.cpp
```diff
@@ -320,7 +348,8 @@ TEST(NodeTest, StdArrayWrongSize) {
 }
 
 TEST(NodeTest, StdVector) {
-  std::vector<int> primes;
```
jbeder, line 323: Can you leave these tests as-is, and then create duplicate test cases with the custom allocator?

ibtaylor:
Looks like line 211 needs to use the appropriate iterator. I'll update in a few.

ibtaylor:
Should be good now.

PR #857, opened 2020-05-07
Supplement test cases for 2.19 ~ 2.22 schema tags.

dota17:


test/integration/node_spec_test.cpp
```diff
@@ -237,7 +237,49 @@ TEST(NodeSpecTest, Ex2_18_MultiLineFlowScalars) {
               "So does this quoted scalar.\n");
 }
 
-// TODO: 2.19 - 2.22 schema tags
+TEST(NodeSpecTest, Ex2_19_Integers) {
```
jbeder, line 240: These don't actually test any tag-related behavior. I don't actually recall the details, so it's possible yaml-cpp doesn't support this. In any case, either test the tag behavior, or leave the TODO and remove this block of tests.

dota17: ok, I see. And I was reading the YAML v1.2 document recently and found that yaml-cpp didn't support the YAML version 1.2 perfectly in some places. I want to know whether you accept the related PR or have the related enhancement  plans?

perlpunk: If this is regarding resolving of types, maybe this repo can help: https://github.com/perlpunk/yaml-test-schema
The format might change, and it might go into https://github.com/yaml/yaml-test-suite at some point.
But I'm using this already for my YAML loader tests.

jbeder: @dota17 If you make the change I suggested (leave the TODO and remove this block of tests), then I'll accept the PR.

PR #858, opened 2020-05-07
Add filename to the BadFile exception

TedLyngmo:
The `BadFile` exception which is thrown when failing to open a file now
contains the filename.

Minor changes:
* `LoadFile()` and `LoadAllFromFile()` now passes the filename (which is a
 `std::string`) directly to the `std::ifstream` constructor which has been
 valid since C++11.
* Ran `clang-format` on `include/yaml-cpp/exceptions.h`

### DIFF
--- a/include/yaml-cpp/exceptions.h
+++ b/include/yaml-cpp/exceptions.h
@@ -136,7 +136,7 @@ inline const std::string INVALID_NODE_WITH_KEY(const std::string& key) {
   stream << "invalid node; first invalid key: \"" << key << "\"";
   return stream.str();
 }
-}
+}  // namespace ErrorMsg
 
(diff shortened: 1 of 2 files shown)


include/yaml-cpp/exceptions.h
```diff
@@ -281,7 +281,7 @@ class YAML_CPP_API EmitterException : public Exception {
 
 class YAML_CPP_API BadFile : public Exception {
  public:
-  BadFile() : Exception(Mark::null_mark(), ErrorMsg::BAD_FILE) {}
+  BadFile(const std::string& filename) : Exception(Mark::null_mark(), std::string(ErrorMsg::BAD_FILE) + ": " + filename) {}
```
jbeder, line 284: Mark this explicit.

PR #868, opened 2020-05-25
Update scanscalar.cpp

phrocker:
Removed damn.  happened to notice this through another project ( on a git submodule ) while skimming code to understand things. I use yaml-cpp on other projects and think it's great, so I thought I'd do my tiny little part. Totally fine if you think this is unnecessary or unwarranted.


jbeder:
I think it's fine to keep this. Adds a bit of color.

PR #873, opened 2020-05-26
Fix spelling errors identified by spell scanning

dota17:


jbeder:
These are all in googletest, which is just pulled directly from their source; please file this PR with them, and we'll get the fixes that way.

dota17:
oops, my fault!

PR #876, opened 2020-06-03
clang-tidy cleanups

neheb:


dota17:
It's good!

src/directives.cpp
```diff
@@ -1,11 +1,11 @@
 #include "directives.h"
 
 namespace YAML {
-Directives::Directives() : version{true, 1, 2}, tags{} {}
```
jbeder, line 4: Prefer to keep these explicit initialize statements (throughout).

src/emitterstate.cpp
```diff
@@ -94,15 +89,13 @@ EmitterNodeType::value EmitterState::NextGroupType(
   if (type == GroupType::Seq) {
     if (GetFlowType(type) == Block)
       return EmitterNodeType::BlockSeq;
-    else
```
jbeder, line 97: This is a 2x2 grid of results, and probably worth keeping as-is.

src/node_data.cpp
```diff
@@ -209,12 +206,9 @@ node* node_data::get(node& key, shared_memory_holder /* pMemory */) const {
     return nullptr;
   }
 
-  for (node_map::const_iterator it = m_map.begin(); it != m_map.end(); ++it) {
-    if (it->first->is(key))
-      return it->second;
-  }
-
-  return nullptr;
+  auto it = std::find_if(m_map.begin(), m_map.end(), [&](const std::pair<node*, node*> node)
```
jbeder, line 209: I think this still reads better as a for loop, but perhaps you can change it to:

for (const auto& kv : m_map) {
  if (kv.first->is(key)) {
    return it->second;
  }
}
return nullptr;

src/node_data.cpp
```diff
@@ -230,9 +224,10 @@ node& node_data::get(node& key, shared_memory_holder pMemory) {
       throw BadSubscript(m_mark, key);
   }
 
-  for (node_map::const_iterator it = m_map.begin(); it != m_map.end(); ++it) {
-    if (it->first->is(key))
-      return *it->second;
+  auto it = std::find_if(m_map.begin(), m_map.end(), [&](const std::pair<node*, node*> node)
```
jbeder, line 227: same here

src/nodeevents.cpp
```diff
@@ -32,8 +32,8 @@ void NodeEvents::Setup(const detail::node& node) {
     return;
 
   if (node.type() == NodeType::Sequence) {
-    for (detail::const_node_iterator it = node.begin(); it != node.end(); ++it)
-      Setup(**it);
+    for (auto && it : node)
```
jbeder, line 35: Why && instead of const&?

src/nodeevents.cpp
```diff
@@ -77,9 +77,8 @@ void NodeEvents::Emit(const detail::node& node, EventHandler& handler,
       break;
     case NodeType::Sequence:
       handler.OnSequenceStart(Mark(), node.tag(), anchor, node.style());
-      for (detail::const_node_iterator it = node.begin(); it != node.end();
-           ++it)
-        Emit(**it, handler, am);
+      for (auto && it : node)
```
jbeder, line 80: ditto

src/nodeevents.cpp
neheb: No idea. clang-tidy did this.

src/node_data.cpp
neheb: Sure. Using algorithms seems to be preferred by some. They result in better code generation sometimes. I’ve seen that with std::accumulate.

neheb:
Ran all commits through git clang-format.

neheb:
Removed member init commit.

include/yaml-cpp/node/detail/impl.h
```diff
@@ -125,13 +127,11 @@ inline node* node_data::get(const Key& key,
       throw BadSubscript(m_mark, key);
   }
 
-  for (node_map::const_iterator it = m_map.begin(); it != m_map.end(); ++it) {
-    if (it->first->equals(key, pMemory)) {
-      return it->second;
-    }
-  }
+  auto it = std::find_if(m_map.begin(), m_map.end(), [&](const kv_pair m) {
```
jbeder, line 130: I still prefer the for loop.

include/yaml-cpp/node/convert.h
```diff
@@ -255,9 +252,7 @@ template <typename T, typename A>
 struct convert<std::list<T,A>> {
   static Node encode(const std::list<T,A>& rhs) {
     Node node(NodeType::Sequence);
-    for (typename std::list<T,A>::const_iterator it = rhs.begin();
-         it != rhs.end(); ++it)
-      node.push_back(*it);
+    std::copy(rhs.begin(), rhs.end(), std::back_inserter(rhs));
```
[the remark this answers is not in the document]
neheb: I did. That's embarrassing.

Issue #752, opened 2019-10-02
yaml-cpp Emitter generates bad long keys


till-s:
This map is parsed correctly

    map1: &map_anch
      key: value
    map2: *map_anch

but the parsed hierarchy is re-emitted (when using long-key format) as

    ? map1
    : &1 ? key
      : value
    ? map2
    : *1

which cannot be parsed again. The correct output would be (newline after the anchor &1)

    ? map1
    : &1
      ? key
      : value
    ? map2
    : *1

A simple program reproducing the bug (just remove the '.txt' extension): 
[bug-longkey.cc.txt](https://github.com/jbeder/yaml-cpp/files/3679111/bug-longkey.cc.txt)

till-s:
See also [this issue](https://github.com/jbeder/yaml-cpp/issues/707)

dota17:
Fixed by #879.

PR #879, opened 2020-06-15
fix issue752: generate right long keys

dota17:
#752

### DIFF
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -625,14 +625,14 @@ TEST_F(EmitterTest, GlobalLongKeyOnSeq) {
   out << Key << Alias("key") << Value << Alias("value");
   out << EndMap;
 
-  ExpectEmit("? &key\n"
-             "  - a\n"
-             "  - b\n"
-             ": &value\n"
-             "  - c\n"
-             "  - d\n"
-             "? *key\n"
-             ": *value");
+  ExpectEmit(R"(? &key
+  - a
+  - b
+: &value
+  - c
+  - d
+? *key
+: *value)");
 }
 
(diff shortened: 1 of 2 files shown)


test/integration/emitter_test.cpp
```diff
@@ -612,6 +612,38 @@ TEST_F(EmitterTest, SimpleGlobalSettings) {
...
+  out << Key << Anchor("key");
+  out << BeginSeq << "a"
+      << "b" << EndSeq;
+  out << Value << Anchor("value");
+  out << BeginSeq << "c"
+      << "d" << EndSeq;
+  out << Key << Alias("key") << Value << Alias("value");
+  out << EndMap;
+
+  ExpectEmit("? &key\n  - a\n  - b\n: &value\n  - c\n  - d\n? *key\n: *value");
```
jbeder, line 628: Can you use raw literals here? I know the other examples don't do this, but they were from before that feature was widely accepted. I think it would be more readable here.

dota17: Do you mean raw string literal?  I try it
```
  ExpectEmit(
R"(? &key
  - a
  - b
: &value
  - c
  - d
? *key
: *value)");
```
haha, It is intuitional, but the indents and alignment look like not confortable. 
```
  ExpectEmit("? &key\n"
             "  - a\n"
             "  - b\n"
             ": &value\n"
             "  - c\n"
             "  - d\n"
             "? *key\n"
             ": *value");
```
This format is both readable and intuitional, but the escape char is not good.
What's you opinoin about this?

jbeder: I’d go with the R style. The first line doesn’t line up right, but subsequent ones do and that’s probably fine.

PR #888, opened 2020-06-15
[clang-tidy] use auto

neheb:
Found with modernize-use-auto

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/src/directives.cpp
+++ b/src/directives.cpp
@@ -5,7 +5,7 @@ Directives::Directives() : version{true, 1, 2}, tags{} {}
 
 const std::string Directives::TranslateTagHandle(
     const std::string& handle) const {
-  std::map<std::string, std::string>::const_iterator it = tags.find(handle);
+  auto it = tags.find(handle);
   if (it == tags.end()) {
     if (handle == "!!")
       return "tag:yaml.org,2002:";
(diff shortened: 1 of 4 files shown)


src/stream.cpp
```diff
@@ -151,7 +151,7 @@ inline UtfIntroCharType IntroCharTypeOf(std::istream::int_type ch) {
 
 inline char Utf8Adjust(unsigned long ch, unsigned char lead_bits,
                        unsigned char rshift) {
-  const unsigned char header =
+  const auto header =
```
jbeder, line 154: Leave this one as-is; I don’t think auto helps here.

neheb: This was automatically done. The idea is to avoid duplicating the type name since there's a static_cast.

jbeder: I know, but I don’t agree with the philosophy. I prefer auto only when (1) the type is obvious and (2) the type name is unwieldy to write. This matches (1) but not (2).

neheb: OK. Added nolint.

jbeder: Could you keep off the NOLINT? I don’t think that adds anything either, since there are so many different linters we could use.

neheb: OK.

PR #893, opened 2020-06-15
partially fix clang compilation

neheb:
Missing header and mistaken algorithm usage.

Also removed it name from range loops. It's not correct.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


neheb:
Actually this fully fiixes it.

PR #903, opened 2020-06-22
reset flags node at the EndedGroup method.

### LINKED ISSUE #902: a YAML::NewLine at the end of a sequence brake the next node
eg:

having a sort of this code snippet:

```
out << YAML::Key << "k" << YAML::Value
					<< YAML::Flow << YAML::BeginSeq;
				for (int i = 0; i < 2; i++) {
					out << YAML::BeginMap
						<< YAML::Key << "i" << YAML::Value << i
						<< YAML::EndMap 
						<< YAML::Newline
						;
				}
				out << YAML::EndSeq;
```

if that node is running through another outer loop of Begin/End Map is generating a sort of this result:
```yml
nodeA:   
    k: [{i: 0},
 {i:1},
  ]NodeB:
    k: [{i: 0},
 {i:1},
  ]

```
instead without the `YAML::NewLine` is:
```yml
nodeA:  
    k: [{i: 0}, {i:1}]
  NodeB:
    k: [{i: 0}, {i:1}]
```

it seems that the last new line before ending the sequence should be "ignored" and is not, breaking the next node alignment and resulting in an invalid YAML file.

of course there is a workaround in the code, and just avoid the last new line to be emitted.

but is very not elegant and require to do n-1 in the loop and repeat the inside loop block without the newline,
a better solution would be nice if the `yaml-cpp` library handle that edge case, basically removing the newline (or ignoring it) when the `EndSeq` is emitted.

---

Raffaello:
I have written the tests related to the #902 issue 
and fix it,

in this case would be enough to reset 1 flag, but i extended with all 3 as a derived semantic.

hope it is correct.


test/integration/emitter_test.cpp
```diff
@@ -1212,6 +1212,42 @@ TEST_F(EmitterTest, NaN) {
...
+      << YAML::Newline;
+  out << YAML::BeginMap << YAML::Key << "i" << YAML::Value << 1 << YAML::EndMap
+      << YAML::Newline;
+  out << YAML::EndSeq;
+  out << YAML::EndMap;
+
+  out << YAML::EndMap;
+
+  ExpectEmit(
+      "NodeA:\n"
```
jbeder, line 1239: Can you make this a multiline string (`R"(...)"`)?

test/integration/emitter_test.cpp
```diff
@@ -1212,6 +1212,42 @@ TEST_F(EmitterTest, NaN) {
 	  "bar: .nan");
 }
 
+TEST_F(EmitterTest, ComplexFlowSeqEmbeddingAMapWithNewLine) {
```
jbeder, line 1215: Could you add a test similar to this example, but with an alias (or a tag)? That would help determine whether this is a correct change.

src/emitterstate.cpp
```diff
@@ -175,8 +175,11 @@ void EmitterState::EndedGroup(GroupType::value type) {
   // some global settings that we changed may have been overridden
   // by a local setting we just popped, so we need to restore them
   m_globalModifiedSettings.restore();
-
+
```
dota17, line 178: remove the extra whitespace.

test/integration/emitter_test.cpp
Raffaello: do you mean replacing with this instead?

```
ExpectEmit(R"(NodeA:
  k: [{i: 0},
    {i: 1},
    ]
NodeB:
  k: [{i: 0},
    {i: 1},
    ])");
```

test/integration/emitter_test.cpp
Raffaello: i have done with alias;
for the tag test, i notice there is no too much documentation neither and it makes me quite difficult translate this minimal test to an equivalent using a tag (local?), basically i need some help/ more information around YAML tags to do it.

jbeder: Yeah, I think just using `LocalTag("foo")` should work fine; it should produce `!foo` as a prefix to whichever node you're applying it to.

test/integration/emitter_test.cpp
```diff
@@ -1212,6 +1212,83 @@ TEST_F(EmitterTest, NaN) {
...
+  out << YAML::Newline;
+  out << EndSeq << EndMap;
+
+  out << Key << "NodeA" << Alias("Node");
+  out << Key << "NodeB" << Alias("Node");
+
+  out << EndMap;
+
+  ExpectEmit(R"(Node: &Node
+  k: [{i: 0},
```
jbeder, line 1265: Can put an anchor on each of these nodes (the sequence, each map in the sequence, and both key and value within each map)?

(And when you do tags, please do the same for that.)

Raffaello: This kind of yaml file not sure is possible to emit:

```yml
Aliases:
 - {&i i: &v0 0}
 - {&i i: &v1 1}
NodeA:
  k: [{*i : *v0}, {*i : *v1}]
NodeB:
  k: [*i : *v0, *i : *v1]
```

should be a valid yaml, using to validate: https://yamlvalidator.com/


the actual test that is failing will be:
```c
TEST_F(EmitterTest, ComplexFlowSeqEmbeddingAMapUsingAliases4) {
  out << BeginMap;

  out << Key << "Aliases" << Value << BeginSeq;
  out << BeginMap << Anchor("i") << Key << "i" << Anchor("v0") << Value << 0
      << EndMap;
  out << BeginMap << Key << Anchor("i") << "i" << Value << Anchor("v1") << 1
      << EndMap;
  out << EndSeq;

  out << Key << "NodeA" << Value << BeginMap;
  out << Key << "k" << Value << Flow << BeginSeq;
  out << BeginMap << Key << Alias("i") << Value << Alias("v0") << EndMap;
  out << BeginMap << Key << Alias("i") << Value << Alias("v1") << EndMap;
  out << EndSeq << EndMap;

  out << Key << "NodeB" << Value << BeginMap;
  out << Key << "k" << Value << Flow << BeginSeq;
  out << Key << Alias("i") << Value << Alias("v0");
  out << Key << Alias("i") << Value << Alias("v1");
  out << EndSeq << EndMap;

  out << EndMap;

  ExpectEmit(R"(Aliases:
 - {&i i: &v0 0}
 - {&i i: &v1 1}
NodeA:
  k: [{*i : *v0}, {*i : *v1}]
NodeB:
  k: [*i : *v0, *i : *v1])");
}
```

This test is going to fail with the actual master code, so should be considered in a different issue/PR to avoid scope creep in this one. Of course there analog test checking the newline will fail too.


Let me know your thoughts, i can commit anyway including this pair of tests and the CI should fail too.


p.s.
this test is the one asked for ` both key and value within each map`
[maybe consider a better naming for the review too, at this moment i am not in a really poetic mind form :)]

test/integration/emitter_test.cpp
```diff
@@ -1212,6 +1212,296 @@ TEST_F(EmitterTest, NaN) {
...
+  out << EndMap;
+
+  ExpectEmit(R"(Keys: [&k0 {i: 0}, &k1 {i: 1}]
+NodeA:
+  k: [*k0, *k1]
+NodeB:
+  k: [*k0, *k1])");
+}
+
+TEST_F(EmitterTest, ComplexFlowSeqEmbeddingAMapWithNewLineUsingAliases4) {
```
Raffaello, line 1433: [this test failing i think for a separate bug for what i am trying to fix in this PR]

test/integration/emitter_test.cpp
```diff
@@ -1212,6 +1212,296 @@ TEST_F(EmitterTest, NaN) {
...
+  k: [{*i : *v0},
+  {*i : *v1},
+  ]
+NodeB:
+  k: [*i : *v0,
+  *i : *v1,
+  ])");
+}
+
+TEST_F(EmitterTest, ComplexFlowSeqEmbeddingAMapUsingAliases4) {
```
Raffaello, line 1472: [this test failing i think for a separate bug for what i am trying to fix in this PR]

jbeder: I agree, thanks for finding this bug. Could you remove this test and file a new bug with this (failing) test case?

test/integration/emitter_test.cpp
jbeder: Thanks for investigating. See below; if you remove the failing test, this PR looks good.

Issue #904, opened 2020-06-25
Fails on MSVC without NOMINMAX defined.


Dekken:
https://github.com/jbeder/yaml-cpp/blob/97d1ae547cc9935cf851801ba0a247b4be2191de/include/yaml-cpp/node/convert.h#L123

PR #905, opened 2020-06-25
refactor to allow MSVC without #define NOMINMAX

Dekken:
fix for https://github.com/jbeder/yaml-cpp/issues/904

reference: https://github.com/microsoft/cppwinrt/issues/479

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -120,8 +120,8 @@ typename std::enable_if<(std::is_same<T, unsigned char>::value ||
 ConvertStreamTo(std::stringstream& stream, T& rhs) {
   int num;
   if ((stream >> std::noskipws >> num) && (stream >> std::ws).eof()) {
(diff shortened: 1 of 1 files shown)


include/yaml-cpp/node/convert.h
```diff
@@ -120,15 +120,16 @@ typename std::enable_if<(std::is_same<T, unsigned char>::value ||
...
+    if (num >= (std::numeric_limits<T>::min)() &&
+        num <= (std::numeric_limits<T>::max)()) {
       rhs = num;
       return true;
     }
   }
   return false;
 }
 
+
```
jbeder, line 132: Stray extra newline?

Dekken: oh yes one moment

Dekken: fix

Dekken: fixed

Issue #137, opened 2015-03-30
Ability to change how YAML::Null is emitted


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. emmiter << YAML::Null

What is the expected output? What do you see instead?

{{{
~
}}}

What version of the product are you using? On what operating system?
0.2.7

Please provide any additional information below.

It'd be nice if you could modify how YAML::Null is represented, I have a file I 
have to generate that way:

{{{
- foo: 123
- bar:
- baz: "yeah"
}}}

To mean, "no values" for "bar"... right now this emits "- bar: ~".
```

Original issue reported on code.google.com by `philippe...@gmail.com` on 29 Nov 2011 at 4:16

outofthecave:
It would be nice if one could choose between all valid representations of `null` in YAML, i.e. `~`, `null`, `Null`, `NULL`, or no output, see http://yaml.org/type/null.html.

jbeder:
Fixed #906.

PR #906, opened 2020-06-28
Support kinds of emitter format-setting for YAML::Null.

dota17:
Add four emitter styles for YAML::Null,  like  #137 
```C++
  LowerNull   // null
  UpperNull   // NULL
  CamelNull   // Null
  TildeNull   // ~
```


src/emitter.cpp
```diff
@@ -893,7 +912,8 @@ Emitter& Emitter::Write(const _Null& /*null*/) {
 
   PrepareNode(EmitterNodeType::Scalar);
 
-  m_stream << "~";
+  const char* name = ComputeNullName();
+  m_stream << name;
```
jbeder, line 916: Just inline the function call.

src/emitter.cpp
```diff
@@ -770,6 +774,21 @@ const char* Emitter::ComputeFullBoolName(bool b) const {
                          // these answers
 }
 
+const char* Emitter::ComputeNullName() const {
+  const EMITTER_MANIP nullFmt = m_pState->GetNullFormat();
```
jbeder, line 778: Inline this in the switch (*) statement.

src/emitter.cpp
```diff
@@ -770,6 +774,21 @@ const char* Emitter::ComputeFullBoolName(bool b) const {
...
+  switch (nullFmt) {
+    case LowerNull:
+      return "null";
+    case UpperNull:
+      return "NULL";
+    case CamelNull:
+      return "Null";
+    case TildeNull:
+    default:
```
jbeder, line 787: Explicitly write

// fallthrough

test/integration/emitter_test.cpp
```diff
@@ -1047,6 +1047,37 @@ TEST_F(EmitterTest, BoolFormatting) {
...
+  out.SetNullFormat(LowerNull);
+  out << Null;
+  out.SetNullFormat(UpperNull);
+  out << Null;
+  out.SetNullFormat(CamelNull);
+  out << Null;
+  out.SetNullFormat(TildeNull);
+  out << Null;
+  out << EndSeq;
+  ExpectEmit("- null\n- NULL\n- Null\n- ~");
```
jbeder, line 1061: I know the related tests don't do this, but I think it'd be easier to read if this sequence is written in flow style (and below).

test/integration/emitter_test.cpp
```diff
@@ -1047,6 +1047,37 @@ TEST_F(EmitterTest, BoolFormatting) {
...
+  out << LowerNull << Null;
+  out << UpperNull << Null;
+  out << CamelNull << Null;
+  out << TildeNull << Null;
+  out << EndSeq;
+  ExpectEmit("- null\n- NULL\n- Null\n- ~");
+}
+
+TEST_F(EmitterTest, NullFormattingOnNode) {
+  Node n(Load("null"));
```
jbeder, line 1075: Can you run through all four cases here too?

src/emitter.cpp
```diff
@@ -770,6 +774,21 @@ const char* Emitter::ComputeFullBoolName(bool b) const {
...
+  switch (m_pState->GetNullFormat()) {
+    case LowerNull:
+      return "null";
+    case UpperNull:
+      return "NULL";
+    case CamelNull:
+      return "Null";
+    case TildeNull:
+      return "~";
+    default:  // fallthrough
```
jbeder, line 787: What I meant was:

```
case TildeNull:
  // fallthrough
default:
  return "~";
```

This is the same as the previous behavior you had, but with the comment.

What you have now would not return any value in the default case, which would be undefined behavior.

dota17: haha, I misunderstanded your meaning.

Issue #607, opened 2018-07-05
Carriage return not escaped


yxliang01:
When the character `'\r'` is in the input string, it didn't get escaped.
Example code to reproduce:

```c
YAML::Emitter out;
  out << YAML::BeginMap;
  out << YAML::Key << "input";
  out << YAML::Value << std::string("\rabc").c_str();
  out << YAML::EndMap;

  assert(out.good());
std::cout << out.c_str();
```
The output would produce `input: ^Mabc` where `^M` represents `'\r'` without escaping. 
`out.SetOutputCharset(YAML::EscapeNonAscii);` doesn't change this bahaviour.

However, when `out.SetStringFormat(YAML::DoubleQuoted)` is used, it escapes.

Thanks

SGSSGene:
This is fixed in current `yaml-cpp-0.9.0` release. (Might have been part of an earlier release, not sure).

PR #910, opened 2020-06-30
Support to escape CR (\r) in load/emit

dota17:
#607

### DIFF
--- a/src/exp.h
+++ b/src/exp.h
@@ -69,7 +69,7 @@ inline const RegEx& Hex() {
 inline const RegEx& NotPrintable() {
   static const RegEx e =
       RegEx(0) |
-      RegEx("\x01\x02\x03\x04\x05\x06\x07\x08\x0B\x0C\x7F", REGEX_OR) |
+      RegEx("\x01\x02\x03\x04\x05\x06\x07\x08\x0B\x0C\x0D\x7F", REGEX_OR) |
       RegEx(0x0E, 0x1F) |
       (RegEx('\xC2') + (RegEx('\x80', '\x84') | RegEx('\x86', '\x9F')));
   return e;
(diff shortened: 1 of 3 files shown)


src/exp.h
```diff
@@ -69,7 +69,7 @@ inline const RegEx& Hex() {
 inline const RegEx& NotPrintable() {
   static const RegEx e =
       RegEx(0) |
-      RegEx("\x01\x02\x03\x04\x05\x06\x07\x08\x0B\x0C\x7F", REGEX_OR) |
+      RegEx("\x01\x02\x03\x04\x05\x06\x07\x08\x0B\x0C\x0D\x7F", REGEX_OR) |
```
jbeder, line 72: `0xD` is considered printable, according to the spec:

https://yaml.org/spec/1.2/spec.html#printable%20character//

dota17: Yes, you are right. It is not reasonable to place `\x0D` in `NotPrintable()`. 
May I establish a  new `Exp::CR()` or  `Exp::CarriageReturn()`  to achieve the escape of `\r`, like the `Exp::Tab()`?

https://github.com/jbeder/yaml-cpp/blob/3f381f13a07f007941b250e1563f67ded4dc8b1a/src/emitterutils.cpp#L179

jbeder: I don't think this is correct; where in the spec does it disallow \r from appearing in a plain scalar?

dota17: I have two reasons to do this:
1.  in the 5.1 chapter of spec, it says `Any excluded characters must be presented using escape sequences. In addition, any allowed characters known to be non-printable should also be escaped. `. I think `CR` belong to this.
2.  If the `\r` is allow to appear in a plain scallar, the "aaa\rbbb" will be parsed as "bbb". It may be not an expected result.

Issue #899, opened 2020-06-20
Hang in parser


fuzzah:
Hi.
This simple string (file) results in hang in utils/parse: `[:]`
as well as this one: `{:]`
To reproduce one can use:
```
echo "[:]" | ./parse
```
 
This is as of 5a9ab177bad9665ac5eb8f5b12435052123b5b6f

dota17:
This problem is similar to #740, I'm trying to fix it.

jbeder:
Fixed by #911.

Issue #740, opened 2019-09-19
scantoken.cpp:ScanPlainScalar() seg fault on bad input


ClarkColeman:
Using bad input generated by a fuzzer, a simple driver program that asks yaml-cpp to do a Node::Load() will terminate with either a seg fault or a SIGKILL. Snapshots of the backtrace in gdb indicate that there is no infinite recursion; the run-time stack never gets more than about 25 stack frames in depth.

Instead, it seems that memory allocations accumulate due to the last line of ScanPlainScalar(), where we see m_tokens.push() called.

Just before the end of ScanPlainScalar(), there is a commented out conditional exception throw that was probably intended to handle this case:

```
  // finally, check and see if we ended on an illegal character
  // if(Exp::IllegalCharInScalar.Matches(INPUT))
  //    throw ParserException(INPUT.mark(), ErrorMsg::CHAR_IN_SCALAR);
```

This commenting out was done in a 2008 commit:

[](https://github.com/jbeder/yaml-cpp/commit/4cfa233888da422d8e6a6211fc0e40f28ab1eb33)

It is not possible to simply reactivate the commented-out code, nor to revert to the older code, because neither Exp::IllegalCharInScalar() nor the previous Exp::IllegalColonInScalar() exists in the current code base.

An input file that causes the error is:

[ascii-input0.txt](https://github.com/jbeder/yaml-cpp/files/3631660/ascii-input0.txt)

Line 38 is the cause of the problem:

   `<< : [:]ore-max-padding, *auto-init-128]`

This was originally, before fuzzing:

   `<< : [*more-max-padding, *auto-init-128]`

PR #911, opened 2020-07-01
fix issue740: parse colon in plain scalar correctly

dota17:
According to the yaml specification, a colon can be seen as a part of a plain scalar if it is not followed by white space in most cases.

https://github.com/jbeder/yaml-cpp/blob/3f381f13a07f007941b250e1563f67ded4dc8b1a/src/exp.h#L167

 In function EndScalarInFlow() , `RegEx(':')+RegEx(",]}", REGEX_OR) `means if a ":" is followed by a  ","  or  " ] "   or  " }", the ":" will not be a part of a plain scalar. It is against the spec, so I removed it, and this solved the problem of #740 and #899.

`*ex7_3` and `*ex7_17` in `specexamples.h` have a little difference with the spec content, This difference will cause the corresponding testcases can't work as expected because of the new rule. I changed them to be the same as spec.

### DIFF
--- a/test/specexamples.h
+++ b/test/specexamples.h
@@ -552,7 +552,7 @@ const char *ex7_2 =
 
 const char *ex7_3 =
     "{\n"
-    "  ? foo : ,\n"
+    "  ? foo :,\n"
     "  : bar,\n"
     "}\n";
 
@@ -641,8 +641,8 @@ const char *ex7_17 =
     "{\n"
     "unquoted : \"separate\",\n"
     "http://foo.com,\n"
-    "omitted value: ,\n"
-    " : omitted key,\n"
+    "omitted value:,\n"
+    ": omitted key,\n"
     "}";
 
 const char *ex7_18 =
--- a/test/specexamples.h
+++ b/test/specexamples.h
@@ -552,7 +552,7 @@ const char *ex7_2 =
 
 const char *ex7_3 =
     "{\n"
-    "  ? foo :,\n"
+    "  ? foo : ,\n"
     "  : bar,\n"
     "}\n";
 
@@ -641,8 +641,8 @@ const char *ex7_17 =
     "{\n"
     "unquoted : \"separate\",\n"
     "http://foo.com,\n"
-    "omitted value:,\n"
-    ": omitted key,\n"
+    "omitted value: ,\n"
+    " : omitted key,\n"
     "}";
 
 const char *ex7_18 =
--- a/src/exp.h
+++ b/src/exp.h
@@ -110,7 +110,7 @@ inline const RegEx& Value() {
   return e;
 }
 inline const RegEx& ValueInFlow() {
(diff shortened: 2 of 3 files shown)


perlpunk:
The spec [ns-plain-char](https://yaml.org/spec/1.2/spec.html#ns-plain-char(c)) says:
```
[129] ns-plain-safe-in ::= ns-char - c-flow-indicator
[130] ns-plain-char(c) ::= ( ns-plain-safe(c) - “:” - “#” )
                         | ( /* An ns-char preceding */ “#” )
                         | ( “:” /* Followed by an ns-plain-safe(c) */ )
```
That means, I believe, that for example a colon followed by a comma or another flow-indicator will end a plain scalar in flow.
So, as far as I understand, this PR will make this example (7.17) invalid (you added spaces in `test/specexamples.h`):

```
{
unquoted : "separate",
http://foo.com,
omitted value:,  # colon + comma
: omitted key,
}
```
But this is valid (it's an example in the spec after all), and part of the YAML Test Suite:
https://github.com/yaml/yaml-test-suite/blob/master/test/4ABK.tml

dota17:
Ok, I see. @perlpunk ,thanks for your explaination. I have a misunderstanding about this and confuse the `“°”` with `“·”`. I will close this PR.

test/specexamples.h
```diff
@@ -641,8 +641,8 @@ const char *ex7_17 =
     "{\n"
     "unquoted : \"separate\",\n"
     "http://foo.com,\n"
-    "omitted value:,\n"
-    ": omitted key,\n"
+    "omitted value: ,\n"
```
perlpunk, line 644: If you have a closer look at the spec example 7.17, you'll notice that it contains
```
omitted value:°,
```
where the `°` stands for the empty node, so it should not be replaced with a space.

dota17: ha-ha, yes. And if one is not careless, it's easy to confuse them, just like me.

perlpunk:
It would be cool to fix #899 though.
When I ran that recently in [yaml-runtimes](https://github.com/yaml/yaml-runtimes) (generating the https://matrix.yaml.io/) it froze my computer, and I had to hard reset it, so I'm seriously considering taking out yaml-cpp from the test matrix :-/

dota17:
ok, I will solve this problem as soon as possible

dota17:
@perlpunk Should processor throw an exception when it parses the `[:]`? Or, parse it as `[~: ~]`. I know that pyyaml dose throw an exception.

perlpunk:
@dota17 it is actually an edge case.
The [reference parser](https://hackage.haskell.org/package/YamlReference) and [HsYAML](https://github.com/haskell-hvr/HsYAML) and a couple of other libraries parse that the same as `[ null : null ]`, but since it is an edge case and noone should write YAML like that, I think it's not that important what the result is, as long as it does not hang.

dota17:
Update the solution and testcases for the bug in issue741/899. The previous solution has been deleted.

Issue #912, opened 2020-07-02
Unable to configure non-DLL build (/MT, not /MD) with CMake 3.17.3 and VS 2019


tomlogic:
Disclaimer: Although I'm an experienced embedded C programmer, I'm relatively new to CMake, Visual Studio and building Windows libraries.

I'm trying to find a CMake configuration of `yaml-cpp` to get a 32-bit Visual Studio 2019 solution with the runtime library set to `/MT` and `/MTd`.

I'm using the following to configure my build, and even added some debug output to `CMakeLists.txt` to verify the `MSVC_RUNTIME_LIBRARY` setting:

```
$ cmake .. -AWin32 -DYAML_MSVC_SHARED_RT=OFF -DYAML_BUILD_SHARED_LIBS=OFF
-- Building for: Visual Studio 16 2019
-- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.
-- The CXX compiler identification is MSVC 19.26.28806.0
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.26.28801/bin/Hostx64/x86/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.26.28801/bin/Hostx64/x86/cl.exe - works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- CMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>$<$<BOOL:OFF>:DLL>"
--
 Properties for TARGET yaml-cpp:
   yaml-cpp.MSVC_RUNTIME_LIBRARY = "MultiThreaded$<$<CONFIG:Debug>:Debug>$<$<BOOL:OFF>:DLL>"

-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - not found
-- Found Threads: TRUE
-- The C compiler identification is MSVC 19.26.28806.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.26.28801/bin/Hostx64/x86/cl.exe
[...]

PR #913, opened 2020-07-02
msvc: enable CMP0091 to use MSVC_RUNTIME_LIBRARY

tomlogic:
Per https://cmake.org/cmake/help/latest/policy/CMP0091.html, we need to
enable policy CMP0091 if we want to make use of MSVC_RUNTIME_LIBRARY
and/or CMAKE_MSVC_RUNTIME_LIBRARY.  Fixes issue #912.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,12 @@
 # 3.5 is actually available almost everywhere, but this a good minimum
 cmake_minimum_required(VERSION 3.4)
+
+# enable MSVC_RUNTIME_LIBRARY target property
+# see https://cmake.org/cmake/help/latest/policy/CMP0091.html
+if(POLICY CMP0091)
+  cmake_policy(SET CMP0091 NEW)
+endif()
+
 project(YAML_CPP VERSION 0.6.3 LANGUAGES CXX)
 
 include(CMakePackageConfigHelpers)


tomlogic:
This resolved my problem, and I was able to verify it by grepping the generated vcxproj files in my CMake build directory.
```
$ grep RuntimeLibrary *.vcxproj
yaml-cpp.vcxproj:      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
yaml-cpp.vcxproj:      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
yaml-cpp.vcxproj:      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
yaml-cpp.vcxproj:      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
```
The linked documentation says that we need to explicitly set the NEW behavior on policy CMP0091, otherwise it's apparently ignored.  Maybe @slurps-mad-rips or @JPGygax68 could weigh in, since they made related changes to `CMakeLists.txt`.

Were you able to generate non-DLL runtime library builds with the current configuration?  Is this the proper fix?

Note that I first tried setting this after the `project()` definition, but the policy needs to be set beforehand.

slurps-mad-rips:
An alternative would be to `set(CMAKE_POLICY_DEFAULT_CMP0091 NEW)` in any parent projects if you're using this via `add_subdirectory`, or place it before the call to `project()`. Then the `if()` isn't needed in that case (this tends to be faster with how CMake sets things up)

tomlogic:
@slurps-mad-rips, per the CMake website documentation for CMAKE_POLICY_DEFAULT_CMP:
> This variable should not be set by a project in CMake code; use cmake_policy(SET) instead. Users running CMake may set this variable in the cache (e.g. -DCMAKE_POLICY_DEFAULT_CMP<NNNN>=<OLD|NEW>) to set a policy not otherwise set by the project. Set to OLD to quiet a policy warning while using old behavior or to NEW to try building the project with new behavior.

I would think it's OK for us to be setting it like this, because we're depending on it for the MSVC_RUNTIME_LIBRARY feature.  Maybe we could accomplish the change by expanding the definition of `backport-msvc-runtime` to be defined if the policy is set to OLD?

slurps-mad-rips:
> Maybe we could accomplish the change by expanding the definition of backport-msvc-runtime to be defined if the policy is set to OLD?

Or if it's undefined. That said, the (IMHO real) issue is that CMake almost never actually sets these values because the `cmake_policy(VERSIONS)` form was broken for some time. I personally tend to actually set the `DEFAULT_CMP` form because the policy scoping is so broken and misunderstood it's easier to just "set it and forget it". Also if someone wants "OLD" behavior over "NEW" behavior it means that their CMake file probably did the wrong thing in the first place 😅 

Even if you set a policy to `OLD` CMake will still alert for deprecations or removals, and the whole point of setting `NEW` sometimes is so that CMake doesn't complain (even when it's technically not supposed to)

> (And I'm not referencing the project from a parent project. I need the build artifacts for another project that doesn't use CMake.)

Totally understand.

tomlogic:
I feel like my proposed patch is simple, corrects a failure (can't generate /MT and /MTd builds), and follows recommendations for CMake.

Are there any arguments against merging this?  Maybe @JPGygax68 has an opinon?

jbeder:
For my part (having little insight into CMake or MSVC), if @slurps-mad-rips approves this change, I'll merge it.

slurps-mad-rips:
I'm fine with it. I just have a personal aversion to CMake's `if()` command 🤣
Why branch when you can not? 😛

JPGygax68:
All I did in my PR was fix a mistake: the variable `CMAKE_MSVC_RUNTIME_LIBRARY` was already being set, but its value was dependent on the wrong setting ("build shared" instead of "use MSVC shared libraries").
I wasn't aware that the correct functioning of that variable was dependent on a policy being enabled; but of course I have no objection to doing so.

tomlogic:
> I'm fine with it. I just have a personal aversion to CMake's `if()` command 🤣
> Why branch when you can not? 😛

I think I understand the need for the `if()`.  If I try to set a policy that doesn't exist, that should be an error condition.  IIRC, builds with older versions of CMake will work because of the fallback code to explicitly pass in `-MT` (for example).

It's unfortunate that this particular policy doesn't default to `NEW` as many other policies do.  Years from now if we set a new minimum version, we might be able to remove this policy code.

slurps-mad-rips:
>  If I try to set a policy that doesn't exist, that should be an error condition. IIRC, builds with older versions of CMake will work because of the fallback code to explicitly pass in -MT (for example).

No `if()` is needed if you set `CMAKE_POLICY_DEFAULT_CMP` to `NEW`, which is what I recommend folks do, CMake documentation be danged. This is, after all, the same tool that doesn't have `--target test` build the tests before trying to run them :P

jbeder:
SG, I'll merge this. If anyone has any better suggestions in the future, happy to merge those too :)

jbeder:
Ha, @slurps-mad-rips, I merged this at the same time you just commented. It sounds like @tomlogic suggested they can't do your suggestion since they don't have a parent project? In any case, happy to merge future PRs, as always.

slurps-mad-rips:
It can be placed above the `cmake_minimum_required` call, but before `project()`. Because of how scoping works in CMake, even parent project's won't be affected by setting it that way (due to policy scoping). Doesn't matter, though, I was simply explaining why I prefer setting a variable over an if-statement :)

NickolayGerasimenko:
@jbeder @tomlogic What is the up-to date and MOST correct way to use CMAKE_MSVC_RUNTIME_LIBRARY?
I've spent 3 hours googling all stack overflow and cannot force to /MT/ MD in MSVC 2019 project.

bruxisma:
Ahh. After investigating briefly, unfortunately [the documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html) states:

> Note that with [Visual Studio Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators) the native build system may choose to add its own default runtime library selection flag.

This probably won't occur if you use make or ninja. You can at least place a `if (CMAKE_GENERATOR MATCHES "^Visual Studio")` for your existing attempt.

(That said, if you have the ability to use a cmake presets, I'd recommend using Visual Studio's capabilities to use them directly alongside the Ninja Multi-Config generator.)

PR #916, opened 2020-07-07
fix issue501: handle the long key properly.

### LINKED ISSUE #501: Long key not handled properly
This should work, but does not:

int main(void)
{
    YAML::Node nn;
    nn[std::string(2000, 'a')] = 0;
    YAML::Load(YAML::Dump(nn));
    return 0;
}

--> Terminate called after throwing an instance of 'YAML::ParserException'
  what():  yaml-cpp: error at line 1, column 2001: illegal map value

I believe the problem is that the long key should be emitted as such:

? LONGKEY
: value

dota17:
fix #501 
write the long key (> 1024) properly.

### DIFF
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -266,8 +266,8 @@ TEST_F(EmitterTest, SimpleLongKey) {
 }
 
 TEST_F(EmitterTest, SingleLongKey) {
-  const std::string& shortKey = std::string(1024, 'a');
-  const std::string& longKey = std::string(1025, 'a');
+  const std::string shortKey(1024, 'a');
+  const std::string longKey(1025, 'a');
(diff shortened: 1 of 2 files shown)


test/integration/emitter_test.cpp
```diff
@@ -266,16 +266,23 @@ TEST_F(EmitterTest, SimpleLongKey) {
 }
 
 TEST_F(EmitterTest, SingleLongKey) {
+  const std::string& shortKey = std::string(1024, 'a');
```
jbeder, line 269: 1) it’s more idiomatic to have this be a value, not a reference
2) you can construct it directly (rather than use =)

dota17: ok, I have modified it.

PR #918, opened 2020-07-14
Fix compilation for Qt-projects

hkarel:
Not use 'emit' as variable name, in Qt Framework it reserved word


hkarel:
Renamed 'emit' to 'v'. The commentary is removed.

PR #919, opened 2020-07-14
fix issue917: handle the trailing TAB of block scalar

### LINKED ISSUE #917: Not stripping whitespace after block scalars in map
Adding this test to specexamples:
`const char* exRyan = "key: value\t \t#Comment\n";`

Found the value comparison fails:
`EXPECT_EQ("value", doc["key"].as<std::string>());`

Where the return of `as<string>` is "value\t"

This should have stripped the white space after value, but it does not.

dota17:
fix #917

+ For this issue, I think the `TAB` belongs to the separation spaces in spec v1.2 and the trailing `TAB` of scalar should be trimed.
+ the pyyaml do not allow the trailing TAB of block scalar and will throw the "found character '\t' that cannot start any token".   I think this is because pyyaml does not support the  spec v1.2 perfectly.

### DIFF
--- a/src/scanscalar.cpp
+++ b/src/scanscalar.cpp
@@ -204,7 +204,7 @@ std::string ScanScalar(Stream& INPUT, ScanScalarParams& params) {
 
   // post-processing
   if (params.trimTrailingSpaces) {
-    std::size_t pos = scalar.find_last_not_of(' ');
+    std::size_t pos = scalar.find_last_not_of(" \t");
     if (lastEscapedChar != std::string::npos) {
       if (pos < lastEscapedChar || pos == std::string::npos) {
         pos = lastEscapedChar;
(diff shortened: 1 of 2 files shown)


jbeder:
Yes, I think this is right.

7.16 talks about [flow mappings](https://yaml.org/spec/1.2/spec.html#ns-flow-map-implicit-entry(n,c)), where it defines:

```
ns-flow-map-yaml-key-entry(n,c) | ::= | ns-flow-yaml-node(n,c)( ( s-separate(n,c)?    c-ns-flow-map-separate-value(n,c) )\| e-node )
```

and if you trace `s-separate(n,c)` it is:

```
s-separate-in-line	::=	s-white+
```

which is defined as space or tab.

jbeder:
My bad, I was looking at flow definitions. It's still true, though, for block, in the same section (7.4.2):

```
ns-s-implicit-yaml-key(c)	::=	ns-flow-yaml-node(n/a,c) s-separate-in-line?
/* At most 1024 characters altogether */
```

And the `s-separate-in-line` is spaces or tabs.

Issue #590, opened 2018-05-25
parse string value null fail


jingshaoqi:
I have test.yaml contents follow:
port: 1234
server: null

but parse server fail:
std::string str = node["server"].as<<std::string>>();

yaml-cpp: error at line 0, column 0: bad conversion
if change null to other, it's ok

kcollinssibley:
I have also run into this issue on yaml-cpp-0.6.2.
This snippet of a yaml log file was generated using yaml-cpp-0.5.3:
```yaml
...
inventory:
  type: NULL
  value: ~
...
```
When upgrading my system to use yaml-cpp-0.6.2, I get the same error when trying to parse the log file created using yaml-cpp-0.5.3:
```
terminate called after throwing an instance of 'YAML::TypedBadConversion<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >'
  what():  yaml-cpp: error at line 2, column 9: bad conversion
Aborted (core dumped)
```

If I regenerate the file using yaml-cpp-0.6.2, the null value is handled differently:
```yaml
...
inventory:
  type: "NULL"
  value: ~
...
```

It is not feasible for us to regenerate all of our yaml files previously generated using yaml-cpp-0.5.3. This bug is preventing us from updating yaml-cpp and using new features elsewhere in out system.

jbeder:
Reasonable. If you want to submit a PR that allows parsing null as a string, happy to review it.

HensonLi:
@jbeder  Is there any PR or ticket relevant to how to parse a node with a null key
just like : 
a:
   b:1
     :2
   c:3

jbeder:
@HensonLi a null key is like any other key. Please ask usage questions like this on Stack Overflow.

HensonLi:
Got it， thanks for your reply, will post the issue onto Stack Overflow
later!


[image: Mailtrack]
<https://mailtrack.io?utm_source=gmail&utm_medium=signature&utm_campaign=signaturevirality5&>
Sender
notified by
Mailtrack
<https://mailtrack.io?utm_source=gmail&utm_medium=signature&utm_campaign=signaturevirality5&>
19/03/01
上午9:12:14

Jesse Beder <notifications@github.com> 于2019年3月1日周五 上午12:36写道：

dota17:
@jingshaoqi 
[...]

PR #920, opened 2020-07-16
fix issue590: add the support to null type for as<std::string>

dota17:
#590 

* Add the support to null type for `as<std::string>`.
* `as<std::string>` outputs the `"null"` for null type node. This can be changed as `"~"`, `"Null"`, etc.

### DIFF
--- a/test/integration/load_node_test.cpp
+++ b/test/integration/load_node_test.cpp
@@ -314,6 +314,8 @@ TEST(NodeTest, IncorrectFlow) {
 TEST(NodeTest, LoadTildeAsNull) {
   Node node = Load("~");
   ASSERT_TRUE(node.IsNull());
+  EXPECT_EQ(node.as<std::string>(), "null");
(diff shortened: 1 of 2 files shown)


test/integration/load_node_test.cpp
```diff
@@ -314,6 +314,8 @@ TEST(NodeTest, IncorrectFlow) {
 TEST(NodeTest, LoadTildeAsNull) {
   Node node = Load("~");
   ASSERT_TRUE(node.IsNull());
+  EXPECT_EQ(node.as<std::string>(), "null");
```
jbeder, line 317: Are you sure this is the behavior we want? If `node` is null, shouldn't `node.as<std::string>()` be an error?

jbeder: Never mind, you know what, I see I commented on that bug that this is reasonable. I'll approve.

Issue #743, opened 2019-09-24
Flow + Newline + EmptyMap bug


MidoriYakumo:
Add following test case:

```c++
TEST_F(EmitterTest, FlowNewlineEmpty)
{
        out << YAML::BeginMap;
        out << YAML::Key << "foo";
        out << YAML::Value << YAML::Flow << YAML::BeginMap << YAML::Newline << YAML::EndMap;
        out << YAML::EndMap;

        ExpectEmit("foo: {\n  }\n");
}
```

but result is "foo: {\n  {}\n"

MidoriYakumo:
Can be fixed with diff:

```diff
diff --git a/src/emitter.cpp b/src/emitter.cpp
index 016beb1..91ee5b7 100644
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -198,6 +198,7 @@ void Emitter::EmitEndSeq() {
   if (!good())
     return;
 
+  bool is_native_flow = m_pState->CurGroupFlowType() == FlowType::Flow;
   if (m_pState->CurGroupChildCount() == 0)
     m_pState->ForceFlow();
 
@@ -205,7 +206,7 @@ void Emitter::EmitEndSeq() {
     if (m_stream.comment())
       m_stream << "\n";
     m_stream << IndentTo(m_pState->CurIndent());
-    if (m_pState->CurGroupChildCount() == 0)
+    if (m_pState->CurGroupChildCount() == 0 && !is_native_flow)
       m_stream << "[";
     m_stream << "]";
   }
@@ -228,6 +229,7 @@ void Emitter::EmitEndMap() {
   if (!good())
     return;
 
+  bool is_native_flow = m_pState->CurGroupFlowType() == FlowType::Flow;
   if (m_pState->CurGroupChildCount() == 0)
     m_pState->ForceFlow();
 
@@ -235,7 +237,7 @@ void Emitter::EmitEndMap() {
     if (m_stream.comment())
       m_stream << "\n";
     m_stream << IndentTo(m_pState->CurIndent());
-    if (m_pState->CurGroupChildCount() == 0)
+    if (m_pState->CurGroupChildCount() == 0 && !is_native_flow)
       m_stream << "{";
     m_stream << "}";
   }
diff --git a/test/integration/emitter_test.cpp b/test/integration/emitter_test.cpp
index 3a5783b..e70753f 100644
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
[...]

PR #921, opened 2020-07-16
fix issue743: handle the empty content of flow sep/map correctly during emitting.

dota17:
#743  
* fix bug: emit the correct empty seq/map with` newline`/`comment`/`tag`/`anchor`.
* The detailed output contents are shown in test case. Are they the right ones？ @jbeder  ha-ha, I'm not very sure for these.

### DIFF
--- a/src/emitterstate.cpp
+++ b/src/emitterstate.cpp
@@ -160,11 +160,12 @@ void EmitterState::EndedGroup(GroupType::value type) {
     return SetError(ErrorMsg::UNEXPECTED_END_MAP);
   }
 
-  // if have the unhandled Tag/Anchor, set the error message
-  if (m_hasTag)
+  if (m_hasTag) {
     SetError(ErrorMsg::INVALID_TAG);
-  if (m_hasAnchor)
+  }
+  if (m_hasAnchor) {
     SetError(ErrorMsg::INVALID_ANCHOR);
+  }
 
   // get rid of the current group
   {
--- a/src/emitterstate.cpp
+++ b/src/emitterstate.cpp
@@ -160,6 +160,12 @@ void EmitterState::EndedGroup(GroupType::value type) {
     return SetError(ErrorMsg::UNEXPECTED_END_MAP);
   }
 
+  // if have the unhandled Tag/Anchor, set the error message
+  if (m_hasTag)
+    SetError(ErrorMsg::INVALID_TAG);
+  if (m_hasAnchor)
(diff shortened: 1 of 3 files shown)


test/integration/emitter_test.cpp
```diff
@@ -138,6 +138,70 @@ TEST_F(EmitterTest, EmptyFlowSeq) {
   ExpectEmit("[]");
 }
 
+TEST_F(EmitterTest, EmptyBlockSeqWithBegunContent) {
+  out << BeginSeq;
+  out << BeginSeq << Comment("comment") << EndSeq;
+  out << BeginSeq << Newline << EndSeq;
+  out << BeginSeq << Anchor("test") << EndSeq;
```
jbeder, line 145: I think the behavior here should either be:

1. An anchor to a null entry in this sequence, e.g. `[&test ~]`
2. An error.

Similarly for the tag one. What do you think?

dota17: I prefer the second one. In essence，these two (anchor and tag) are the wrong formats.  yaml-cpp should not emit these.

dota17: @jbeder I have updated the code. When the empty anchor/tag is emitted, yaml-cpp will set the error message `INVALID_ANCHOR`/`INVALID_TAG`.

src/emitterstate.cpp
```diff
@@ -160,6 +160,12 @@ void EmitterState::EndedGroup(GroupType::value type) {
     return SetError(ErrorMsg::UNEXPECTED_END_MAP);
   }
 
+  // if have the unhandled Tag/Anchor, set the error message
```
jbeder, line 163: This comment isn't necessary.

src/emitterstate.cpp
```diff
@@ -160,6 +160,12 @@ void EmitterState::EndedGroup(GroupType::value type) {
...
+  if (m_hasTag)
```
jbeder, line 164: Put {} around the body of this if statement, even though it's only one line.

(I've been meaning to go through and enforce that throughout the codebase, but haven't had the chance.)

dota17: done. Maybe I can change the clang-format Configuration File && Optimization Option, so we can clang-format all files quickly.

src/emitterstate.cpp
dota17: ok, done.

PR #922, opened 2020-07-16
Allowing partial specialization of the convert struct

BenjaminNavarro:
This is done by adding a second, defaulted, template parameter that can
be used in conjunction of std::enable_if

An example of usage is added to the tutorial

The code and example has been taken from #310

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -199,3 +199,79 @@ YAML::Node node = YAML::Load("start: [1, 3, 0]");
 Vec3 v = node["start"].as<Vec3>();
 node["end"] = Vec3(2, -1, 0);
 ```
+
+## Partial specialization
+
+If you need to specialize the `convert` struct for a set of types instead of just one you can use partial specialization with the help of `std::enable_if` (SFINAE).
+
+Here is a small example showing how to partially specialize the `convert` struct for all types deriving from a base class:
+
+```cpp
+// Base class
+class A
+{
+public:
+  A() = default;
+  A(int a) : a{a} {}
+
+  // virtual load/emit methods
+  virtual void load(const YAML::Node &node) {
+    a = node["a"].as<int>();
+  }
+
+  virtual YAML::Node emit() const  {
+    YAML::Node node;
+    node["a"] = a;
+    return node;
+  }
+  
+  int a;
+};
+
+// Derived class
+class B : public A
+{
+public:
+  B() = default;
+  B(int a, int b) : A{a}, b{b} {}
+
+  // override virtual load/emit methods
+  virtual void load(const YAML::Node &node) override {
+    A::load(node);
+    b = node["b"].as<int>();
+  }
+
+  virtual YAML::Node emit() const override {
+    YAML::Node node = A::emit();
+    node["b"] = b;
+    return node;
+  }
+
+  int b;
+};
+
+// Implementation of convert::{encode,decode} for all classes derived from or being A
+namespace YAML {
+  template<typename T> 
+  struct convert<T, typename std::enable_if<std::is_base_of<A, T>::value>::type> {
+    static Node encode(const T &rhs) {
+      Node node = rhs.emit();
+      return node;
+    }
+
+    static bool decode(const Node &node, T &rhs) {
+      rhs.load(node);
+      return true;
+    }
+  };
+}
+```
+
+Which can then be use like this:
+```cpp
+YAML::Node node = YAML::Load("{a: 1, b: 2}");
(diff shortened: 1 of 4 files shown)


BenjaminNavarro:
How should I approach the testing of this? It's easy to check that it works for the correct subset of types but not that it doesn't for the others since it would create a compilation error

BenjaminNavarro:
I added a small test to make sure that the feature works and to potentially avoid future regressions.

Let me know if you require more testing

BenjaminNavarro:
@jbeder Could you review this PR when you have time?

docs/Tutorial.md
```diff
@@ -198,4 +198,74 @@ Then you could use `Vec3` wherever you could use any other type:
...
+
+## Partial specialization
+
+If you need to specialize the `convert` struct for a set of types instead of just one you can use partial specialization with the help of `std::enable_if` (SFINAE).
+
+Here is a small example showing how to partially specialize the `convert` struct for all types deriving from a base class:
+
+```cpp
+// Base class
+class A
```
dota17, line 211: Inconsistent indentation. Match the format of new content with the  above.

test/integration/convert_partial_specialization_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  int a;
+};
+
+// Derived class
+class B : public A {
+ public:
+  // override virtual load/emit methods
+  virtual void load(const YAML::Node &node) override {
```
dota17, line 25: Even though it's a test case, I think the constructors are needed, like #310. It is also convenient for later usage.

test/integration/convert_partial_specialization_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  static bool decode(const Node &node, T &rhs) {
+    rhs.load(node);
+    return true;
+  }
+};
+
+namespace {
+
+TEST(ConvertPartialSpecializationTest, encode_base) {
```
dota17, line 57: `encode_base -> EncodeBaseClass` just like the other test cases' name.

docs/Tutorial.md
BenjaminNavarro: Fixed. Didn't realize it was tabs instead of spaces

test/integration/convert_partial_specialization_test.cpp
BenjaminNavarro: Added

test/integration/convert_partial_specialization_test.cpp
BenjaminNavarro: Done

test/integration/convert_partial_specialization_test.cpp
dota17: How about updating the Tutorial.md at the same time. And use this in testcase.
```C++  
  A a(1);
  B b(1, 2);
```

BenjaminNavarro: I updated the test cases to use the constructors and added them in the tutorial.
However the tutorial doesn't make use of them

pettni:
I am also interested in getting this merged, was actually about to submit the same merge request myself. This change would facilitate using ```yaml-cpp``` in conjunction with reflection tools such as ```boost::hana``` and ```boost::fusion```, and I don't see any drawbacks of adding the second default template argument.

yuriy1337:
Is there a plan for getting this merged. I am also interested in this functionality.

SGSSGene (changes requested):
@BenjaminNavarro I think this is a great addition. If you still need/want this I would be happy to accept this PR after a rebase/merge with the current master and the missing newline.

test/integration/convert_partial_specialization_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+TEST(ConvertPartialSpecializationTest, DecodeDerivedClass) {
+  B b(1, 2);
+  Node n;
+  n = b;
+  EXPECT_EQ(b.a, n["a"].as<int>());
+  EXPECT_EQ(b.b, n["b"].as<int>());
+}
+
+}  // namespace
+}  // namespace YAML
```
SGSSGene: There seems to be a new line missing.

BenjaminNavarro:
Rebase done and newline added.

Today is the sixth anniversary of this PR, looks like a good day to merge it !

SGSSGene (approved):
Thank you! This is a great feature!

I didn't manage it to make it exactly 6 years! But waiting another year seems unnecessary.

PR #924, opened 2020-07-18
Reduce std::string copies

k0zmo:
When indexing (using `Node::operator[]` and friends) with a `Key` which is defined as a template type we early convert it to `std::string` - even when the Key is already a `std::string`. This is not really needed as Node's implementation does work with a templated key type - both `const char*` and `std::string`.

### DIFF
--- a/include/yaml-cpp/node/detail/impl.h
+++ b/include/yaml-cpp/node/detail/impl.h
@@ -106,7 +106,11 @@ inline bool node::equals(const T& rhs, shared_memory_holder pMemory) {
 }
 
 inline bool node::equals(const char* rhs, shared_memory_holder pMemory) {
-  return equals<std::string>(rhs, pMemory);
+  std::string lhs;
+  if (convert<std::string>::decode(Node(*this, std::move(pMemory)), lhs)) {
+    return lhs == rhs;
+  }
+  return false;
 }
 
 // indexing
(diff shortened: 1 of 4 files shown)


include/yaml-cpp/node/detail/impl.h
```diff
@@ -106,7 +106,11 @@ inline bool node::equals(const T& rhs, shared_memory_holder pMemory) {
 }
 
 inline bool node::equals(const char* rhs, shared_memory_holder pMemory) {
-  return equals<std::string>(rhs, pMemory);
+  std::string lhs;
+  if (convert<std::string>::decode(Node(*this, std::move(pMemory)), lhs)) {
```
dota17, line 110: `std::move(pMemory)` can optimize the performance with move semantics, is my understading right? Why not add this in line102? So that we do not need to change the line 109.

k0zmo: I'd like to make `shared_memory_holder` optimizations (move semantics or simply passing it as a `const&`) as a whole in different PRs.
This change however, is not about adding `std::move`, but not creating `std::string` from `rhs` just to compare it to `lhs`. We can do it without creating temporary string.

include/yaml-cpp/node/impl.h
```diff
@@ -315,51 +315,6 @@ inline void Node::push_back(const Node& rhs) {
...
-template <std::size_t N>
-struct to_value_t<char[N]> {
-  explicit to_value_t(const char* t_) : t(t_) {}
-  const char* t;
-  using return_type = std::string;
-
-  const std::string operator()() const { return t; }
-};
-
-// converts C-strings to std::strings so they can be copied
```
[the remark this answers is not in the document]
jbeder: I'm actually not sure myself. I may have tricked myself into thinking the values were being copied somewhere.

PR #958, opened 2020-11-11
Make dll.h compatible with -fvisibility=hidden

orgads:
Originally from Qt Creator.


vmiheer:
Ideally it would be great to replace dll.h with https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html.

PR #975, opened 2021-01-28
Always quote strings with colons

Saviq:
Otherwise they may get misinterpreted, or cause a syntax error even.


Saviq:
For reference, a digit-and-colon-only string may get interpreted as a 60-base integer:

https://yaml.org/type/int.html

SGSSGene:
I believe this PR is outdated and not accurate with yaml-spec 1.2.0.

Using 60-base integer has ben drop moving from yaml-spec 1.1 to 1.2.0.

>  The binary and sexagesimal integer formats have been dropped.

see https://yaml.org/spec/1.2.2/ext/changes/

Saviq:
Sure.

PR #984, opened 2021-04-18
Hide most of non-public symbols by default

pinotree:
Make use of CMake features to generate `dll.h` with export macros, and to hide all the symbols not explicitly marked for exporting.

This PR provides a more complete implementation of #644 and #958.


pinotree:
Hi @jbeder!

Would it be possible to take a look at this? Thanks in advance!

jbeder (approved):

src/node_data.cpp
```diff
@@ -13,7 +13,7 @@
 
 namespace YAML {
 namespace detail {
-std::atomic<size_t> node::m_amount{0};
+YAML_CPP_API std::atomic<size_t> node::m_amount{0};
```
egorpugin, line 16: Isn't same YAML_CPP_API should be in header too?

pinotree: No, in the definition is enough, for this internal symbol.

jbeder:
FYI this broke the build (see #1036). I reverted this PR; once you've determined the cause and fix, please send an updated PR and I'll approve. Ideally also please add a test that confirms the fix.

pinotree:
> FYI this broke the build (see #1036). I reverted this PR; once you've determined the cause and fix, please send an updated PR and I'll approve. Ideally also please add a test that confirms the fix.

While I understand the concern, it'd have been nice to at least give me a change to look at it before the so quick revert...

jbeder:
Ah, I understand there are different philosophies here, and so I apologize for not communicating the rationale. I'm of the school of thought "rollback first, as questions later" to get the build to a stable state as quickly as possible. Unless absolutely necessary, I prefer rollbacks to fix-forwards. You can Google "rollback vs. fix forward" to see a litany of arguments for or against :)

pinotree:
Sure, I know and understand that. However, it still does not feel nice that the PR took 5 months to even get any action (which got after my explicit mention), and it got ~1 day with no question asked to be reverted. Even worse when the bug report mentioning a problem is _so generic_ that, at least to me, is barely actionable at all.

jbeder:
Yeah, I agree that bug report wasn't great. I'm sorry about the 5 month
delay, this is a side-side-side project for me so my time is bursty. When
you pinged the PR, it looked simple, and then when I saw the bug report, I
worried that, geez, I've ignored the project for months and now I approve a
PR that broke it, so I rolled back in an abundance of caution.

PhilipDeegan:
before this PR, anyone could take this code and use basically any tooling and build a usable library, but now you have to use cmake or dll.h won't exist.

I would really not like to have to use cmake.

pinotree:
> before this PR, anyone could take this code and use basically any tooling and build a usable library, but now you have to use cmake or dll.h won't exist.

See my answer: https://github.com/jbeder/yaml-cpp/issues/1046#issuecomment-944969641

Just because "it worked" does not mean it was doable "for free", you had (and still have) to do a lot of logic on your own, and this is one bit of logic more done by the build system instead of rolling an own (and limited/buggy) solution.

PR #991, opened 2021-04-23
Make enum's into enum classes

TedLyngmo:
This makes all `enum`'s into `enum class`es except for:

* The internal enum's in `src/stream.cpp`:
    ```
    UtfIntroState
    UtfIntroCharType
    ```
* `EMITTER_MANIP` (covered separately in #989)
* Pseudo enum's (covered separately in #990)

Signed-off-by: Ted Lyngmo <ted@lyncon.se>

PR #990, opened 2021-04-23
Make pseudo enum classes into proper enum classes

TedLyngmo:
This changes some constructs like this
```
    struct FmtScope {
      enum value { Local, Global };
    };
    struct GroupType {
      enum value { NoType, Seq, Map };
    };
    struct FlowType {
      enum value { NoType, Flow, Block };
    };
```
into this
```
    enum class FmtScope { Local, Global };
    enum class GroupType { NoType, Seq, Map };
    enum class FlowType { NoType, Flow, Block };
```
The full list of pseudo enum classes now turned into real enum classes:
```
enum class State { WaitingForSequenceEntry, WaitingForKey, WaitingForValue };
enum class EmitterNodeType { NoType, Property, Scalar, FlowSeq, BlockSeq, FlowMap, BlockMap };
enum class Type { Verbatim, PrimaryHandle, NamedHandle };
enum class EmitterStyle { Default, Block, Flow };
enum class iterator_type { NoneType, Sequence, Map };
enum class NodeType { Undefined, Null, Scalar, Sequence, Map };
enum class CollectionType { NoCollection, BlockMap, BlockSeq, FlowMap, FlowSeq, CompactMap };
enum class FmtScope { Local, Global };
enum class GroupType { NoType, Seq, Map };
enum class FlowType { NoType, Flow, Block };
enum class StringFormat { Plain, SingleQuoted, DoubleQuoted, Literal };
enum class StringEscaping { None, NonAscii, JSON };
```
Signed-off-by: Ted Lyngmo <ted@lyncon.se>

PR #989, opened 2021-04-23
Make EMITTER_MANIP an enum class

TedLyngmo:
This fixed `-Wshadow` warnings emitted when compiling with clang++

Signed-off-by: Ted Lyngmo <ted@lyncon.se>

PR #992, opened 2021-04-23
Enum merge #989 #990 #991

TedLyngmo:
This simply combines the three pull requests #989, #990 and #991 to simplify for anyone who'd like to see if anything in those PR:s causes problems with their code.


jbeder:
I appreciate the PR here, and I support the general idea of enum classes, but because this is a breaking change where the main goal is a style improvement, I'm reluctant to merge it. I'm going to close these.

PR #1005, opened 2021-07-03
Prevent trailing spaces when emitting literal strings

theamarin:
This behavior matches the format expectations of the Red Hat yaml
language server.

### DIFF
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -382,6 +382,20 @@ TEST_F(EmitterTest, ScalarFormat) {
       "crazy\tsymbols that we like");
 }
 
+TEST_F(EmitterTest, LiteralWithoutTrailingSpaces) {
+  out << YAML::BeginMap;
+  out << YAML::Key << "key";
+  out << YAML::Literal;
+  out << "expect that with two newlines\n\n"
+         "no spaces are emitted in the empty line";
+  out << YAML::EndMap;
+
+  ExpectEmit(
+      "key: |\n"
+      "  expect that with two newlines\n\n"
+      "  no spaces are emitted in the empty line");
+}
+
 TEST_F(EmitterTest, AutoLongKeyScalar) {
   out << BeginMap;
   out << Key << Literal << "multi-line\nscalar";
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -385,7 +385,7 @@ TEST_F(EmitterTest, ScalarFormat) {
 TEST_F(EmitterTest, LiteralWithoutTrailingSpaces) {
   out << YAML::BeginMap;
   out << YAML::Key << "key";
-  out << YAML::Literal;
+  out << YAML::Value << YAML::Literal;
   out << "expect that with two newlines\n\n"
          "no spaces are emitted in the empty line";
   out << YAML::EndMap;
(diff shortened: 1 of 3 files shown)


jbeder (changes requested):
Can you write a test or two so I can see the intended changes?

theamarin:
Thank you very much for your very quick reply.
I have added a NodeEmitterTest, called LiteralWithoutTrailingSpaces. It diverges a bit from the style of the previous tests, because I did not know how to use `YAML::Literal` otherwise.
I hope this is okay. Otherwise, I am happy to get a hint how to proceed.
Thank you.

test/node/node_test.cpp
```diff
@@ -769,5 +769,17 @@ TEST_F(NodeEmitterTest, NestFlowMapListNode) {
...
+TEST_F(NodeEmitterTest, LiteralWithoutTrailingSpaces) {
+  YAML::Emitter emitter;
+  emitter << YAML::BeginMap;
+  emitter << YAML::Key << "key";
+  emitter << YAML::Literal;
+  emitter << "value\n\nwith newlines";
+  emitter << YAML::EndMap;
+
+  ASSERT_TRUE(emitter.good());
+  EXPECT_STREQ("key: |\n  value\n\n  with newlines", emitter.c_str());
```
jbeder, line 782: Please use a multi line string literal here, I think it's easier to read.

test/node/node_test.cpp
```diff
@@ -769,5 +769,17 @@ TEST_F(NodeEmitterTest, NestFlowMapListNode) {
 
   ExpectOutput("{position: [1.5, 2.25, 3.125]}", mapNode);
 }
+
+TEST_F(NodeEmitterTest, LiteralWithoutTrailingSpaces) {
```
jbeder, line 773: This test should go in integration-tests/emittertest.cpp

jbeder (changes requested):
Thanks! Some bookkeeping comments.

test/integration/emitter_test.cpp
```diff
@@ -382,6 +382,20 @@ TEST_F(EmitterTest, ScalarFormat) {
...
+TEST_F(EmitterTest, LiteralWithoutTrailingSpaces) {
+  out << YAML::BeginMap;
+  out << YAML::Key << "key";
+  out << YAML::Literal;
+  out << "expect that with two newlines\n\n"
+         "no spaces are emitted in the empty line";
+  out << YAML::EndMap;
+
+  ExpectEmit(
+      "key: |\n"
```
jbeder, line 394: I meant a multi line literal like this:

https://programming-idioms.org/idiom/48/multi-line-string-literal/2680/cpp

test/integration/emitter_test.cpp
```diff
@@ -382,6 +382,20 @@ TEST_F(EmitterTest, ScalarFormat) {
       "crazy\tsymbols that we like");
 }
 
+TEST_F(EmitterTest, LiteralWithoutTrailingSpaces) {
+  out << YAML::BeginMap;
+  out << YAML::Key << "key";
+  out << YAML::Literal;
```
jbeder, line 388: This should have 'YAML::Value' first. I don't think it makes an output difference, though.

jbeder (changes requested):

test/integration/emitter_test.cpp
theamarin: > This should have 'YAML::Value' first. I don't think it makes an output difference, though.
:see_no_evil: sure, thanks.

test/integration/emitter_test.cpp
theamarin: Isn't it better to emphasize (or easier to see) that no spaces are in between the newline characters in a regular string?

jbeder: Fair enough.

jbeder (approved):

PR #1007, opened 2021-07-12
CMake: allow to install yaml-cpp from another CMake project

SpaceIm:
In public repository of conan (package manager), CMakeLists of libraries are wrapped (with `add_subdirectory()`)in a top CMakeLists injecting specific stuff.
With current CMakeLists of yaml-cpp, installation is forced to OFF when yaml-cpp is not the top project, which is quite unusual.

This PR allows to define the install target. It is ON by default if yaml-cpp is the top project, and OFF if not (but downstreams can still enable installation).

There was already a PR fixing this issue https://github.com/jbeder/yaml-cpp/pull/847, but there is no activity.


prince-chrismc (approved):

lkersting:
You now have 3 PRs for this one change: This one, #1011, and #847. This and #1011 are basically the same. It would be great if one of these could be merged.  I have to manually make this change in order to use  yaml-cpp with my projects.

jbeder:
Sure, I think I understand this now. If this PR's conflict is resolved, I'll merge it.

jbeder (approved):

PR #1010, opened 2021-07-15
Implement a customization point for types not default constructible

### LINKED ISSUE #993: custom converter for non default-constructable data-types
It seems in the current implementation it is not possible to convert YAML-nodes directly into non default-constructable data-types. 

The actual breaking point seems to be in 'impl.h:130', where an bare mantle object of the desired type is instantiated on the stack `T t;`, where T is the template-arg for the to-be-read custom-class. I guess the next instructions are invocing the customized `struct bool YAML::convert<T>::decode(Node&, T&)`, which are to fill the fields of the object with sensible data, by modifying the reference as the second argument.

Wouldn't it be more sensible to define the decoding function as ```T YAML::convert<T>::decode(Node&) throw()``` ; thus instead of using the bool as function-return as indicator for successful conversion, rather wrap the decode in an try-except block and so have the possibility to allocate the object on the stack inside the function-body itself?

ozars:
This patch adds a customization point `YAML::decode_dispatcher<T>::dispatch` which:

- allows decoding types that aren't default constructible,
- delegates to `YAML::convert<T>::decode` by default for backward compatibility,
- is enabled only when compiled with C++17 or newer since it depends on `std::optional`.

Notes:

- Compiling with C++17 automatically enables this feature.
- Relevant tests are added. Tested it locally (gcc 10.2 on debian buster) to confirm all tests passes with both C++11 & C++17.
- `YAML_CPP_HAS_DECODE_DISPATCHER` macro is exposed allowing checking this feature.
- Its signature has an `Enable` template parameter giving more flexibility, e.g. to allow SFINAE downstream:
```cpp
namespace YAML {
template <typename T, typename Enable = void>
struct decode_dispatcher {
  static std::optional<T> dispatch(const Node& node) {
   // ...
  }
};
}
```

Example:

  ```cpp
  #include <yaml-cpp/node/impl.h>

  struct Vec3 {
    double x, y, z;
    Vec3(double x, double y, double z) : x(x), y(y), z(z) {}
    // ...
  };

  namespace YAML {
  template<>
  struct decode_dispatcher<Vec3> {
    static std::optional<Vec3> dispatch(const Node& node) {
      if(!node.IsSequence() || node.size() != 3)
        return std::nullopt;
      return Vec3(
        node[0].as<double>(),
        node[1].as<double>(),
        node[2].as<double>()
      );
    }
  };
  }
  ```

Possibly closes: #993.

### DIFF
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -15,6 +15,11 @@
 #include <sstream>
 #include <string>
 
+#if __cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
+#include <optional>
+#define YAML_CPP_HAS_DECODE_DISPATCHER
+#endif
+
 namespace YAML {
 inline Node::Node()
     : m_isValid(true), m_invalidKey{}, m_pMemory(nullptr), m_pNode(nullptr) {}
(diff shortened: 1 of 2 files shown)


SGSSGene:
Closed in favor of #1488

Issue #1003, opened 2021-06-21
Add support for CMake + FetchContent


lambdaknight:
```FetchContent_Declare(
        yaml-cpp
        GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
        GIT_TAG yaml-cpp-0.6.3
)
set(YAML_CPP_BUILD_TESTS OFF)
FetchContent_MakeAvailable(yaml-cpp)
target_link_libraries(${TARGET} yaml-cpp)```

This works just fine. However, with the latest commit (a6bbe0e when submitted), this does not work. Please re-add the ability to use this with FetchContent.

**Comment by Akaame** [comment]
Hello,

```cmake
cmake_minimum_required(VERSION 3.14.0)
project(dummy-project)

set(TARGET main)

add_executable(${TARGET} src/main.cpp)
set_property(TARGET ${TARGET} PROPERTY CXX_STANDARD 20)

include(FetchContent)

FetchContent_Declare(
        yaml-cpp
        GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
        GIT_TAG a6bbe0e50ac4074f0b9b44188c28cf00caf1a723
)

set(YAML_CPP_BUILD_TESTS OFF)
FetchContent_MakeAvailable(yaml-cpp)
target_link_libraries(${TARGET} yaml-cpp)
```

and `GIT TAG yaml-cpp-0.7.0` builds successfully. Can you please open this issue up?

Regards.

SGSSGene:
This works with yaml-cpp-0.9.0.
if this issues reappears, please feel free to open a new ticket.

Issue #970, opened 2021-01-04
yaml-cpp/node/impl.h:206 error 'operator=' should return a reference to *this [-Werror=effc++]


tomeravi:
on Linux build

btbouwens:
Indeed, I also see this:

In file included from /workspace/software/lib/3rdparty/yaml-cpp/src/parse.cpp:7:
/workspace/software/lib/3rdparty/yaml-cpp/include/yaml-cpp/node/impl.h: In member function 'YAML::Node& YAML::Node::operator=(const T&)':
/workspace/software/lib/3rdparty/yaml-cpp/include/yaml-cpp/node/impl.h:206:10: warning: 'operator=' should return a reference to '*this' [-Weffc++]
  206 |   return *this;
      |          ^~~~~

tomeravi:
Do u have any suggestion how to solve it ?

T

בתאריך יום ה׳, 7 בינו׳ 2021, 12:39, מאת btbouwens ‏<notifications@github.com

btbouwens:
No, I don't understand the problem. It looks OK to me actually.
I've seen comments that the -Weffc++ is actually questionable, but the arguments are outside my competence ...

olafmandel:
I opened a [GCC bug #98841](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98841) about this.

olafmandel:
Jakub Jelinek wrote a fix for GCC for this, which "just" has to make it into a future release of GCC.

If I understand the explanation of the bug correctly, there is nothing that can be done to the yaml-cpp code to workaround the warning for older compilers, short of disabling the warning:

```diff
diff --git a/include/yaml-cpp/node/impl.h b/include/yaml-cpp/node/impl.h
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -209,7 +209,15 @@ inline Node& Node::operator=(const T& rhs) {
   if (!m_isValid)
     throw InvalidNode(m_invalidKey);
   Assign(rhs);
+// work around GCC PR c++/98841
+#if defined __GNUC__ && __GNUC__ < 11
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic ignored "-Weffc++"
+#endif
   return *this;
+#if defined __GNUC__ && __GNUC__ < 11
+#  pragma GCC diagnostic pop
+#endif
 }

 inline Node& Node::operator=(const Node& rhs) {
```

[...]

Issue #794, opened 2019-12-05
CMake Export Error When Including yaml-cpp as a Subdirectory in a Project


lkersting:
I have a project that includes yaml-cpp as a subdirectory in my project and uses it as a public dependency to one of my libraries. 
After the new CMake changes I get the following error:

`
CMake Error: install(EXPORT "${PROJECT_NAME}Targets" ...) includes target "${TARGET}" which requires target "yaml-cpp" that is not in the export set.`

It worked with the previous version before `yaml-cpp-targets` was exported by CMake. Now is line 129 of the top level CMakeLists.txt file you only export  `yaml-cpp-targets`  if `YAML_CPP_INSTALL=ON`. The problem is that `YAML_CPP_INSTALL` is turned off when yaml-cpp is built as a subdirectory.

I tried `SET(YAML_CPP_INSTALL ON CACHE INTERNAL "")` and it still did not turn `YAML_CPP_INSTALL` on. 

There should be a way to enable generation of the yaml-cpp install targets when it is included as a subdirectory in another project.

myrddin89:
I have the same problem.

nickelsey:
I am currently having the same issue. YAML_CPP_INSTALL should not be defined with a cmake_dependent_option, because it does not allow for cached values to be used from parent projects, as far as I can tell.

HarryDC:
This is holding us back from using the master, we are adding yaml-cpp via external-project-add, what is the correct solution here ?

jasonbeach:
I just hit this as well with the new version .0.7.0 due to making it a `cmake_dependent_option`

PR #1011, opened 2021-07-15
give YAML_CPP_INSTALL a default value that can be overridden

jasonbeach:
simple cmake fix that adds a default value of OFF when yaml-cpp is not the master project, but allows a user to override that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,8 +99,7 @@ endif()
 
 target_compile_options(yaml-cpp
   PRIVATE
-    $<${not-msvc}:-Wall -Wextra -Wshadow -Wno-long-long>
-    $<$<BOOL:${YAML_CPP_MASTER_PROJECT}>:-Weffc++>
+    $<${not-msvc}:-Wall -Wextra -Wshadow -Weffc++ -Wno-long-long>
     $<${not-msvc}:-pedantic -pedantic-errors>
 
     $<$<AND:${backport-msvc-runtime},${msvc-rt-mtd-static}>:-MTd>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,16 +17,22 @@ include(CTest)
 
 find_program(YAML_CPP_CLANG_FORMAT_EXE NAMES clang-format)
 
+if (NOT DEFINED YAML_CPP_MASTER_PROJECT)
+  set(YAML_CPP_MASTER_PROJECT OFF)
+  if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    set(YAML_CPP_MASTER_PROJECT ON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -93,7 +99,8 @@ endif()
 
 target_compile_options(yaml-cpp
   PRIVATE
-    $<${not-msvc}:-Wall -Wextra -Wshadow -Weffc++ -Wno-long-long>
+    $<${not-msvc}:-Wall -Wextra -Wshadow -Wno-long-long>
```
jbeder, line 102: This looks separate from the primary point of the PR?

CMakeLists.txt
```diff
@@ -17,16 +17,22 @@ include(CTest)
 
 find_program(YAML_CPP_CLANG_FORMAT_EXE NAMES clang-format)
 
+if (NOT DEFINED YAML_CPP_MASTER_PROJECT)
```
jbeder, line 20: Is this pattern common for CMake? It seems like you're trying to solve a pretty common problem.

jbeder (changes requested):
Can you clarify in the PR description what problem this is solving?

jasonbeach:
> Can you clarify in the PR description what problem this is solving?

By making creation of the install targets a cmake dependent option, there's no way to override it.  This makes using it with FetchContent not possible when the consuming library wants to be installed. Issue #1003 and #794 (duplicates of each other).  By making it an option the way I have, a consuming library can override it and force the creation of the install targets.

I'm very curious on why a number of very well known libraries such as this one, libfmt, spdlog, etc) follow the pattern of not automatically creating the install targets when they're not the top level project.  I guess I don't understand the use case or when that would be useful. Could you shed some light on that?

CMakeLists.txt
jasonbeach: it is, but adding the `YAML_CPP_MASTER_PROJECT` option from this PR makes this fix very easy.  It's a partial fix for issue #970 (basically if you're not the toplevel project omit the -Weffc++ option.)  If you'd rather once the other change is approve I can make a new PR for this.

jasonbeach: I guess I need to add the `not-msvc` to the check as well (the option would be erroneously added on msvc when then top-level project) ...may be just as easy to omit all these options if this library is not the top-level project.

CMakeLists.txt
jasonbeach: I'm very curious on why a number of very well known libraries such as this one, libfmt, spdlog, etc) follow the pattern of not automatically creating the install targets when they're not the top level project.  I guess I don't understand the use case or when that would be useful. Could you shed some light on that?

jasonbeach: The big reason the fix is needed is that the current implementation breaks the install ability in projects that include this via CMake's FetchContent. (CMake complains target xxx requires yaml-cpp which is not going to be installed (or something to that effect)

jbeder:
Obviated by #1007.

PR #1014, opened 2021-07-17
add Third Party Integrations

netcan:


jbeder:
I don't understand what netcan is? Is this an integration with some popular framework?￼￼

netcan:
> I don't understand what netcan is? Is this an integration with some popular framework?￼￼

that framework integration with yamlcpp, it use yamlcpp as backend.

PR #1021, opened 2021-08-16
Create target "format" by default only if yaml-cpp is main cmake project

OlivierLDff:
## Use Case

I included yaml-cpp with [`FetchContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html).
I don't want to have in my targets, a `format` target that do not format my main project.

## Fix

I introduce cmake option `YAML_CPP_ENABLE_FORMAT` that is ON by default when yaml-cpp is root project, and OFF otherwise.
Then target `format` is created `if (YAML_CPP_ENABLE_FORMAT AND YAML_CPP_CLANG_FORMAT_EXE)`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,9 @@ cmake_dependent_option(YAML_CPP_INSTALL
 cmake_dependent_option(YAML_MSVC_SHARED_RT
   "MSVC: Build yaml-cpp with shared runtime libs (/MD)" ON
   "MSVC" OFF)
+cmake_dependent_option(YAML_CPP_ENABLE_FORMAT
+  "Enable format target" ON
+  "CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
 
 set(yaml-cpp-type STATIC)
 set(yaml-cpp-label-postfix "static")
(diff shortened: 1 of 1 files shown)


jbeder:
Solved by 0e6e98e8fb1b4e42eff5958fce981c80e7ac2dfd?

OlivierLDff:
Yes but don't you think
```cmake
cmake_dependent_option(YAML_CPP_FORMAT_SOURCE 
  "Enable format target" ON
  "CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
```

would make more sense?

It would ease integration with `FetchContent`, `CPM`, etc...

jbeder:
Is that right? There are some other PRs (e.g. #1007) that seem to prefer `option` instead of `cmake_dependent_option`.

OlivierLDff:
Oh yes I didn't know `cmake_dependent_option` couldn't be overridable. 

Then following code is the best option:

```cmake
option(YAML_CPP_FORMAT_SOURCE "Format source" ${YAML_CPP_MAIN_PROJECT})
```

That would also be consistent with what have been done for: `YAML_CPP_INSTALL`.

To push it a little bit further even 
```cmake
cmake_dependent_option(YAML_CPP_BUILD_TESTS
  "Enable yaml-cpp tests" ON
  "BUILD_TESTING;CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
```
should behave that way. Let's say if you want to run yaml-cpp tests along with your project unit test using ctest for example

PR #1035, opened 2021-09-24
Update cmakelists to use system googletest if available.

LocutusOfBorg:
There is no need to use the embedded gtest code copy in Linux systems, if they already provide the googletest framework system-wide.
Search for it, and fallback to the embedded one if the system one is not detected.

This patch has been also contributed by Simon Quigley <tsimonq2@debian.org>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,7 @@ jobs:
           os: [ubuntu-latest, windows-latest, macos-latest]
           cxx_standard: [11, 17, 20]
           build: [static, shared]
+          googletest: [build, system]
           generator: ["Default Generator", "MinGW Makefiles"]
           exclude:
           - os: macos-latest
@@ -25,14 +26,26 @@ jobs:
             generator: "MinGW Makefiles"
           - os: ubuntu-latest
             generator: "MinGW Makefiles"
+          - os: macos-latest
+            googletest: system
+          - os: windows-latest
+            googletest: system
       env:
         YAML_BUILD_SHARED_LIBS: ${{ matrix.build == 'shared' && 'ON' || 'OFF' }}
+        YAML_USE_SYSTEM_GTEST: ${{ matrix.googletest == 'system' && 'ON' || 'OFF' }}
         CMAKE_GENERATOR: >-
           ${{format(matrix.generator != 'Default Generator' && '-G "{0}"' || '', matrix.generator)}}
         CMAKE_INSTALL_PREFIX: "${{ github.workspace }}/install-prefix"
         CMAKE_BUILD_TYPE: Debug
       runs-on: ${{ matrix.os }}
       steps:
+
+        - uses: awalsh128/cache-apt-pkgs-action@latest
+          if: matrix.os == 'ubuntu-latest'
+          with:
+            packages: googletest libgmock-dev libgtest-dev
+            version: 1.0
+
         - uses: actions/checkout@v3
 
         - name: Configure
@@ -45,6 +58,7 @@ jobs:
               -D CMAKE_INSTALL_PREFIX="${{ env.CMAKE_INSTALL_PREFIX }}" \
               -D CMAKE_BUILD_TYPE=${{ env.CMAKE_BUILD_TYPE }} \
               -D YAML_BUILD_SHARED_LIBS=${{ env.YAML_BUILD_SHARED_LIBS }} \
+              -D YAML_USE_SYSTEM_GTEST=${{ env.YAML_USE_SYSTEM_GTEST }} \
               -D YAML_CPP_BUILD_TESTS=ON
 
         - name: Build
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -4,11 +4,17 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 set(BUILD_MOCK ON CACHE BOOL "" FORCE)
 set(CMAKE_POLICY_DEFAULT_CMP0048 NEW)
 
-add_subdirectory(
-  "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.11.0"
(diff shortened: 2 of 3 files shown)


MrScriptMaster:

test/CMakeLists.txt
```diff
@@ -4,11 +4,18 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
+if (NOT GTEST_FOUND)
+    add_subdirectory(
+      "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.10.0"
+      "${CMAKE_CURRENT_BINARY_DIR}/prefix")
+    include_directories(SYSTEM "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.10.0/googletest/include")
+else()
+    add_subdirectory(
+      "/usr/src/googletest"
+      "${CMAKE_CURRENT_BINARY_DIR}/prefix")
+    include_directories(SYSTEM "/usr/src/googletest/googlemock/include")
```
jirutka, line 17: These paths are wrong, gtest is usually not installed in this location. You should use pkgconf to get the correct paths.

MrScriptMaster:

test/CMakeLists.txt
LocutusOfBorg: Hello, the path is not exported via cmake, this is for custom build in places where the .so dynamic library is not provided.

MrScriptMaster:

jbeder:
Can you add a variant of the Github action that installs GTest first so you can make sure that codepath works?

MrScriptMaster:

jbeder:
> @jbeder sorry but I fail to find how to install googletest-dev... For me it would be something like
> 
> ```
> apt install googletest libgmock-dev libgtest-dev
> ```

What does this comment mean?

LocutusOfBorg:
> > @jbeder sorry but I fail to find how to install googletest-dev... For me it would be something like
> > ```
> > apt install googletest libgmock-dev libgtest-dev
> > ```
> 
> What does this comment mean?

This means, that I'm not able to parse the travis/github actions yml file...
With this PR, if the container installs the 3 above packages, automatically that system package will be built and used.
So, my idea is to have the installation of the 3 packages in a matrix row, right?

LocutusOfBorg:
is it ok now?

test/CMakeLists.txt
```diff
@@ -4,11 +4,20 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
-include_directories(SYSTEM "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.11.0/googletest/include")
+find_package(GTest)
+if (NOT GTEST_FOUND)
+    add_subdirectory(
+      "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.11.0"
+      "${CMAKE_CURRENT_BINARY_DIR}/prefix")
+    include_directories(SYSTEM "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.11.0/googletest/include")
+else()
+    add_subdirectory(
+      "/usr/src/googletest"
```
jbeder, line 15: This seems very unix-specific, and also maybe requires the package to be stored in a specific location? Is there a more neutral way to use it?

.github/workflows/build.yml
```diff
@@ -28,6 +28,13 @@ jobs:
           ${{format(matrix.generator != 'Default Generator' && '-G "{0}"' || '', matrix.generator)}}
       runs-on: ${{ matrix.os }}
       steps:
+
+        - uses: awalsh128/cache-apt-pkgs-action@latest
```
jbeder, line 32: This seems like a random action; is there no standard form? Or maybe a way to just install the dependency in this action?

Also, this looks like you're just adding the installation to the main flow, rather than adding another test. Can you add a x-product in the matrix instead?

jbeder (changes requested):

MrScriptMaster:

.github/workflows/build.yml
LocutusOfBorg: please help me, I don't really understand yaml

jbeder: This is more of a GitHub actions syntax thing. I think there's documentation about it on the Internet.

test/CMakeLists.txt
jirutka: As I wrote in https://github.com/jbeder/yaml-cpp/pull/1035#discussion_r780800274, this is indeed wrong, even for linux. Just `find_package(GTest)` should be sufficient.

LocutusOfBorg: ok done

.github/workflows/build.yml
LocutusOfBorg: ok changed a little bit the matrix

test/CMakeLists.txt
```diff
@@ -4,11 +4,17 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 set(BUILD_MOCK ON CACHE BOOL "" FORCE)
 set(CMAKE_POLICY_DEFAULT_CMP0048 NEW)
 
-add_subdirectory(
-  "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.11.0"
-  "${CMAKE_CURRENT_BINARY_DIR}/prefix")
-
-include_directories(SYSTEM "${CMAKE_CURRENT_SOURCE_DIR}/gtest-1.11.0/googletest/include")
+if(NOT YAML_USE_SYSTEM_GTEST)
```
jbeder, line 7: Flip the if/else so that it's if(YAML_USE_SYSTEM_GTEST).

jbeder (changes requested):

test/CMakeLists.txt
LocutusOfBorg: done!

jbeder (approved):

PR #1049, opened 2021-10-17
test appveyor with shared libraries for dll export/import

PhilipDeegan:
I'm getting 

```
p\\src\\node_data.cpp(16): error C2370: 'm_amount': redefinition; different storage class
p\\include\yaml-cpp/node/detail/node.h(172): note: see declaration of 'm_amount'
```

so checking if it's a general "cmake" thing too

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,7 +18,7 @@ environment:
 before_build:
    - cmd: mkdir build
    - cmd: cd build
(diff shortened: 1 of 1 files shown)


jbeder:
I'm probably going to drop appveyor now that the github actions are working well. Can this be closed?

PR #1052, opened 2021-10-25
redef with diff storage class in VS

dand-oss:
\yaml\src\node_data.cpp(16): error C2370: 'm_amount': redefinition; different storage class
\yaml\include\yaml-cpp/node/detail/node.h(172): note: see declaration of 'm_amount'


PhilipDeegan:
@dand-oss this was merged indirectly via #1056

PR #1055, opened 2021-10-29
Install pkgconfig and cmake files into arch-dependent locations

tgurr:
This restores the CMake / installed files behaviour of version 0.6.3 which changed in 0.7.0, resulting in arch-dependent files ending up in /usr/share instead of $prefix/libdir. This is required on multiarch/cross layouts setting the prefix to e.g.  `/usr/x86_64-pc-linux-gnu` or `/usr/i686-pc-linux-gnu` so a pkg-config file in /usr/share like

```
prefix=/usr/x86_64-pc-linux-gnu
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: Yaml-cpp
Description: A YAML parser and emitter for C++
Version: 0.7.0
Requires:
Libs: -L${libdir} -lyaml-cpp
Cflags: -I${includedir}
```

would be wrong as it contains arch-specific code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,9 +142,7 @@ set_target_properties(yaml-cpp PROPERTIES
   PROJECT_LABEL "yaml-cpp ${yaml-cpp-label-postfix}"
   DEBUG_POSTFIX "${CMAKE_DEBUG_POSTFIX}")
 
-# FIXME(felix2012): A more common place for the cmake export would be
-# `CMAKE_INSTALL_LIBDIR`, as e.g. done in ubuntu or in this project for GTest
-set(CONFIG_EXPORT_DIR "${CMAKE_INSTALL_DATADIR}/cmake/yaml-cpp")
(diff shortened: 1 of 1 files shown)


jbeder:
This sounds reasonable to me, but I've been bitten quite a bit lately by approving CMake changes I don't understand. Could you add a clause to the github action [here](https://github.com/jbeder/yaml-cpp/blob/master/.github/workflows/build.yml) that tests the behavior you'd like to see? That way it'll guarantee that the behavior you want is preserved even if people later update the CMake file.

tgurr:
Thanks for the suggestion, however I'm not familiar with GitHub Actions and/or writing any tests so I'll most likely not being able to help with or contribute anything in this regard.

tgurr:
See also https://github.com/jbeder/yaml-cpp/pull/1020.

jbeder:
You know what, this is fine; if it breaks someone's build then I guess I can ask them to write a test.

PR #1058, opened 2021-11-03
Fix .pc paths for absolute GNUInstallDirs

r-burns:
GNUInstallDirs provided may be absolute paths, in which case appending
to the install prefix is not correct. We can instead use the provided
CMAKE_INSTALL_FULL_* variables, which are precomputed absolute paths.

https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html


jbeder (approved):

r-burns:
Thanks for the speedy review :)

PR #1061, opened 2021-11-04
Fix for current version of G++11

giorgiozoppi:
The current compiler gnu c++ requires enum class instead of what you used otherwise complains about duplicate identifier. Moreover there's actually no need to include gtest inside the repository since in common Linux distribution  we've already have it. Since i use to my final project i will go deeper to fix this project.

	modified:   src/emitter.cpp
	modified:   src/emitterstate.cpp
	modified:   src/emitterstate.h
	modified:   src/emitterutils.cpp
	modified:   src/emitterutils.h
	modified:   test/CMakeLists.txt


giorgiozoppi:
i close it.

Issue #1046, opened 2021-10-11
"Hide most of non-public..." PR breaks Windows build


noelotterness:
PR #984 broke the Windows use of the yaml-cpp branch. It was fixed when it was backed out on Sept 28th and it has broken the Windows usage again with the revert of the revert.
Here is what happens with the broken submission:
```"D:\a\IIS-Module\IIS-Module\main\Native-IIS.vcxproj" (default target) (2) ->
(ClCompile target) -> 
  D:\a\IIS-Module\IIS-Module\yaml-cpp\include\yaml-cpp\parser.h(13,10): fatal error C1083: Cannot open include file: 'yaml-cpp/dll.h': No such file or directory [D:\a\IIS-Module\IIS-Module\main\Native-IIS.vcxproj]
  D:\a\IIS-Module\IIS-Module\yaml-cpp\include\yaml-cpp\parser.h(13,10): fatal error C1083: Cannot open include file: 'yaml-cpp/dll.h': No such file or directory [D:\a\IIS-Module\IIS-Module\main\Native-IIS.vcxproj]
  D:\a\IIS-Module\IIS-Module\yaml-cpp\include\yaml-cpp\parser.h(13,10): fatal error C1083: Cannot open include file: 'yaml-cpp/dll.h': No such file or directory [D:\a\IIS-Module\IIS-Module\main\Native-IIS.vcxproj]
  D:\a\IIS-Module\IIS-Module\yaml-cpp\include\yaml-cpp\parser.h(13,10): fatal error C1083: Cannot open include file: 'yaml-cpp/dll.h': No such file or directory [D:\a\IIS-Module\IIS-Module\main\Native-IIS.vcxproj]
```

noelotterness:
First I cannot provide the source code since it is not a public repository. The build system files are what are generated by Visual Studio to build a native C++ DLL for IIS. I include the directories where the generated source is placed. The problem is that there is something in the yaml-cpp source that requires the dll.h file and that is no longer being generated for the Windows 2019 build.
Perhaps it has been moved to a new location?
The Project include directive is for "yaml-cpp\include" with some pathing to obtain the correct location based upon either a local build or a GitHub build.
The problem is that it all used to work prior to your commit and the code was configured as described in the documentation. If there is something different I need to do to make it work now (also I do not use ExternalProject_Add()), that would be great.

pinotree:
[...]

PR #1063, opened 2021-11-19
Revert "Revert "Revert "Hide most of non-public symbols by default (#984)" (#1038)""

jbeder:
Reverts jbeder/yaml-cpp#1045

See the discussion in #1046. The issue with this PR is that it starts using the build system to generate source code, which (a) makes it no longer possible to just add the sources to a project without using CMake, and (b) requires that to support any other build system (e.g. Bazel, which I'd like to support), we'd have to duplicate this source-generating logic in that other build system also.

@pinotree If you see this, I'm open to PRs that solve the problem you were trying to solve without generating source code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,11 +14,6 @@ include(CMakeDependentOption)
 include(CheckCXXCompilerFlag)
 include(GNUInstallDirs)
 include(CTest)
-include(GenerateExportHeader)
-
-set(CMAKE_C_VISIBILITY_PRESET hidden)
-set(CMAKE_CXX_VISIBILITY_PRESET hidden)
-set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
 find_program(YAML_CPP_CLANG_FORMAT_EXE NAMES clang-format)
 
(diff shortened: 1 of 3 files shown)


jbeder:
Aaaand there's an unresolved linker error for the `YAML::Null` and `YAML::detail::node::m_amount` symbols. I don't get it, since they appear to be defined. But I don't have access to a Windows machine, so if anyone sees this, I'd love some help getting this PR to work.

hkarel:
I passed the tests, but essentially I repeated my own decision about which I wrote in #1046.  See my branch [revert-1045-hidden-visibility-again](https://github.com/hkarel/yaml-cpp/tree/revert-1045-hidden-visibility-again), commits: [WIP](https://github.com/hkarel/yaml-cpp/commit/b8818e810a6d258164d5a42781b8bdd376938b4f), [WIP2](https://github.com/hkarel/yaml-cpp/commit/e4e735adc966d103a1d9b22bdd7c0ad4a29e75c9), [WIP3](https://github.com/hkarel/yaml-cpp/commit/11f67d70b5976be100022f2815f871cc8bb49e5e), [WIP4](https://github.com/hkarel/yaml-cpp/commit/1ef0f6c20aab1458357ba16d2eb46324e082a965)
My dll.h file - it compilation from cmake-generation for GCC, MSVC, MIGW

jbeder:
If you’re happy with the code, can you submit a PR?

hkarel:
Done. PR #1064

PR #1064, opened 2021-11-20
The dll.h is made static again (revert commit 0733aeb4)

hkarel:
New dll.h is assembly from cmake-generation for GCC, MSVC, MIGW

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,11 +31,13 @@ cmake_dependent_option(YAML_MSVC_SHARED_RT
   "MSVC: Build yaml-cpp with shared runtime libs (/MD)" ON
   "MSVC" OFF)
 
-set(yaml-cpp-type STATIC)
-set(yaml-cpp-label-postfix "static")
 if (YAML_BUILD_SHARED_LIBS)
   set(yaml-cpp-type SHARED)
   set(yaml-cpp-label-postfix "shared")
+else()
+  set(yaml-cpp-type STATIC)
+  set(yaml-cpp-label-postfix "static")
+  add_definitions(-DYAML_CPP_STATIC_DEFINE)
 endif()
 
 set(build-shared $<BOOL:${YAML_BUILD_SHARED_LIBS}>)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,11 +14,6 @@ include(CMakeDependentOption)
 include(CheckCXXCompilerFlag)
 include(GNUInstallDirs)
 include(CTest)
-include(GenerateExportHeader)
-
-set(CMAKE_C_VISIBILITY_PRESET hidden)
-set(CMAKE_CXX_VISIBILITY_PRESET hidden)
-set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
 find_program(YAML_CPP_CLANG_FORMAT_EXE NAMES clang-format)
 
@@ -86,7 +81,6 @@ set_property(TARGET yaml-cpp
 target_include_directories(yaml-cpp
   PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
   PRIVATE
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>)
@@ -142,12 +136,6 @@ write_basic_package_version_file(
   "${PROJECT_BINARY_DIR}/yaml-cpp-config-version.cmake"
   COMPATIBILITY AnyNewerVersion)
 
-generate_export_header(yaml-cpp
-  BASE_NAME YAML_CPP
-  EXPORT_FILE_NAME "${PROJECT_BINARY_DIR}/include/yaml-cpp/dll.h"
-  EXPORT_MACRO_NAME YAML_CPP_API
-)
-
 configure_file(yaml-cpp.pc.in yaml-cpp.pc @ONLY)
 
 if (YAML_CPP_INSTALL)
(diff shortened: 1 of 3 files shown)


include/yaml-cpp/dll.h
```diff
@@ -1,33 +1,61 @@
...
+#      else
+         /* We are using this library */
+#        pragma message( "Defining YAML_CPP_API for DLL import" )
+#        define YAML_CPP_API __declspec(dllimport)
+#      endif
+#    endif
+#    ifndef YAML_CPP_NO_EXPORT
+#      define YAML_CPP_NO_EXPORT
+#    endif
+#  else /* No _MSC_VER */
```
jbeder, line 29: What cases (compilers/operating systems) does this else case catch? What's the reason for the differing definition?

CMakeLists.txt
```diff
@@ -31,11 +31,13 @@ cmake_dependent_option(YAML_MSVC_SHARED_RT
...
 
-set(yaml-cpp-type STATIC)
-set(yaml-cpp-label-postfix "static")
 if (YAML_BUILD_SHARED_LIBS)
   set(yaml-cpp-type SHARED)
   set(yaml-cpp-label-postfix "shared")
+else()
+  set(yaml-cpp-type STATIC)
+  set(yaml-cpp-label-postfix "static")
+  add_definitions(-DYAML_CPP_STATIC_DEFINE)
```
jbeder, line 40: It seems more natural to me to add a definition for the shared lib, which is the unusual case. (I don't feel *super* strongly, but since it looks like it doesn't matter one way or another, I thought I'd say something.)

include/yaml-cpp/dll.h
```diff
@@ -1,33 +1,61 @@
...
-     (__GNUC__ >= 4))  // GCC supports "pragma once" correctly since 3.4
-#pragma once
+// Definition YAML_CPP_STATIC_DEFINE using to building YAML-CPP as static
+// library (definition created by CMake or defined manually)
+
+// Definition yaml_cpp_EXPORTS using to building YAML-CPP as dll/so library
+// (definition created by CMake or defined manually)
+
+#ifdef YAML_CPP_STATIC_DEFINE
+#  define YAML_CPP_API
```
jbeder, line 11: Do you need to define these symbols at all in the static case?

jbeder (changes requested):
Thanks, this is awesome. Could you rebase this against master and reopen the pull request? That will trigger the actions (I'm not sure how to trigger them otherwise) so it'll run the automatic tests.

include/yaml-cpp/dll.h
hkarel: As far as I know, in Windows all functions are hidden by default for export. You must explicitly use the `__declspec(dllexport)` directive to export functions from the library. On the contrary, in Linux all functions by default are exports. Sometimes in Linux (often for reasons of project build speed) there is a need to hide some of the functions. To do this, using the directive `__attribute__((("hidden"))`. If in the future will be need to hide some of the functions for assembly on Linux/Unix platforms, this macro will be useful. Now it's just a formal record.

CMakeLists.txt
hkarel: I am not a specialist at CMake, so this change is made by virtue of my little knowledge in this assembly system. I guess you can change this section as you see fit.

include/yaml-cpp/dll.h
hkarel: Yes, for a static assembly, the YAML_CPP_STATIC_DEFINE must be defined in the yaml-cpp library, as well as in the target application. 
For the static case, the macros YAML_CPP_API and YAML_CPP_NO_EXPORT must be dummy.

hkarel:
> Could you rebase this against master and reopen the pull request? That will trigger the actions (I'm not sure how to trigger them otherwise) so it'll run the automatic tests. 

I relocated PR to master

include/yaml-cpp/dll.h
PhilipDeegan: @jbeder it's much the same as the suggested form from gcc https://gcc.gnu.org/wiki/Visibility

PhilipDeegan:
> > Could you rebase this against master and reopen the pull request? That will trigger the actions (I'm not sure how to trigger them otherwise) so it'll run the automatic tests.
> 
> I relocated PR to master

it's possible you might have to make a new PR if you missing the github actions file(s), although @jbeder can [probably run it manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow)

hkarel:
> > > Could you rebase this against master and reopen the pull request? That will trigger the actions (I'm not sure how to trigger them otherwise) so it'll run the automatic tests.
> > 
> > 
> > I relocated PR to master
> 
> it's possible you might have to make a new PR if you were missing the github actions file(s) originally, although @jbeder can [probably run it manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow)

I tried creating a new PR (copy of this PR). It didn't work, I was redirected to this page. Apparently github keeps track of PR copies.

PhilipDeegan:
it normally lets you create a new one but shows you the existing one in case you just want that one 
but it looks like you have it done here https://github.com/hkarel/yaml-cpp/actions/runs/1484364184

include/yaml-cpp/dll.h
jbeder: What I mean is, in the static case, do you need to define YAML_CPP_API at all?

include/yaml-cpp/dll.h
jbeder: Thanks, sounds good to me.

CMakeLists.txt
jbeder: OK, I guess it really doesn't matter much. I'll merge the code now and any improvements can be made after the fact.

jbeder (approved):

include/yaml-cpp/dll.h
hkarel: M-m-m...perhaps I don't quite understand the question (translation difficulties). But still I'll try to answer: theoretically, for the variant of static build the definition YAML_CPP_API is not needed, but from a practical point of view, YAML_CPP_API must be present - otherwise the project will not build

PR #1069, opened 2021-11-25
Add an `into()` operator on Node to do optional assignment.

psigen:
This adds an `into()` operator to the Node class modeled off the `get_to()` operator in the popular nlohmann::json library.
(See: https://github.com/nlohmann/json#basic-usage)

This operator assigns a casted value to an existing variable if that node has a defined value, and otherwise leaves the variable unchanged.  It returns `true` if the variable was overwritten and `false` if the variable was unchanged.

This operator is extremely useful in applications where YAML is being used as a configuration source that overrides default values coming from another source, e.g.:

```c++
struct MyConfig {
  int foo = 4;
  double bar = 5.0;
  std::string baz = "six";
};

MyConfig config;
node["foo"].into(config.foo);
node["bar"].into(config.bar);

if (!node["baz"].into(config.baz))
  std::cerr << "Using default value for .baz!" << std::endl;
```

### DIFF
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -163,6 +163,14 @@ inline T Node::as(const S& fallback) const {
   return as_if<T, S>(*this)(fallback);
 }
 
+template <typename T>
+inline bool Node::into(T& target) const {
+  if (!this->IsDefined())
+    return false;
+  target = this->as<T>();
+  return true;
+}
+
 inline const std::string& Node::Scalar() const {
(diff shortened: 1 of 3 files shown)


jbeder:
I'm not sure this new API adds enough functionality to be worth adding. I'll close this PR, but if you want to advocate for this, then open an issue, and if you can get a lot of people +1ing that issue, then I'll reconsider.

psigen:
Created an issue to track interest https://github.com/jbeder/yaml-cpp/issues/1073

PR #1075, opened 2021-12-16
[mingw] add pipeline for the MinGW compiler.

TheVice:
Hello.

Proposed to add build and test binaries at MinGW environment.

Currently, according to used OS based loop _[ubuntu-latest, windows-latest, macos-latest]_ at pipeline workflow, **MinGW** pipeline required some of duplication in code in order to proper build and test only when Windows OS used.
If pipeline will changed to the compiler based loop, for example _[clang, gcc, mingw, msvc]_, it will be less duplicate code.

From better resources view - OS based loop, that currently used, is better - because Windows runner run only once and used to build for two compilers - **MSVC** and **MinGW**.

Thank you.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,31 +10,32 @@ jobs:
       strategy:
         matrix:
           os: [ubuntu-latest, windows-latest, macos-latest]
+          build: [static, shared]
+          generator: ["Default Generator", "MinGW Makefiles"]
+          exclude:
+          - os: macos-latest
+            build: shared
+          - os: macos-latest
+            generator: "MinGW Makefiles"
+          - os: ubuntu-latest
+            generator: "MinGW Makefiles"
+      env:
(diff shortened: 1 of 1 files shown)


jbeder (changes requested):
This is great, and I really appreciate the improvement; since you're adding this, would you be willing to do a small refactor to avoid the code duplication? I think you can use the `matrix` field to simplify this. Something like:

```
matrix:
  os: [ubuntu-latest, windows-latest, macos-latest]
  build: [static, shared]
include:
  os: [windows-latest]
  generator: [MinGW Makefiles] // I think this would cause the steps to run once with no 'generator' variable, and then once with this pair of windows-latest/MinGW
exclude: // this is already excluded in the default build
  os: macos-latest
  build: shared
```

TheVice:
Sure.

For some reason approach with 'include' generator only for the 'windows-latest' do not success - static and shared builds in this case will be outside of the loop for 'mingw', but both 'exclude' of 'mingw' generator for Ubuntu and macOS is work as expected.

jbeder (approved):

Issue #1031, opened 2021-09-10
enable static linking of yaml-cpp


christian-rauch:
Querying the static yaml-cpp libraries via
```bash
pkg-config --libs --static yaml-cpp
```
provides the same linking flags as for the shared library:
```
-L/usr//usr/lib/x86_64-linux-gnu -lyaml-cpp
```
whereas it should provide the static library `-lyaml-cpp.a`.

For CMake, you could adapt the solution from: https://gitlab.inria.fr/Puffertools/pufferscaletest/pufferscale-exp/blob/master/cmake/Modules/FindYamlCpp.cmake which lets one set `YAMLCPP_USE_STATIC_LIBS` before searching the library.

gelldur:
I build it like this:

```
###############################################################################
TYPE=Release
INSTALL=${BUILD_DIR}/install-${TYPE}
BUILD=build-${TYPE}
echo "Building: $BUILD <-------------------------------------------------------"
(cd ${BUILD_DIR} && rm -rf $BUILD && mkdir -p $BUILD) || exit 1
(cd ${BUILD_DIR}/$BUILD && cmake ${DIR}/../yaml-cpp \
	-DBUILD_SHARED_LIBS=OFF \
	-DYAML_BUILD_SHARED_LIBS=OFF \
	-DBUILD_TESTING=OFF \
	-DCMAKE_BUILD_TYPE=${TYPE} \
	-DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang \
	-DCMAKE_INSTALL_PREFIX=${INSTALL}) || exit 1
(cd ${BUILD_DIR}/$BUILD && cmake --build . --target install -j 8) || exit 1
(cd ${BUILD_DIR} && rm -rf $BUILD) # Cleanup
###############################################################################

###############################################################################
TYPE=Debug
INSTALL=${BUILD_DIR}/install-${TYPE}
BUILD=build-${TYPE}
echo "Building: $BUILD <-------------------------------------------------------"
(cd ${BUILD_DIR} && rm -rf $BUILD && mkdir -p $BUILD) || exit 1
(cd ${BUILD_DIR}/$BUILD && cmake ${DIR}/../yaml-cpp \
	-DBUILD_SHARED_LIBS=ON \
	-DYAML_BUILD_SHARED_LIBS=ON \
	-DBUILD_TESTING=OFF \
	-DCMAKE_BUILD_TYPE=${TYPE} \
[...]

PR #1076, opened 2021-12-16
set CMAKE_POSITION_INDEPENDENT_CODE if BUILD_SHARED_LIBS=OFF

christian-rauch:


gelldur:
This will set PIC for all targets (what if someone adds library as `sub-directory`, maybe set_property would be better?
```set_property(TARGET lib1 PROPERTY POSITION_INDEPENDENT_CODE ON)```

christian-rauch:
If `BUILD_SHARED_LIBS=OFF` then every library, that hasn't explicitly set `SHARED` or `STATIC` will build as `STATIC`. I think that PIC should therefore apply to all static libraries that the cmake project builds.

milipili:
Property on targets should be used instead. It is not rare that CMake projects are included into another, and deciding to set `CMAKE_POSITION_INDEPENDENT_CODE` will just break everything.

It should always be per target. Also why setting position indepdendant code if building a static library? In many cases we dont want that option. Only if it is included into a bigger shared library

christian-rauch:
> In many cases we dont want that option.

Why? You will need that option if you want to link your static library into an executable.

christian-rauch:
> Property on targets should be used instead.

I set this for the target `yaml-cpp` only. I think this is the only target in the project.

jbeder (approved):

jbeder:
Fixes #1031

Issue #774, opened 2019-10-14
Cannot find headers (and libraries) as CMake variables are unset


mattia-:
Hello,

as per issue title, I'm having a problem when I try to use yaml-cpp in a project I'm working on. Namely, the compiler can't find header files (but I suppose the same will go for libraries) when I try to compile. 

This is due to the fact that the file used by CMake to set the paths of include and link directories does not set the proper values for the variables. The content of `share/cmake/yaml-cpp/yaml-cpp-config.cmake` after installation is;:
```
# - Config file for the yaml-cpp package
# It defines the following variables
#  YAML_CPP_INCLUDE_DIR - include directory
#  YAML_CPP_LIBRARIES    - libraries to link against

# Compute paths
get_filename_component(YAML_CPP_CMAKE_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
set(YAML_CPP_INCLUDE_DIR "")

# Our library dependencies (contains definitions for IMPORTED targets)
include("${YAML_CPP_CMAKE_DIR}/yaml-cpp-targets.cmake")

# These are IMPORTED targets created by yaml-cpp-targets.cmake
set(YAML_CPP_LIBRARIES "")
```

As you can see, the variables `YAML_INCLUDE_DIR` and `YAML_CPP_LIBRARIES` are empty. I believe the problem may lie in the input file `yaml-cpp-config.cmake.in`:
```
# - Config file for the yaml-cpp package
# It defines the following variables
#  YAML_CPP_INCLUDE_DIR - include directory
#  YAML_CPP_LIBRARIES    - libraries to link against

# Compute paths
get_filename_component(YAML_CPP_CMAKE_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
set(YAML_CPP_INCLUDE_DIR "@CONFIG_INCLUDE_DIRS@")

# Our library dependencies (contains definitions for IMPORTED targets)
include("${YAML_CPP_CMAKE_DIR}/yaml-cpp-targets.cmake")

# These are IMPORTED targets created by yaml-cpp-targets.cmake
set(YAML_CPP_LIBRARIES "@EXPORT_TARGETS@")
```

Indeed, `grep`-ing for `CONFIG_INCLUDE_DIRS` or `EXPORT_TARGETS` yields no results in the whole repository, so substitution should give an empty string.

Am I missing something?

Thanks!

mattia

winterheart:
Hello.

[...]

PR #1077, opened 2021-12-30
Cmake export fix

felix2010:
I tried to fix the generation of the file `yaml-cpp-config.cmake`.  Former the variables
`YAML_CPP_CMAKE_DIR` and `YAML_CPP_LIBRARIES` where empty within this file,
now they are set appropriately.

I reworked the file accordingly to cmake documentation as described in the commit message.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,30 +141,30 @@ write_basic_package_version_file(
 configure_file(yaml-cpp.pc.in yaml-cpp.pc @ONLY)
 
 if (YAML_CPP_INSTALL)
-	install(TARGETS yaml-cpp
+  install(TARGETS yaml-cpp
     EXPORT yaml-cpp-targets
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
-	install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/
+  install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
-		FILES_MATCHING PATTERN "*.h")
+                FILES_MATCHING PATTERN "*.h")
   install(EXPORT yaml-cpp-targets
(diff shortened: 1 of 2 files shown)


jbeder:
What behavior will this fix? (That is, how do you reproduce the problem you're seeing, to verify that this fixes it?)

Once you specify that (e.g. sequence of command lines), let's put them in the automated github tests so we can keep these passing.

felix2010:
Configure the project and check the generated `yaml-cpp-config.cmake`
```
cmake -B build && grep -E '\(YAML_CPP_(LIBRARIES|INCLUDE)' build/yaml-cpp-config.cmake
```
Actual output is
 ```
set(YAML_CPP_INCLUDE_DIR "")
set(YAML_CPP_LIBRARIES "")
```
but expected are non empty variables, i.e.
```
set_and_check(YAML_CPP_INCLUDE_DIR "${PACKAGE_PREFIX_DIR}/include")
set(YAML_CPP_LIBRARIES "yaml-cpp")
```
I'm not familiar with github workflow/test, thus I need your advice how to bring this into the automated tests.

jbeder:
Thanks! And what's the goal of having those variables set? That is, what subsequent command with use them and expect them to be nonempty?

felix2010:
The variables are documented entities in `yaml-cpp-config.cmake` introduced with a397ad29252b60f0549a4edb389a46a37577cd02.
Afaik these are convenience variables that each `cmake` package should provide. Initially the variables got expanded in the right way, but this behavior was brocken in 5e9cb0128d1b600ff74505429269fd4dff367f4a.

`ubuntu/jammy` ships with a `yaml-cpp-config.cmake` that sets the variables in a correct way, but only because the maintainer of the package patched the sources (content of `yaml-cpp_0.6.2-4ubuntu1.debian.tar.xz` from https://packages.ubuntu.com/source/focal/yaml-cpp).

`yaml-cpp` is a great library, I like it and I like to have it in my preferred linux distribution. If we require patches from all distribution integrators we risk, that distributions ship with old versions of the library or dropping it from the distro.
That was the main motivation for my PR.

jbeder:
I'm going to merge this because I'm too lazy to try to figure out how to test it properly, and it very likely is the right solution. At some point, it would be nice to have a way to know if this is broken, so I'm open to PRs that just have a new test case in the Github actions file.

jbeder:
Fixes #774

PR #1087, opened 2022-03-03
Seamlessly support non-default-constructable custom class deserialization

### LINKED ISSUE #993: custom converter for non default-constructable data-types

mzoll:
The current API of the `convert<T>::decode` construct is limited to default constructable structs with visible members only. This is limiting and does not follow RAII.

The here patch allows to declare in the custom 'struct convert<T>' for any type `T` a decode method by slightly different, but more instructive, signature `T convert<T>::decode(const Node& node)`. Within that method `T` can be constructed by the user from the data serialized from `node` and the result returned without the need to manipulate the members of an uninitialized reference. Problems in the deserialization process are marked by throwing the `DecodeException`, effectively replacing the old convoluted bool return signaling deserialization success.

While users are free to implement this new API the old API is still supported in static code-path switches based on the `decode`-signature detected.

The complete documentation is found at the bottom of `docs/Tutorial.md`

tests have been added and pass.

This patch closes #993
This patch also shows an alternative solution to merge request #1010


docs/Tutorial.md
```diff
@@ -197,5 +198,50 @@ Then you could use `Vec3` wherever you could use any other type:
...
 Vec3 v = node["start"].as<Vec3>();
-node["end"] = Vec3(2, -1, 0);
-```
\ No newline at end of file
+node["end"] = Vec3{2, -1, 0};
+```
+
+Observe that in the above example the custom type is, decalred as 
+a struct, explicit default constructable and all its members are
+exposed. For non default constructable types like
```
SGSSGene, line 206: Minor typo:
```suggestion
Observe that in the above example the custom type is, declared as 
a struct, explicit default constructable and all its members are
exposed. For non default constructable types like
```

docs/Tutorial.md
SGSSGene, line 190: Wrong return value. Should be `rhs` not `true`.

```suggestion
    return rhs;
```

docs/Tutorial.md
```diff
@@ -197,5 +198,50 @@ Then you could use `Vec3` wherever you could use any other type:
...
+
+```cpp
+class NonDefCtorVec3 : public Vec3 {
+  using Vec3::x;
+  using Vec3::y;
+  using Vec3::z;
+ public:
+  NonDefCtorVec3(double x, double y, double z)
+      : Vec3() { this->x=x; this->y=y; this->z=z;
+  };
```
SGSSGene, line 216: Proper initialization of variables:
```suggestion
  NonDefCtorVec3(double x, double y, double z)
      : Vec3{x, y, z}
  {}
```

docs/Tutorial.md
SGSSGene, line 193: In general, maybe `convert<Vec3>` should keep the old API?

include/yaml-cpp/node/convert.h
```diff
@@ -57,18 +59,24 @@ struct convert<Node> {
     rhs.reset(node);
     return true;
   }
+
+//  static Node decode(const Node& node) {
+//    throw std::runtime_error("this should not have been encountered");
+//    Node rhs;
+//    rhs.reset(node);
+//    return rhs;
+//  }
```
SGSSGene, line 68: Can this be removed?

test/node/node_test.cpp
```diff
@@ -42,6 +42,30 @@ template <class T> using CustomList = std::list<T,CustomAllocator<T>>;
...
+  NonDefCtorVec3(double x, double y, double z)
+      : Vec3() {
+     this->x=x;
+     this->y=y;
+     this->z=z;
+  };
+  bool operator==(const NonDefCtorVec3& rhs) const {
+    return x == rhs.x && y == rhs.y && z == rhs.z;
+  }
+};
```
SGSSGene, line 66: Needs same adjustments as in the Tutorial.md

docs/Tutorial.md
```diff
@@ -197,5 +198,50 @@ Then you could use `Vec3` wherever you could use any other type:
...
+
+Observe that in the above example the custom type is, decalred as 
+a struct, explicit default constructable and all its members are
+exposed. For non default constructable types like
+
+```cpp
+class NonDefCtorVec3 : public Vec3 {
+  using Vec3::x;
+  using Vec3::y;
+  using Vec3::z;
```
SGSSGene, line 212: Using statements are not required.
```suggestion
```

SGSSGene (changes requested):
I see the need/desire for a converter that works for non defaultable types.

Changing the API is not an option. This would break to much and currently the goal is to maintain this library to keep it compatible and reduce number of bugs.

So it would need some mechanism that allows for old and new API to work together.

Additionally, I am not satisfied with the conversion from exception to return value paths:
```
try {
   ...
  } catch(const conversion::DecodeException& e) {
    return false; //not doing this breaks upstream functionality
  } catch (...) {
    throw;
  }
```
This in my eyes should not be. In c++23 I would use `std::expected` and in c++17 `std::optional` not sure what to do in c++11. Maybe `std::variant`?

SGSSGene:
Closed in favor of #1488

PR #1099, opened 2022-04-19
Improve build and fix single cr on windows

### LINKED ISSUE #1098: PR #1094 causes regression on Windows
Making a `Break` to recognize a single `\r` instead of a `\r\n` as it is used on Windows, breaks parsing of block values. For example, when a PGP key is given as a multiline value, then the next key/value pair is not recognized correctly.

vehre-x41:
PR #1094 introduced a regression when block value in a file with CRNL line endings was parsed. The Break regular expression needed to consume both the CR and the NL, but it also needs to consume single CR, when they are not followed by a NL. This PR fixes #1098 and additionally tries to improve bazel build.

### DIFF
--- a/src/exp.h
+++ b/src/exp.h
@@ -37,7 +37,7 @@ inline const RegEx& Blank() {
   return e;
 }
 inline const RegEx& Break() {
-  static const RegEx e = RegEx('\n') | RegEx("\r");
+  static const RegEx e = RegEx('\n') | RegEx("\r\n") | RegEx('\r');
   return e;
 }
 inline const RegEx& BlankOrBreak() {
--- a/test/integration/load_node_test.cpp
+++ b/test/integration/load_node_test.cpp
@@ -345,5 +345,20 @@ TEST(NodeTest, LoadTagWithNullScalar) {
   EXPECT_TRUE(node.IsNull());
 }
 
+TEST(LoadNodeTest, BlockCRNLEncoded) {
+  Node node = Load(
+      "blockText: |\r\n"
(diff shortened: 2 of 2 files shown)


vehre-x41:
Suggestions welcome how to make the bazel build file independent from envoy (I am not a bazel expert).

jbeder (changes requested):
Can you:

1. Add a test that demonstrates how this fails on Windows before this PR, and succeeds now? (There's a continuous Windows build, which obviously didn't catch this issue, so it would be nice to add this case to the test suite.)

2. Split off the Bazel changes, which I have some more comments on.

vehre-x41:
Yes, sure. Could you point me to a test.cc, where this makes most sense?

The testcase would comprise some yaml, like this:
```
certificate_chain:
                inline_string: |
                  -----BEGIN CERTIFICATE-----
                  MIICqDCCAZACCQCquzpHNpqBcDANBgkqhkiG9w0BAQsFADAWMRQwEgYDVQQDDAtm
<snip>
                  m+/R4XnmL4cKQ+5Z
                  -----END CERTIFICATE-----
              private_key:
<snip>
```

vehre-x41:
#1100 now has the split up bazel build file. I will remove it from this pr as soon as the testcase is added.

jbeder:
I think test/integration/load_node_test.cpp would be appropriate.

jbeder (approved):
Thanks!

jbeder (approved):
Thanks!

jbeder (approved):

PR #1100, opened 2022-04-19
Improve bazel build for Windows.

vehre-x41:
Windows builds need some defines being set to use the static linking.
Also add bazel builds to the CI.

Split from #1099 .

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,5 +58,5 @@ jobs:
           shell: bash
           run: |
             cd "${{ github.workspace }}"
-            bazel test :all
+            bazel test test
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
     branches: [ master ]
   workflow_dispatch:
 jobs:
-  build:
+  cmake-build:
       strategy:
         matrix:
           os: [ubuntu-latest, windows-latest, macos-latest]
@@ -40,7 +40,7 @@ jobs:
           shell: bash
           run: cd build && ctest --output-on-failure
 
-  bz-build:
+  bazel-build:
       strategy:
         matrix:
           os: [ubuntu-latest, windows-latest, macos-latest]
@@ -52,13 +52,11 @@ jobs:
           shell: bash
           run: |
             cd "${{ github.workspace }}"
-            bazel build '//:yaml-cpp'
+            bazel build ...
 
         - name: Test
           shell: bash
           run: |
             cd "${{ github.workspace }}"
-            # For some reason 'bazel test //test:test' gets the first / stripped on windows in CI.
-            # Therefore use this ugly version, which is working on all platforms.
-            bazel test test
+            bazel test ...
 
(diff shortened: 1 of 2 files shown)


jbeder (changes requested):
For the Bazel build, I've been hoping to get it into a more supported form, i.e., tested. Could you add a clause in the GitHub action that tests the Bazel compilation on all the same relevant platforms?

That will also help verify that this change works as intended.

vehre-x41:
Sorry for the force pushes, but I did not want to clutter the branch too much. This works now.

Is this what you wanted?

.github/workflows/build.yml
```diff
@@ -39,3 +39,26 @@ jobs:
         - name: Test
           shell: bash
           run: cd build && ctest --output-on-failure
+
+  bz-build:
```
jbeder, line 43: Name this "bazel-build" (no need for abbreviation here).

And can you rename the other one "cmake-build"?

.github/workflows/build.yml
```diff
@@ -39,3 +39,26 @@ jobs:
...
+          os: [ubuntu-latest, windows-latest, macos-latest]
+      runs-on: ${{ matrix.os }}
+      steps:
+        - uses: actions/checkout@v2
+
+        - name: Build
+          shell: bash
+          run: |
+            cd "${{ github.workspace }}"
+            bazel build '//:yaml-cpp'
```
jbeder, line 55: `bazel build ...`

is more common, I believe.

.github/workflows/build.yml
```diff
@@ -39,3 +39,26 @@ jobs:
...
+
+        - name: Test
+          shell: bash
+          run: |
+            cd "${{ github.workspace }}"
+            # For some reason 'bazel test //test:test' gets the first / stripped on windows in CI.
+            # Therefore use this ugly version, which is working on all platforms.
+            bazel test test
```
jbeder, line 63: And similarly

`bazel test ...`

jbeder (changes requested):

.github/workflows/build.yml
vehre-x41: Here, I only got `bazel build :all` to work as expected.

.github/workflows/build.yml
vehre-x41: And this needed to be 'bazel test test'.

'bazel test :all' did not work, because it didn't find any tests.

.github/workflows/build.yml
vehre-x41: Done

.github/workflows/build.yml
vehre-x41: Ok, but that did not work `bazel test ...` reports "no tests found".
`bazel test :all` tries to do something with the imported package repo:http.bzl, which does not apply.

vehre-x41: Now, how to we proceed? This is the best solution I, not being a bazel expert, can come up with...

jbeder (approved):

PR #1102, opened 2022-04-22
fix: revert ABI breakage

liujianqiang-niu:
add patch to revert abi breakage.

Log:

### DIFF
--- a/include/yaml-cpp/node/detail/node_data.h
+++ b/include/yaml-cpp/node/detail/node_data.h
@@ -81,7 +81,7 @@ class YAML_CPP_API node_data {
                     shared_memory_holder pMemory);
 
  public:
-  static const std::string& empty_scalar();
+  static std::string empty_scalar;
 
  private:
   void compute_seq_size() const;
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -166,13 +166,13 @@ inline T Node::as(const S& fallback) const {
 inline const std::string& Node::Scalar() const {
   if (!m_isValid)
     throw InvalidNode(m_invalidKey);
-  return m_pNode ? m_pNode->scalar() : detail::node_data::empty_scalar();
+  return m_pNode ? m_pNode->scalar() : detail::node_data::empty_scalar;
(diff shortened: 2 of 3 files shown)


liujianqiang-niu:
> 

The symbol _ZN4YAML6detail9node_data12empty_scalarB5cxx11E (YAML::detail::node_data::empty_scalar[abi:cxx11]) is no longer provided by /lib64/libyaml-cpp.so.0.7.

jbeder:
It looks like this is from #559. But your PR passes the windows/shared tests. Do you have any idea why the original PR felt like they needed to change it?

shenlebantongying:
This damage was already done, downstream already fixed this breakage in some way (like recompiling the chain of dependency). If you revert or release again, they have to do it all over again. Please evaluate if they have to redo everything before reverting!

This bug might be an expected behaviour now.

There are more, but here are some examples:



```
https://stackoverflow.com/questions/52713571/mongod-symbol-lookup-error-mongod-undefined-symbol-zn4yaml6detail9node-data
[https://forums.opensuse.org/showthread.php/538693-Blender-failed-to-launch-after-update-symbol-lookup-error-undefine-symbol 1](https://forums.opensuse.org/showthread.php/538693-Blender-failed-to-launch-after-update-symbol-lookup-error-undefine-symbol)
https://bugzilla.opensuse.org/show_bug.cgi?id=1160171
https://bugzilla.redhat.com/show_bug.cgi?id=1773357
https://bugzilla.redhat.com/show_bug.cgi?id=1774164
https://bbs.archlinux.org/viewtopic.php?id=248014
https://github.com/jbeder/yaml-cpp/issues/484
```

shenlebantongying:
IIRC It happened between 0.6.2 -> 0.6.3.

A minor version bump -> The `libc.so.6` name didn't change -> distros assume ABI didn't change -> breakages.

Some distro maybe considers 0.6->0.7 is also a minor change, thus we got this pull request to revert it.

(I cannot confirm but it seems to recompile can just solve the problem.)

jbeder:
OK, sounds good, I'll close this PR.

PR #1112, opened 2022-05-31
Small perfomance optimization in Directives and Tag methods

hyperxor:
I'm making a small perfomance improvement for `Directives::TranslateTagHandle` and `Tag::Translate` methods. Having `const` in return type forces coping operation instead of moving one. You can read [here](https://stackoverflow.com/questions/57761077/return-const-value-prevent-move-semantics) about it or watch [this](https://www.youtube.com/watch?v=dGCxMmGvocE) video.

### DIFF
--- a/src/directives.cpp
+++ b/src/directives.cpp
@@ -3,7 +3,7 @@
 namespace YAML {
 Directives::Directives() : version{true, 1, 2}, tags{} {}
 
-const std::string Directives::TranslateTagHandle(
+std::string Directives::TranslateTagHandle(
     const std::string& handle) const {
   auto it = tags.find(handle);
   if (it == tags.end()) {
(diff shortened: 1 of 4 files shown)


src/directives.cpp
```diff
@@ -3,15 +3,19 @@
...
 Directives::Directives() : version{true, 1, 2}, tags{} {}
 
-const std::string Directives::TranslateTagHandle(
+std::string Directives::TranslateTagHandle(
     const std::string& handle) const {
+  std::string result;
   auto it = tags.find(handle);
   if (it == tags.end()) {
     if (handle == "!!")
-      return "tag:yaml.org,2002:";
```
jbeder, line 11: Removing const makes sense, but why do you need to get rid of the early returns?

hyperxor: @jbeder I was thinking that RVO wouldn't work because several return statements can influence on it. Ok, I have checked it on clang compiler, everything is ok. I have reverted this part of changes and will research it on different compilers. It was wrong to add this changes to this PR.

jbeder (approved):

PR #1117, opened 2022-06-28
Basic update to CMakeLists.txt to address #1116

stephematician:
Changes are as per https://github.com/jbeder/yaml-cpp/issues/1116#issuecomment-1166975890

I also updated CONTRIBUTING.md as it took me a moment to find the correct test script (it is no longer called 'run_tests') - might help someone else.

PR #1119, opened 2022-07-07
CMake: Attach YAML_CPP_STATIC_DEFINE to yaml-cpp target

W4RH4WK:
I am using yaml-cpp via `add_subdirectory`. When compiling under Windows (with clang) I get the following warning, despite linking statically.

```
C:/Users/w4rh4wk/git/ph3lib/external/yaml-cpp/include\yaml-cpp/dll.h:22:17: warning: Defining YAML_CPP_API for DLL import [-W#pragma-messages]
#        pragma message( "Defining YAML_CPP_API for DLL import" )
```

The problem seems to be that the `YAML_CPP_STATIC_DEFINE` is not attached to the `yaml-cpp` target. Therefore, my code does not have this define set.

This commit attaches the define to the `yaml-cpp` target, such that targets linking against yaml-cpp (using `target_link_libraries`) also have it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,6 @@ if (YAML_BUILD_SHARED_LIBS)
 else()
   set(yaml-cpp-type STATIC)
   set(yaml-cpp-label-postfix "static")
-  add_definitions(-DYAML_CPP_STATIC_DEFINE)
 endif()
 
 set(build-shared $<BOOL:${YAML_BUILD_SHARED_LIBS}>)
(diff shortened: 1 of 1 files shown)


stephematician:
Any reason you made this a new PR? These changes are the same as #1117

W4RH4WK:
No, I missed #1117. Sorry.

stephematician:
All good - great to see we came to the same conclusion at least; I think it's the right approach to fix the problem. Here's hoping it gets accepted!

Issue #764, opened 2019-10-06
Declaration shadowing warnings from clang


asmaloney:
With the latest Apple clang version 11.0.0 (clang-1100.0.33.8), I'm getting three warnings about shadowed variables:

```
In file included from 3rdParty/include/yaml-cpp/yaml.h:12:
3rdParty/include/yaml-cpp/emitterstyle.h:12:25: warning: declaration shadows a variable in namespace 'YAML' [-Wshadow]
  enum value { Default, Block, Flow };
                        ^
3rdParty/include/yaml-cpp/emittermanip.h:52:3: note: previous declaration is here
  Block,
  ^
In file included from 3rdParty/include/yaml-cpp/yaml.h:12:
3rdParty/include/yaml-cpp/emitterstyle.h:12:32: warning: declaration shadows a variable in namespace 'YAML' [-Wshadow]
  enum value { Default, Block, Flow };
                               ^
3rdParty/include/yaml-cpp/emittermanip.h:51:3: note: previous declaration is here
  Flow,
  ^
In file included from 3rdParty/include/yaml-cpp/yaml.h:16:
In file included from 3rdParty/include/yaml-cpp/node/node.h:18:
3rdParty/include/yaml-cpp/node/type.h:12:27: warning: declaration shadows a variable in namespace 'YAML' [-Wshadow]
  enum value { Undefined, Null, Scalar, Sequence, Map };
                          ^
3rdParty/include/yaml-cpp/null.h:23:27: note: previous declaration is here
extern YAML_CPP_API _Null Null;
                          ^
3 warnings generated.
```

jbeder:
I'm not super concerned about that kind of shadowing. In any case, it's part of the public API and I don't think I want to change it for the sake of avoiding a shadowing problem that's unlikely to arise. Thanks for the report, though!

mateusz-plociennik:
Related: [Fix plenty of clang warnings `declaration shadows a variable`: using class enum instead of enum](https://github.com/jbeder/yaml-cpp/pull/815)

PR #1124, opened 2022-08-04
Minimize warnings when not the top-level project

### LINKED ISSUE #970: yaml-cpp/node/impl.h:206 error 'operator=' should return a reference to *this [-Werror=effc++]

BMBurstein:
Should fix #970 and #764 when trying to add yaml-cpp to other project


OlivierLDff:
You should use `YAML_CPP_MAIN_PROJECT `  for consistency.
https://github.com/jbeder/yaml-cpp/blob/c73ee34704c512ebe915b283645aefa9f424a22f/CMakeLists.txt#L12-L15

PR #1140, opened 2022-10-04
Include property nodes in the long key check

### LINKED ISSUE #1139: YAML output is invalid for nested maps with tags
When using the stream-based YAML output writing a map with a custom type tag containing a map with a custom type tag as the first key results in invalid output.

| | Version |
| ---------- | ----- |
| yaml-cpp | `0.7.0` and 4ae4cb7309c07b2d8623d7e4f01efa6321441366 |
| gcc | `12.1.0` |

This code is a (decently minimal) snippet that reproduces the bug
```{c++}
YAML::Emitter out;

out << YAML::LocalTag("outerMap");
out << YAML::BeginMap;
out << YAML::LocalTag("innerMap") << YAML::BeginMap << "key" << "value" << YAML::EndMap << "outer value";
out << YAML::EndMap;
```

## Expected Behaviour
The following output would be one possibility for correct output
```{yaml}
!outerMap
? !innerMap
  key: value
: outer value
```
## Observed Behaviour
The following YAML is written to the output stream:
```{yaml}
!outerMap
!innerMap
  key: value
: outer value
```
This is not valid YAML as per the 1.2 standard as implicit map keys (such as the inner map) are restricted to a single line ([YAML 1.2.2 Spec](https://yaml.org/spec/1.2.2/#rule-ns-s-implicit-yaml-key)), and fails to be parsed by any of the YAML parsers I've tried.
 
## Cause
There are existing checks that cause a key to be written as a `YAML::LongKey` if a sequence or mapping is used as the key. This check does not work as the first item written to the stream is not the key, but a local tag. 

 ## Workaround
 Adding the `YAML::LongKey` configuration item causes the maps to be written out as explicit keys, which are not subject to the same restrictions

zenno-leaky-bucket:
This fixes #1139

The existing check for map and sequence node has been expanded to also include property nodes:
- Use the long format for keys with properties
- Add new tests involving tagged map and sequence collections as keys
- Change tagged key tests to expect long keys

### DIFF
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -533,7 +533,8 @@ void Emitter::BlockMapPrepareNode(EmitterNodeType::value child) {
     if (m_pState->GetMapKeyFormat() == LongKey)
       m_pState->SetLongKey();
     if (child == EmitterNodeType::BlockSeq ||
-        child == EmitterNodeType::BlockMap)
+        child == EmitterNodeType::BlockMap ||
+        child == EmitterNodeType::Property)
       m_pState->SetLongKey();
 
     if (m_pState->CurGroupLongKey())
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -431,6 +431,38 @@ TEST_F(EmitterTest, BlockMapAsKey) {
   ExpectEmit("? key: value\n  next key: next value\n: total value");
 }
 
+TEST_F(EmitterTest, TaggedBlockMapAsKey) {
+  out << BeginMap;
+  out << Key;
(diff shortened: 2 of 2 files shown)


zenno-leaky-bucket:
This is quite a conservative approach, there are cases where having a property does not require the map key to be in the long-format. For example, the existing test-case was valid output without having a long key. Without the ability to look ahead or change previous output in the emitter stream these cases cannot be detected.

One possible alternative is to emit an error if any key is generated that extends over multiple-lines without using long-keys. This would leave existing output intact, but is not consistent with the current handling of block-map keys, and requires extra user input.

PR #1146, opened 2022-10-21
Support for building with exceptions disabled

### LINKED ISSUE #196: Please make throw optional (ie support build without exceptions)
```
What steps will reproduce the problem?
1. Include standard yaml-cpp headers from a c++ file and build with gcc using 
-fno-exeptions

What is the expected output? What do you see instead?
The build breaks with the following error:

In file included from /usr/include/yaml-cpp/node.h:132:0,
                 from /usr/include/yaml-cpp/yaml.h:13,
                 from /home/dev00/dev/cpp/PuzzleGame/DuckLib/DuckUtils/ObjectLoader.cpp:10:
/usr/include/yaml-cpp/nodeimpl.h: In instantiation of ‘const YAML::Node* 
YAML::Node::FindValue(const T&) const [with T = std::basic_string<char>]’:
/usr/include/yaml-cpp/nodeimpl.h:69:36:   required from here
/usr/include/yaml-cpp/nodeimpl.h:33:26: error: exception handling disabled, use 
-fexceptions to enable

I would really appreciate an #ifdef or equivalent to make sure "throw" 
statements can be left out. I know this can leave some errors undetected but 
it's a requirement to have exceptions disabled in my project. Having some error 
state to check or some return value wherever possible would be ideal.

What version of the product are you using? On what operating system?
yaml-cpp 0.3.0 from portage on Sabayon Linux 64-bit

Please provide any additional information below.
grepping for "throw" in the "include" directory yields:
./node/detail/impl.h:64:                    throw BadSubscript();
./node/detail/impl.h:92:                    throw BadSubscript();
./node/detail/impl.h:135:                    throw BadInsert();
./node/impl.h:106:                throw TypedBadConversion<T>();
./node/impl.h:111:            throw TypedBadConversion<T>();
./node/impl.h:122:                throw TypedBadConversion<std::string>();

```

Original issue reported on code.google.com by `de...@gmx.it` on 7 Apr 2013 at 5:59

### LINKED ISSUE #799: Question - Compiling without exception support
Hiya,

I'm interested in trying out this library for integration into a game engine. However I'm compiling for a platform where C++ exception support is disabled; any `throw` statements cause a fatal error when compiling for that platform. Hence I can't use this library as-is.

I did find #196, but seeing as that discussion is almost 7 years old, I was wondering if you have any plans to reconsider?

Cheers!

### LINKED ISSUE #930: Is it possiable to disable exception?
Is it possiable to disable exception?
In some environment, exception and RTTI is disabled, I found RTTI is only used by gtest, but there is no way to disable exception. I think it's better to provide a error handle instead of ```throw``` when exception is disabled.

twestenkarl:
There are currently three issues for supporting compilation without exceptions
resolves #196 (closed with no changes)
resolves #799 (open, no interaction)
resolves #930 (open, no interaction)

these issue mention the following use case:
using the library for a project where external constraints force it to build without exceptions (e.g. game development, embedded)

My use case is a C++ application that I compile with exceptions enabled for Linux/Windows but also with emscripten for WebAssembly where I disable exceptions.

1) A new macro `YAML_CPP_NORETURN` to annotate functions as not returning in `dll.h`
2) A new function `YAML_throw<ExceptionType>(args...)` in `exception.h`
   this function will throw an exception unless exceptions are disabled in the compiler,
   detected by checking the pre-defined macro `__cpp_exceptions`
   In this case the exception class will be instantiated, and the user-provided function
   `YAML::handle_exception(const char*)` will be called on the exception's `what()` method
3) if exceptions are disabled,and the library's user does not provide `YAML::handle_exception`,
   there will be a linker error
4) all other files have been changed automatedly by running the following sed commands
   `sed -i "s/throw \([A-Za-z]*\)(\(.*\))/YAML_throw<\1>(\2)/g"` throw statements for non-templated exceptions
   `sed -i "s/throw \(.*\)<\(.*\)>(/YAML_throw<\1<\2> >(/g"` throw statements for templated exceptions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(YAML_CPP_DISABLE_UNINSTALL  "Disable uninstallation of yaml-cpp" OFF)
 option(YAML_USE_SYSTEM_GTEST "Use system googletest if found" OFF)
 option(YAML_ENABLE_PIC "Use Position-Independent Code " ON)
 option(YAML_CPP_USE_STRICT_FLAGS "Uses strict compilation flags e.g.: -Wall" ${YAML_CPP_MAIN_PROJECT})
+option(YAML_CPP_DISABLE_EXCEPTIONS "Disable use of exceptions see YAML::handle_exception(_local)." OFF)
 
 cmake_dependent_option(YAML_CPP_BUILD_TESTS
   "Enable yaml-cpp tests" OFF
@@ -31,7 +32,7 @@ cmake_dependent_option(YAML_MSVC_SHARED_RT
   "CMAKE_SYSTEM_NAME MATCHES Windows" OFF)
 set(YAML_CPP_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/yaml-cpp"
   CACHE STRING "Path to install the CMake package to")
- 
+
 if (YAML_CPP_FORMAT_SOURCE)
     find_program(YAML_CPP_CLANG_FORMAT_EXE NAMES clang-format)
 endif()
@@ -121,6 +122,10 @@ target_sources(yaml-cpp
 
 if (NOT DEFINED CMAKE_DEBUG_POSTFIX)
   set(CMAKE_DEBUG_POSTFIX "d")
+endif()
+
+if (YAML_CPP_DISABLE_EXCEPTIONS)
+  target_compile_definitions(yaml-cpp PUBLIC -DYAML_CPP_DISABLE_EXCEPTIONS)
 endif()
 
 set_target_properties(yaml-cpp PROPERTIES
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -224,7 +224,7 @@ public:
     node["a"] = a;
     return node;
   }
-  
+
   int a;
 };
 
@@ -252,7 +252,7 @@ public:
 
 // Implementation of convert::{encode,decode} for all classes derived from or being A
 namespace YAML {
-  template<typename T> 
+  template<typename T>
   struct convert<T, typename std::enable_if<std::is_base_of<A, T>::value>::type> {
     static Node encode(const T &rhs) {
       Node node = rhs.emit();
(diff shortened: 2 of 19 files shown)


twestenkarl:
Notes: 
* this commit does not change the CI config to build with exceptions disabled (which it should add), because I am not familiar with the CI config
* I tried running clang-format on the source tree, but it would also change files that are unchanged by this commit

t123yh:
I noticed a few places where it still contains throw, which caused build errors for me:

https://github.com/twestenkarl/yaml-cpp/blob/5b1c606801529ed90ee7f3eaf6afc3e51d547562/include/yaml-cpp/depthguard.h#L54

https://github.com/twestenkarl/yaml-cpp/blob/5b1c606801529ed90ee7f3eaf6afc3e51d547562/src/parser.cpp#L83

t123yh:
I removed the remaining throws and built it successfully without exceptions.

https://github.com/t123yh/yaml-cpp/commit/12520bd8817077b1b025d810aa135045c73f4893

SGSSGene:
@jbeder what is your opinion about an "exception free" option for yaml-cpp

@twestenkarl This is a great PR. I do not have a use case for a exception free environment, but this of course doesn't mean other might not have it. If @jbeder  sees fit for this, there are two things that this PR still needs:
1. I think the mechanism should also be shielded by a cmake option, such that we can also use it if `__cpp_exceptions` evaluates to true.
2. We need to add the README or Tutorial (or something) that explains this feature.

jbeder:
I suppose it's reasonable, but I never felt like it was worth the readability cost - it seems like it's pretty invasive, since yaml-cpp depends on exceptions pretty heavily for the parsing side of things.

I'd check to see how relevant this still is in 2026 - compilers surely have improved, embedded systems are more powerful, and the last comment here was 2023.

SGSSGene:
> I'd check to see how relevant this still is in 2026 - compilers surely have improved, embedded systems are more powerful, and the last comment here was 2023.

I also have no clue, I just would like to close all related issues :-).

So I made a huge overhaul of the PR:
- Changed `YAML::Yaml_throw` to `YAML::raise` which makes the calls less noisy:
  ```
  throw SomException(param1, param2);       // original
  
  YAML_throw<SomException>(param1, param2); // previous suggestion
  
  raise<SomeException>(param1, param2);     // new suggestion
  ```
  Now the code for a `throw` is not much shorter than with `raise`.
- I had some worries about testing two version of yaml-cpp with and without exception.
To combat this, there is a `handle_exception` handle, which will be called if available. This is independent
of availability of exceptions or not making this mechanism always available. If someone wants to suppress exception throwing one must set cmake option `YAML_CPP_DISABLE_EXCEPTIONS`.
- Because of this new registration mechanism was required, for this 4 new functions are introduced:
  - `set_handle_exception`, `set_handle_exception_local`
  - `get_handle_exception`, `get_handle_exception_local`
  
  The difference between the local version is only for the current thread (via thread_local). The other handler
  is a global handler for the complete process. The local handler is always preferred over the global one.
  The set-methods are expecting a function pointer with the signature `void my_handler(const char* what)` or a nullptr to deactivate them.
- I also added some documentation to the Tutorial

Issue #833, opened 2020-03-12
Iterating a sequence goes beyond end


benstadin:
My application crashes in release mode (in debug mode all seems fine). I believe the reason is because the const_iterator for the sequence goes beyond end. I've put some test code to output the current item number. There are 8 items in the sequence, yet the iterator is at the ninth when the crash happens. 

Test data (notice there are 8 sequence items only):
```
admin:
    description: Super user with all permissions
    '/v1/providers':
    - 'CREATE_ANY'
    - 'READ_ANY'
    - "UPDATE_ANY"
    - 'DELETE_ANY'
    - 'CREATE_OWN'
    - 'READ_OWN'
    - 'UPDATE_OWN'
    - 'DELETE_OWN'
```

Code snippet:

for(auto roleIter = root.begin(); roleIter != root.end(); roleIter++) {
        // step 1. get role
        
        if (!roleIter->first.IsScalar() || !roleIter->first.IsDefined()) {
            return nullptr;
        }
        
        const YAML::Node &role = (*roleIter).second;
        
        if (!roleIter->second.IsMap()) {
            return nullptr;
        }

        for(auto roleUrlIter = role.begin(); roleUrlIter != role.end(); roleUrlIter++) {
            if (!roleUrlIter->second.IsDefined() || !roleUrlIter->second.IsSequence()) {
                return nullptr;
            }
          
            const YAML::Node &url = roleUrlIter->second;
            std::cout << "size " << url.size() << std::endl;
            int cnt = 0;
            for(YAML::const_iterator permIter = url.begin(); permIter != url.end(); ++permIter) {
                // step 3. get permissions associated to rule and url
                std::cout << "cnt " << cnt << std::endl;
                cnt++;
                
                std::string permName = permIter->as<std::string>(); // <-- crash here invalid memory
            }

The debug output is:
```
size 8
cnt 0
cnt 1
cnt 2
cnt 3
cnt 4
cnt 5
cnt 6
cnt 7
cnt 8 // <-- crashes here, should be 7 only
```

Teknologicus:
I'm having crashes with a map of a map and YAML::const_iterator's

File yaml-cpp-test.yml:
```
---
key0:
  key1: value1
  key2: value2

```
C++ code:
```
#include <yaml-cpp/yaml.h>

int main() {
  YAML::Node map = YAML::LoadFile("yaml-cpp-test.yml");

[...]

PR #1169, opened 2023-02-02
Test(Node): (const)iterator on sequence

MatthijsBurgh:
This should be a valid test for the issues described in https://github.com/jbeder/yaml-cpp/issues/833.

So these tests should fail. Otherwise these test are not valid or there is a different in how the tests are build versus how users build against this library.

### DIFF
--- a/test/node/node_test.cpp
+++ b/test/node/node_test.cpp
@@ -323,6 +323,38 @@ TEST(NodeTest, IteratorOnConstUndefinedNode) {
   }
   EXPECT_EQ(0, count);
 }
+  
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
The test don't seem to be run in GH actions. See #1170 for that one.

Though that one show the added tests don't fail, which means we haven't found a way yet to reproduce the issue of #833 in a clean environment.

MatthijsBurgh:
Although this test doesn't exposes the issue  of #833, this is still a good addition to the tests.

MatthijsBurgh:
Ping @jbeder

PR #1170, opened 2023-02-02
Make sure tests are run in CI

MatthijsBurgh:
Issue:
Tests are not found by ctest. Therefore all CMake CI jobs, don't run any tests at all. Both GH actions and appveyor.

Solution:
In the old situation the `cmake_dependent_option` for the `YAML_CPP_BUILD_TESTS` option, required `BUILD_TESTING` to be set `ON`. Though it was never set. It will be set `ON` when including `CTest`, but that was conditionally on `YAML_CPP_BUILD_TESTS`. So a deadlock situation was created.

So I have chosen to always include CTest, And always set `YAML_CPP_BUILD_TESTS` to `OFF`. (Therefore it is possible to replace `cmake_dependent_option` by a normal `option`)

To build tests, `YAML_CPP_BUILD_TESTS` needs to be set to `ON` by CLI argument.


MatthijsBurgh:
Ping @jbeder

jbeder:
Can you split off the new tests into a different PR and focus this on whatever config changes you're making?

Also, can you explain (1) the problem and (2) why this is the right solution?

MatthijsBurgh:
@jbeder I have update the main comment.

I don't have a solution yet for the appveyor tests. It tries to look for `C:/projects/yaml-cpp/build/test/Debug/yaml-cpp-tests.exe`, but can't find it.
The GH action jobs look for the test at `D:\a\yaml-cpp\yaml-cpp\build\test\yaml-cpp-tests.exe` and do find it. Ofcourse the paths are a bit different at the beginning. But the appveyor jobs looks for `build/test/Debug/yaml-cpp-tests.exe` while the GH actions jobs look for `build/test/yaml-cpp-tests.exe` without the extra `Debug` folder. So we need to figure out how to change the build path or the location it searches.

jbeder (approved):

jbeder:
This is wild. Thanks for catching it. I don't think we need the appveyor ones, since GH actions added Windows to their options. Can you remove the change to appveyor in this PR, and then submit another PR to delete the appveyor stuff?

MatthijsBurgh:
Done and done (https://github.com/jbeder/yaml-cpp/pull/1176)

jbeder (approved):

PR #1171, opened 2023-02-07
Don't install embedded copy of GTest

### LINKED ISSUE #488: Embedding of gtest/gmock in source tree
The current approach used by yaml-cpp is to embed gtest/gmock in `test/gmock-1.7.0`.  While this works fine for the most part, it doesn't work well when you're integrating it into a larger build which also includes its own gtest, e.g. a later version such as 1.8.0.  It would be nice if this could be made optional.  The headers and libraries can conflict.

As a suggestion, this is the approach I took in a different project: [GTest.cmake](https://github.com/ome/ome-files-cpp/blob/v0.2.0/cmake/GTest.cmake).  Here, we pass in a variable `GTEST_SOURCE` which points to the location of the sources.  If unset, we simply use `find_package` and use an external version.  If set, we use the internal version (or the version pointed to outside the source tree).  An option could be added to yaml-cpp to do an equivalent action, e.g. `build-gtest` or `build-gmock`.  It could even be defaulted based upon the result of `find_package`.

Also worth noting: gtest 1.8 combines gmock and no longer enforces the requirement to "vendor" into a source tree.  It's now locally installable again, with the proviso that it needs building with a compatible set of compiler options (so no different than most other libraries!).

Regards,
Roger

PhilMiller:
Google Test itself documents the `INSTALL_GTEST` option as something that projects embedding it should set to `OFF`. Leaving it on means that projects downstream from libraries that embed it are likely to encounter conflicting copies. This is particularly annoying because GTest does not maintain anything like a stable API, and so causes builds to fail if include paths pick up an inappropriately installed copy ahead of the one that the code wanted.

Fixes #488

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -2,6 +2,7 @@ find_package(Threads REQUIRED)
 
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 set(BUILD_MOCK ON CACHE BOOL "" FORCE)
+set(INSTALL_GTEST OFF CACHE BOOL "" FORCE)
 set(CMAKE_POLICY_DEFAULT_CMP0048 NEW)
 
 add_subdirectory(


nigels-com:
Looks good.

ahogen:
Related, but maybe out of scope for this MR -- this project's documentation generator is recursing into the GoogleTest directories and publishing its docs. Things like GoogleTest's own "how to contribute" type docs are being shown in this project. Took me a sec to realize what was going on. Rather confusing. 

https://codedocs.xyz/jbeder/yaml-cpp/pages.html

jbeder (approved):

Issue #1180, opened 2023-03-21
About CTest in CMakeLists.txt


skjsnb:
Hi 👋

I have read the `CMakeLists.txt` file in the root directory of the yam-cpp repository, and I noticed that `CTest` is globally referenced. Is there any reason for doing so? Would it be better to put it inside `if(YAML_CPP_BUILD_TESTS)`?

```
if(YAML_CPP_BUILD_TESTS)
    include(CTest)
    add_subdirectory(test)
endif()
```

Thank you for your attention.

jbeder:
Sure! I basically accept any PRs about CMake.

MatthijsBurgh:
Yes there is a reason, see #1170

PR #1181, opened 2023-03-21
Move CTest to the test entry

skjsnb:
If `CTest` is a global reference, then `YAML_CPP_BUILD_TESTS` will be called and the compilation product will be generated regardless of whether `YAML_CPP_BUILD_TESTS` is enabled or not. So it makes more sense to put `include(CTest)` inside the `if (YAML_CPP_BUILD_TESTS)` statement.

```cmake
if(YAML_CPP_BUILD_TESTS)
  include(CTest)
  add_subdirectory(test)
endif()

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,6 @@ include(CMakePackageConfigHelpers)
 include(CMakeDependentOption)
 include(CheckCXXCompilerFlag)
 include(GNUInstallDirs)
-include(CTest)
 
 option(YAML_CPP_BUILD_CONTRIB "Enable yaml-cpp contrib in library" ON)
 option(YAML_CPP_BUILD_TOOLS "Enable parse tools" ON)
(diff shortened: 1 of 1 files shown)


jbeder:
Fixes #1180

MatthijsBurgh:
This one needs to be reverted.

I added changed this, because no tests were run in CI. (#1170)

This PR reverts that all again. See https://github.com/jbeder/yaml-cpp/actions/runs/4480153307/jobs/7877729295?pr=1181#step:6:17 `No tests were found!!!`

So I suggest to revert this PR. And after it, @skjsnb can think about a different solution.

I don't see why the solution before this PR was an issue to any one. As it checks for `YAML_CPP_BUILD_TESTS` before including the test sub dir.

skjsnb:
@MatthijsBurgh  I'm sorry for causing trouble to you. I indeed didn't think about the part concerning the CI process, so I agree with your opinion.

MatthijsBurgh:
@jbeder thanks for the quick response

PR #1207, opened 2023-07-29
Add whitespace in tag match regex

Inori:
I'm not sure if this fix matches YAML standard, but sometimes these is whitespace in tag suffix, for example, it is widely used in Unity Engine's meta file, like this:
```
%YAML 1.1
%TAG !u! tag:unity3d.com,2011:
--- !u!1 &29818312667802682
GameObject:
  m_ObjectHideFlags: 0
  m_CorrespondingSourceObject: {fileID: 0}
  m_PrefabInstance: {fileID: 0}
  m_PrefabAsset: {fileID: 0}
  serializedVersion: 6
```
Without the patch, we'll only get `unity3d.com,2011:1` for the tag, but what I need is `unity3d.com,2011:1 &29818312667802682`.
Or are there some better ways to achieve this?

### DIFF
--- a/src/exp.h
+++ b/src/exp.h
@@ -139,7 +139,7 @@ inline const RegEx& URI() {
   return e;
 }
 inline const RegEx& Tag() {
-  static const RegEx e = Word() | RegEx("#;/?:@&=+$_.~*'()", REGEX_OR) |
+  static const RegEx e = Word() | RegEx(" #;/?:@&=+$_.~*'()", REGEX_OR) |
                          (RegEx('%') + Hex() + Hex());
   return e;
 }


jbeder:
This is likely not what you're looking for - the `&` represents an anchor (that can be referred to later by an alias with `*`), so this should be parsed as (1) a tag and (2) an anchor for the entire document. This is what yaml-cpp does; you can see this easily by adding an alias at the end:

```
%YAML 1.1
%TAG !u! tag:unity3d.com,2011:
--- !u!1 &29818312667802682
GameObject:
  m_ObjectHideFlags: 0
  m_CorrespondingSourceObject: {fileID: 0}
  m_PrefabInstance: {fileID: 0}
  m_PrefabAsset: {fileID: 0}
  serializedVersion: 6
foo: *29818312667802682
```

and running this through `util/parse`:

```
$ ./util/parse test.yaml
!<tag:unity3d.com,2011:1> &1
GameObject:
  m_ObjectHideFlags: 0
  m_CorrespondingSourceObject: {fileID: 0}
  m_PrefabInstance: {fileID: 0}
  m_PrefabAsset: {fileID: 0}
  serializedVersion: 6
foo: *1
```

(Note that yaml-cpp renames the anchor to `1`, which is its standard behavior.)

PR #1211, opened 2023-08-16
Specify CMake policy range to avoid deprecation warning

craigscott-crascit:
CMake 3.27 started issuing a deprecation warning for any
cmake_minimum_required() call that specified a minimum
version older than 3.5. Specifying a version range instead of
a simple minimum version avoids that warning without
raising the minimum supported CMake version. The NEW
policy behavior will be used for all policies introduced up to
CMake 3.14 with this change.

I checked the effects of the policies between CMake 3.4 and 3.14,
and I believe the NEW behavior of all those policies should be fine
for the CMakeLists.txt files in this project. I didn't go any higher
because policy CMP0092 added by CMake 3.15 would need further
checking (that policy affects the default compiler warning flags).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
-# 3.5 is actually available almost everywhere, but this a good minimum
-cmake_minimum_required(VERSION 3.4)
+# 3.5 is actually available almost everywhere, but this a good minimum.
+# 3.14 as the upper policy limit avoids CMake deprecation warnings.
+cmake_minimum_required(VERSION 3.4...3.14)
 
 # enable MSVC_RUNTIME_LIBRARY target property
 # see https://cmake.org/cmake/help/latest/policy/CMP0091.html


jbeder:
What does it mean for the minimum to be a range of versions?

craigscott-crascit:
The `cmake_minimum_required()` command actually has two effects:

* It specifies what is the minimum CMake version that can be used. When a version range A...B is given, the bottom of that range (A) is the minimum CMake version.
* The other effect of the command is to set CMake's policies. These policies are how CMake preserves backward compatibility, while still allowing changes in behavior to be introduced in a controlled way with later versions. When a version range A...B is given to `cmake_minimum_required()`, all CMake policies introduced in CMake versions up to the lesser of the top of that range (B) and the actual CMake version being used will be set to NEW.

It is the second of those two points that is the reason for specifying a version range rather than a single version number. By giving a version range A...B, you are saying "You must use at least CMake version A, but NEW policy settings can be used for up to CMake version B". Let's take the example of the change added in this PR:

```cmake
cmake_minimum_required(VERSION 3.4...3.14)
```

What this means is the CMake version must be at least 3.4 (same as before this PR). If I use CMake 3.14 or higher, all CMake policies up to `CMP0088` will be set to NEW (see the [CMake policy documentation](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html) for the list of policies and what they do). If I use CMake 3.4, only policies up to `CMP0065` will be set to NEW. If I use CMake 3.10, only policies up to `CMP0071` would be set to NEW, and so on.

The upper end of the version range is also what CMake looks at when it is checking whether to emit a deprecation message for a project asking for behavior that is too old. CMake 3.27 started issuing such a warning if the project required behavior for CMake 3.4 or earlier. The project is still allowed to set a low _minimum_ CMake version. The deprecation message is tied to the _top_ of the specified version range because that is what limits CMake's _behavior_, and it is that behavior that the deprecation message is about, not the minimum CMake version.

jbeder (approved):

jbeder:
Thanks for the explanation!

PR #1215, opened 2023-08-18
CMake: Allow to disable uninstall

### LINKED ISSUE #1151: CMake uninstall target
The uninstall target in the project is always shown even if you don't have the install target. It should really be gated behind the `${YAML_CPP_INSTALL}`. As a workaround in my projects I have been adding a dummy target before fetching yaml-cpp to prevent this:
```
# Get rid of yaml uninstall target
add_library(uninstall INTERFACE)

set(YAML_CPP_BUILD_TESTS OFF CACHE BOOL "")
set(YAML_CPP_BUILD_TOOLS OFF CACHE BOOL "")
set(YAML_CPP_BUILD_CONTRIB OFF CACHE BOOL "")
set(YAML_BUILD_SHARED_LIBS OFF CACHE BOOL "")
set(YAML_CPP_INSTALL OFF CACHE BOOL "")
set(YAML_CPP_FORMAT_SOURCE OFF CACHE BOOL "")
FetchContent_Declare(
    yaml-cpp
    GIT_REPOSITORY      https://github.com/jbeder/yaml-cpp
    GIT_TAG             1b50109f7bea60bd382d8ea7befce3d2bd67da5f
)

FetchContent_MakeAvailable(yaml-cpp)
```

Megamouse:
`add_custom_target(uninstall ...` causes a clash with the same target in rtmidi (https://github.com/thestk/rtmidi). "add_custom_target cannot create target "uninstall" because another target with the same name already exists.  The existing target is a custom target created in source directory"

So it's probably easiest to simply allow to disable the uninstall in the first place.

Also gates the uninstall target unless `YAML_CPP_INSTALL` is set, which fixes #1151.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,7 +199,7 @@ if (YAML_CPP_FORMAT_SOURCE AND YAML_CPP_CLANG_FORMAT_EXE)
 endif()
 
 # uninstall target
-if(NOT YAML_CPP_DISABLE_UNINSTALL AND NOT TARGET uninstall)
+if(YAML_CPP_INSTALL AND NOT YAML_CPP_DISABLE_UNINSTALL AND NOT TARGET uninstall)
   configure_file(
     "${CMAKE_CURRENT_SOURCE_DIR}/cmake_uninstall.cmake.in"
     "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,8 @@ option(YAML_CPP_BUILD_TOOLS "Enable parse tools" ON)
 option(YAML_BUILD_SHARED_LIBS "Build yaml-cpp shared library" ${BUILD_SHARED_LIBS})
 option(YAML_CPP_INSTALL "Enable generation of yaml-cpp install targets" ${YAML_CPP_MAIN_PROJECT})
 option(YAML_CPP_FORMAT_SOURCE "Format source" ON)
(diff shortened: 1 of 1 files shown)


jbeder:
That strikes me as really weird; why can't every project have an uninstall? Are you sure it's not misconfigured somehow? (Note I'm not a CMake expert so maybe it's best to ask the CMake people.)

HunterBelanger:
I have just run into the problem that was mentioned by @Megamouse above and in #1151 . My particular case comes from trying to use FetchContent to get both yaml-cpp and Eigen in the same project. Both define an uninstall target, and this then leads to a CMake error. The only way to avoid this problem is to disable the creation of the uninstall target.

I would prefer the approach @duncanspumpkin recommended, shielding the uninstall target creation with the YAML_CPP_INSTALL option. You should certainly only make the uninstall target if you also have an install target. That being said, there might still be cases where you want an install target, but no uninstall target. I think the option @Megamouse added to disable the uninstall target should be kept, but a check should also be added to ensure that YAML_CPP_INSTALL is also set to ON.

CMakeLists.txt
```diff
@@ -197,7 +199,7 @@ if (YAML_CPP_FORMAT_SOURCE AND YAML_CPP_CLANG_FORMAT_EXE)
 endif()
 
 # uninstall target
-if(NOT TARGET uninstall)
+if(NOT YAML_CPP_DISABLE_UNINSTALL AND NOT TARGET uninstall)
```
HunterBelanger, line 202:
```suggestion
if(YAML_CPP_INSTALL AND NOT YAML_CPP_DISABLE_UNINSTALL AND NOT TARGET uninstall)
```

Megamouse:
I applied the suggested change as well.

jbeder (approved):

PR #1225, opened 2023-09-09
Revert "node/convert: relax the check for string_view (#1222)"

### LINKED ISSUE #1223: Newest version doesn't build anymore
Today I am getting the following error when running make.
```
[  2%] Building CXX object CMakeFiles/yaml-cpp.dir/src/convert.cpp.o
In file included from /home/git/actions-runner/_work/dep/yaml-cpp/src/convert.cpp:3:
/home/git/actions-runner/_work/dep/yaml-cpp/include/yaml-cpp/node/convert.h:20:10: fatal error: version: No such file or directory
 #include <version>
          ^~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/yaml-cpp.dir/build.make:102: CMakeFiles/yaml-cpp.dir/src/convert.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:912: CMakeFiles/yaml-cpp.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```
I believe this is due to commit 6262201182a0ee357d8a456e6d1a8e5984e4c6dd

tchaikov:
This reverts commit 6262201182a0ee357d8a456e6d1a8e5984e4c6dd.

in 62622011, we wanted address the needs to use the `string_view` converter in C++98, but that requirement was based on wrong preconditions. `std::string_view` was introduced in C++17, and popular standard libraries like libstdc++ and libc++ both provide `std::string_view` when the source is built with C++17.

furthermore 62622011 is buggy. because it uses `<version>` to tell the feature set provided by the standard library. but `<version>` is a part of C++20. so this defeats the purpose of the change of 62622011.

Fixes #1223

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -17,9 +17,8 @@
 #include <type_traits>
 #include <valarray>
 #include <vector>
-#include <version>
 
-#ifdef __cpp_lib_string_view
+#if __cplusplus >= 201703L
 #include <string_view>
 #endif
 
@@ -94,7 +93,7 @@ struct convert<char[N]> {
   static Node encode(const char* rhs) { return Node(rhs); }
 };
 
-#ifdef __cpp_lib_string_view
+#if __cplusplus >= 201703L
 template <>
 struct convert<std::string_view> {
   static Node encode(std::string_view rhs) { return Node(std::string(rhs)); }


include/yaml-cpp/node/convert.h
```diff
@@ -17,9 +17,8 @@
 #include <type_traits>
 #include <valarray>
 #include <vector>
-#include <version>
 
-#ifdef __cpp_lib_string_view
+#if __cplusplus >= 201703L
```
0xFireWolf, line 21:
```suggestion
#if ((defined(_MSVC_LANG) && _MSVC_LANG >= 201703L) || __cplusplus >= 201703L)
```

include/yaml-cpp/node/convert.h
```diff
@@ -94,7 +93,7 @@ struct convert<char[N]> {
   static Node encode(const char* rhs) { return Node(rhs); }
 };
 
-#ifdef __cpp_lib_string_view
+#if __cplusplus >= 201703L
```
0xFireWolf, line 96:
```suggestion
#if ((defined(_MSVC_LANG) && _MSVC_LANG >= 201703L) || __cplusplus >= 201703L)
```

0xFireWolf (changes requested):

include/yaml-cpp/node/convert.h
tchaikov: after reading https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170, i still don't understand why your proposal is better. what exactly problem are you trying to address?

jbeder: Agreed. @0xFireWolf I'm going to merge this PR as-is, and if you want to suggest a change, please open a new PR with an explanation.

jbeder (approved):

include/yaml-cpp/node/convert.h
jbeder: Sorry, I didn't see the long conversation on #1222. I finally read that, and I understand this suggestion, @0xFireWolf, and I'd accept this as a PR with a link to the explanation in #1222.

PR #1232, opened 2023-10-12
Need to set CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY to lib folder in CMakeLists.txt

slambon:
There's currently a warning message when using YAML-CPP as static library in debug mode because the library yaml-cppd.lib is expecting / missing pdb information. By setting the pdb generation folder to the lib folder, the compiler is no longer complaining.

It now looks like this:
```
.\RootBuildFolder\Release
- yaml-cpp.lib

.\RootBuildFolder\Debug
- yaml-cpp static.pdb
- yaml-cppd.lib
```

See more about the problem:
https://stackoverflow.com/questions/25843883/how-to-remove-warning-lnk4099-pdb-lib-pdb-was-not-found

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,8 @@ else()
   set(yaml-cpp-label-postfix "static")
 endif()
 
+set(CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY "${CMAKE_INSTALL_LIBDIR}/../")
+
 set(build-shared $<BOOL:${YAML_BUILD_SHARED_LIBS}>)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -47,6 +47,8 @@ else()
   set(yaml-cpp-label-postfix "static")
 endif()
 
+set(CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY "${CMAKE_INSTALL_LIBDIR}/../")
```
jbeder, line 50: It strikes me as a little weird that you're using a parent directory here. Is there a more direct way to put this file in the right spot?

jbeder (commented):
Is there a way to test this? For example, how would you make sure the CI flags this if there's a regression?

PR #1240, opened 2023-10-30
CMake: hide all symbols by default in shared lib

SpaceIm:
Hide private symbols irrespective of compiler when yaml-cpp shared is built with CMake.
This PR is similar to https://github.com/jbeder/yaml-cpp/pull/984, but it does not remove `dll.h` (which was the reason why it has been reverted in https://github.com/jbeder/yaml-cpp/pull/1038). So it doesn't break users not relying on CMake to build yaml-cpp.

Also simplify `dll.h`, which had lot of dead code, and weird logic for entering in delcspec branch (actually it's windows specific, so `_WIN32` is fine).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,12 @@ set_property(TARGET yaml-cpp
   PROPERTY
     CXX_STANDARD_REQUIRED ON)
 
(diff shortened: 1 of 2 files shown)


nigels-com:
Has conflicts now.

PR #1243, opened 2023-11-08
Adding support for handling YAML Merge Key (#41)

nlescoua:
Support for YAML Merge keys ( <<: [*dict1, *dict2] ) is added. The merge key is a specific scalar with value << (and tag !!merge) that implies that during node construction, the map (or sequence of maps) are merged into the current map. The priority rules are that each key from maps within the value associated with  << are added iff the key is not yet present in the current map (and first map gets higher priority). Test cases have been added accordingly.


bradnelson:
I tested this PR and it appears to work as expected for my project.  Thanks for the contribution, @nlescoua !

src/nodebuilder.cpp
```diff
@@ -69,11 +70,29 @@ void NodeBuilder::OnMapStart(const Mark& mark, const std::string& tag,
...
   node.set_style(style);
   m_mapDepth++;
+  m_mergeDicts.emplace_back();
+}
+
+void MergeMapCollection(detail::node& map_to, detail::node& map_from,
+                        detail::shared_memory_holder& pMemory) {
+  const detail::node& const_map_to = map_to;
+  for (auto j = map_from.begin(); j != map_from.end(); j++) {
+    detail::node* s = const_map_to.get(*j->first, pMemory);
```
azat, line 80: This compares the `shared_ptr` of the keys, so it will not work correctly, this can be visible by iterating over keys, here is a fix for this case - https://github.com/jbeder/yaml-cpp/pull/1279

It is based on your PR, with this fix and a test on top.

test/integration/load_node_test.cpp
```diff
@@ -173,6 +173,47 @@ TEST(LoadNodeTest, CloneAlias) {
...
+      "&stuff { << : *foo, b : 3}, w: { << : [*stuff, *bar], c: 4 }, v: { '<<' "
+      ": *foo } , u : {!!merge << : *bar}, t: {!!merge << : *bar, h: 3} }");
+  EXPECT_EQ(NodeType::Map, node["z"].Type());
+  EXPECT_EQ(NodeType::Map, node["w"].Type());
+  EXPECT_FALSE(node["z"]["<<"]);
+  EXPECT_EQ(1, node["z"]["a"].as<int>());
+  EXPECT_EQ(3, node["z"]["b"].as<int>());
+  EXPECT_EQ(1, node["z"]["c"].as<int>());
+
+  EXPECT_EQ(1, node["w"]["a"].as<int>());
```
azat, line 199: I've also changed the order of merging, to make it more compatible, see - [`2a1aada` (#1279)](https://github.com/jbeder/yaml-cpp/pull/1279/commits/2a1aada07351e8bca2b3f77d5eb0b9ed25228639)

nlescoua:
Superseeded by #1279 -- Thanks @azat

Issue #1080, opened 2022-01-11
Segfault with _GLIBCXX_DEBUG defined


joejustesen:
You can also cause a segfault by defining D_GLIBCXX_DEBUG when building the unit tests and running them.

alex-thiessen-for-siemens:
Could it be because

> you can only link code compiled with debug mode and code compiled without debug mode if no instantiation of a container is passed between the two translation units.

, yet `yaml-cpp` **mixes** header-implemented code (affected by the `_GLIBCXX_DEBUG` flag on user side) with library (linked) code (unaffected by the `_GLIBCXX_DEBUG` flag on user side)?

https://gcc.gnu.org/onlinedocs/libstdc++/manual/debug_mode_using.html#debug_mode.using.mode

If that the case, then it's not a bug.

alex-thiessen-for-siemens:
To my knowledge, segfaults you observed are due to build issues in versions ≤0.6.3.

From now on, `yaml-cpp` is proven to be `_GLIBCXX_DEBUG`-compatible (even with pedantic checks for implementation-defined behavior) as part of PR checks.

PR #1247, opened 2023-11-26
workflows/build.yml: Use debug std C++ library on Ubuntu

rico-chet:
Protect from regressions due to use of undefined or
implementation-specific behavior when using `std::` containers and smart
pointers.

This only has effect on platforms with the GNU standard C++ library.

Refer to https://gcc.gnu.org/onlinedocs/libstdc++/manual/debug_mode.html.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,7 @@ jobs:
           ${{format(matrix.generator != 'Default Generator' && '-G "{0}"' || '', matrix.generator)}}
         CMAKE_INSTALL_PREFIX: "${{ github.workspace }}/install-prefix"
         CMAKE_BUILD_TYPE: Debug
+        CMAKE_CXX_FLAGS_DEBUG: ${{ matrix.googletest == 'build' && '-g -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC' || '-g' }} 
       runs-on: ${{ matrix.os }}
       steps:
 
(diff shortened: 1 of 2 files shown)


rico-chet:
should prove that #1080 is **not** an issue

.github/workflows/build.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           ${{format(matrix.generator != 'Default Generator' && '-G "{0}"' || '', matrix.generator)}}
         CMAKE_INSTALL_PREFIX: "${{ github.workspace }}/install-prefix"
         CMAKE_BUILD_TYPE: Debug
+        CMAKE_CXX_FLAGS_DEBUG: ${{ matrix.googletest == 'build' && '-g -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC' || '-g' }}
```
jbeder, line 40: Why did you put the googletest == build clause here? Does the system googletest fail with these flags?

rico-chet: Yes, 'system' failed to build, just 'build' worked.

I didn't dig deeper, but it's a common effect in the _debug_ mode of the GNU standard C++ library that it can **only**  be mixed with non-_debug_ binaries when those do **not** exchange data with the _debug_-generated code in e.g. ˋstd::ˋ containers or smart pointers. I thus assumed that the system-provided _googletest_ (which is obviously non-_debug_) doesn't mix with the _debug_ structures, by design. So I excluded these variants from being built in the _debug_ mode.

jbeder: OK, sounds good. Can you add a comment explaining this?

jbeder (approved):

PR #1256, opened 2023-12-19
Analyze with cppcheck

alex-thiessen-for-siemens:


jbeder:
Can you add something to the GitHub actions so I can see what behavior this is supposed to have?

alex-thiessen-for-siemens:
Hi Jesse, I opened this PR accidentally. I intended to make this work in my fork first, which I haven't succeeded yet.

All in all, observable difference should be that `cppcheck` is run in the build step on latest Ubuntu.

Locally, I had a false positive getting triggered by a piece of code which is slightly obscure (https://github.com/jbeder/yaml-cpp/blob/c7639e81d5f00a5b47a2c9bd668f10d74b949071/src/scanner.cpp#L311C27-L311C27). This might be worth refactoring.

PR #1279, opened 2024-05-03
Adding support for handling YAML Merge Key (#41) - proper keys comparison for merging

azat:
**NOTE: this is based on https://github.com/jbeder/yaml-cpp/pull/1243, but includes a fix for proper merging (by comparing the key itself, instead of shared_ptrs, see the last patch).**

Support for YAML Merge keys ( <<: [*dict1, *dict2] ) is added. The merge
key is a specific scalar with value << (and tag !!merge) that implies
that during node construction, the map (or sequence of maps) are merged
into the current map. The priority rules are that each key from maps
within the value associated with  << are added iff the key is not yet
present in the current map (and first map gets higher priority). Test
cases have been added accordingly.


src/nodebuilder.cpp
```diff
@@ -69,11 +71,38 @@ void NodeBuilder::OnMapStart(const Mark& mark, const std::string& tag,
   node.set_tag(tag);
   node.set_style(style);
   m_mapDepth++;
+  m_mergeDicts.emplace_back();
+}
+
+void MergeMapCollection(detail::node& map_to, detail::node& map_from,
+                        detail::shared_memory_holder& pMemory) {
```
SGSSGene, line 78:
```suggestion
static void MergeMapCollection(detail::node& map_to, detail::node&& map_from,
                        detail::shared_memory_holder& pMemory) {
```
This function should be `static` or in an anonymous namespace to not leak any symbols to other translation units.

I also believe this is also a good case for a r-value reference for `map_from`. So I would use a `&&` to make clear that `map_from` will be in an unclear state after this call.

src/nodebuilder.cpp
```diff
@@ -69,11 +71,38 @@ void NodeBuilder::OnMapStart(const Mark& mark, const std::string& tag,
...
+  }
 }
 
 void NodeBuilder::OnMapEnd() {
   assert(m_mapDepth > 0);
+  detail::node& collection = *m_stack.back();
+  auto& toMerge = *m_mergeDicts.rbegin();
+  /// The elements for merging should be traversed in reverse order to prefer last values.
+  for (auto it = toMerge.rbegin(); it != toMerge.rend(); ++it) {
+    MergeMapCollection(collection, **it, m_pMemory);
```
SGSSGene, line 102: If `MergeMapCollection` is being adjusted as suggested above `**it` must be changed to `std::move(**it)`

SGSSGene (changes requested):
Requested Change:
- `MergeMapCollection` must be `static` or in an anonymous namespace.

I think, the R-Value (&&)/std::move suggestion is a matter of taste. This library is not very strong on move semantics anyways.

src/nodebuilder.cpp
azat: >This function should be static or in an anonymous namespace to not leak any symbols to other translation units.

Done

>I also believe this is also a good case for a r-value reference for map_from. So I would use a && to make clear that map_from will be in an unclear state after this call.

I'm not sure about this one, since `map_from` is in valid state upon return from `MergeMapCollection`, what make you thinks the opposite?

SGSSGene: I don't know what I was thinking last month....It seems fine way it is :sweat_smile:

src/nodebuilder.cpp
```diff
@@ -107,15 +136,40 @@ void NodeBuilder::Pop() {
...
     PushedKey& key = m_keys.back();
     if (key.second) {
-      collection.insert(*key.first, node, m_pMemory);
-      m_keys.pop_back();
+      detail::node& nk = *key.first;
+      if (nk.type() == NodeType::Scalar &&
+          ((nk.tag() == "tag:yaml.org,2002:merge" && nk.scalar() == "<<") ||
+           (nk.tag() == "?" && nk.scalar() == "<<"))) {
+        if (node.type() == NodeType::Map) {
+          m_mergeDicts.rbegin()->emplace_back(&node);
```
SGSSGene: I think the following is easier to read:
```suggestion
          m_mergeDicts.back().emplace_back(&node);
```

PR #1283, opened 2024-05-27
Add missing iostream header

rurban:


jbeder:
Why?

rurban:
Try with gcc 14

Jesse Beder ***@***.***> schrieb am Mo., 27. Mai 2024, 16:33:

jbeder:
It's customary to do things like post the exact error message.

rurban:
Ok, this has to be reworked. 
It fails on my ubuntu 24.04 clang-tidy with missing iostream:
```
/usr/local/include/yaml-cpp/parser.h:10:10: error: 'iostream' file not found [clang-diagnostic-error]
   10 | #include <iostream>
```

but my fedora core 40 with gcc 14 warns about a missing iostream header. will check there

PR #1297, opened 2024-07-17
fix: use C locale by default

### LINKED ISSUE #1282: Numbers are emitted wrongly under non-C locales.
From a [question on StackOverflow](https://stackoverflow.com/questions/78504953):
```cpp
int main(){
  std::locale::global(std::locale(""));
  YAML::Emitter out;
  out << 2112;
  std::cout << "Emitter for 2112 = " << out.c_str() << '\n';
}
```

When run under `LC_ALL=en_US.UTF-8` this produces `2,112` which parses as a string `"2,112"`.
Under `LC_ALL=nl_BE.UTF-8` this becomes `2.112`, which is interpreted by other parsers as a floating point number.

SGSSGene:
fix #1282

This explicitly sets locale 'C' to all `std::stringstream` objects that output YAML.

### DIFF
--- a/include/yaml-cpp/emitter.h
+++ b/include/yaml-cpp/emitter.h
@@ -141,6 +141,7 @@ inline Emitter& Emitter::WriteIntegralType(T value) {
   PrepareNode(EmitterNodeType::Scalar);
 
   std::stringstream stream;
+  stream.imbue(std::locale("C"));
   PrepareIntegralStream(stream);
   stream << value;
   m_stream << stream.str();
@@ -158,6 +159,7 @@ inline Emitter& Emitter::WriteStreamable(T value) {
   PrepareNode(EmitterNodeType::Scalar);
 
   std::stringstream stream;
+  stream.imbue(std::locale("C"));
   SetStreamablePrecision<T>(stream);
 
   bool special = false;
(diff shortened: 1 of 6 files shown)


jbeder:
In terms of verification: the issue #1282 talks about "running under" `LC_ALL=en_US.UTF-8` and similar. How does the test you wrote here relate to that? Is that an environment variable? (Can that be set/unset in the test?)

SGSSGene:
> In terms of verification: the issue #1282 talks about "running under" `LC_ALL=en_US.UTF-8` and similar. How does the test you wrote here relate to that? Is that an environment variable? (Can that be set/unset in the test?)

So, by default the locale() is set to"'"' at startup, meaning LC_ALL has no influence. But in there case someone is setting `std::locale::global(std::locale(""))` which changes the global setting to follow LC_ALL/LC_NUMERIC/etc. Instead of setting it to `std::locale("")` which reads from LC_ALL we can set it to a value of our wishes.

A locale set to `en_US.UTF-8" would print `2014` as `2,014`
Setting the same to `de_DE.UTF-8" would print `2014" as `2.014`. Both things we really do not want.

So instead I overwrite locally for every std::stringstream it explicitly to "C", which I believe is the behavior we want.

jbeder (approved):

PR #1305, opened 2024-08-05
Fix compile warning with -Wshadow and unused-includes

RonsonNamek:


RonsonNamek:
image
image
image

this commit fix these warning.

PR #1312, opened 2024-08-22
fix: parse files with '\r' symbols as line ending correctly

### LINKED ISSUE #1309: Carriage return is not considered as a line break character
When using the following code snippet:
 
```
YAML::Node config;
config = YAML::Load(string_stream_from_yaml_file);
size_t size = config["channels"].size();
```
with `my_file.yaml` having this content:

```
yaml_content = (
    "channels:\r"
    "- my_channel\r"
    "- my_second_channel\r"
)
```
`size` is equal to 1, while it's equal to 2 when the `yaml_content` doesn't have the `\r` character and only `<LF>`.

The [yaml spec](https://yaml.org/spec/1.2.2/#54-line-break-characters) seems to be allowing both `<CR>` and `<LF>` as line break characters.

Thanks!

SGSSGene:
Fix #1309

The stream class only considers '\n' as new lines. This works as long as the line ending is '\n' or '\r\n'.
But for files with only '\r' (as with old mac text files) it doesn't work as expected.

This PR extends the stream class. It remembers the first '\r' or '\n' encounter and uses that as a line ending.

### DIFF
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -262,7 +262,24 @@ char Stream::get() {
   AdvanceCurrent();
   m_mark.column++;
 
-  if (ch == '\n') {
+  // if line ending symbol is unknown, set it to the first
+  // encountered line ending.
+  // if line ending '\r' set ending symbol to '\r'
+  // other wise set it to '\n'
+  if (!m_lineEndingSymbol) {
+    if (ch == '\n') { // line ending is '\n'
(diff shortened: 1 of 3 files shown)


jbeder (approved):

Hind-M:
Nice! Thank you for fixing this!
Is there an upcoming release scheduled soon?
Would be really great! 🙂

PR #1318, opened 2024-09-12
fix(src,include,test): fixed multiple cases where a bad yaml was accepted

FedeDP:
Also, i added test cases.

The yaml parser did not catch some strings related yaml issues, like:
```yaml
"foo","bar"
```
This is linked to the discussion here: https://github.com/falcosecurity/falco/issues/3281

Note: i am not an expert therefore please let me know if i can do anything better (or if the fix makes no sense at all :D )

### DIFF
--- a/test/integration/load_node_test.cpp
+++ b/test/integration/load_node_test.cpp
@@ -366,30 +366,31 @@ TEST(NodeTest, LoadCommaSeparatedStrings) {
   EXPECT_THROW(Load(R"("foo",)"), ParserException);
   EXPECT_THROW(Load(R"(,"foo")"), ParserException);
   EXPECT_THROW(Load(R"(,foo)"), ParserException);
+}
 
-  const char *doc_ok = R"(
-top
-, aa
-)";
-  EXPECT_NO_THROW(Load(doc_ok));
-
-  const char *doc_ok_2 = R"(
-top
-, "aa"
-)";
-  EXPECT_NO_THROW(Load(doc_ok_2));
-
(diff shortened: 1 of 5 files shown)


test/integration/load_node_test.cpp
```diff
@@ -334,6 +334,69 @@ TEST(NodeTest, LoadQuotedNull) {
...
+  EXPECT_THROW(Load(R"("foo","bar")"), ParserException);
+  EXPECT_THROW(Load(R"("foo",bar)"), ParserException);
+  EXPECT_THROW(Load(R"(,)"), ParserException);
+  EXPECT_THROW(Load(R"("foo",)"), ParserException);
+  EXPECT_THROW(Load(R"("foo","")"), ParserException);
+  EXPECT_THROW(Load(R"("foo",)"), ParserException);
+  EXPECT_THROW(Load(R"(,"foo")"), ParserException);
+  EXPECT_THROW(Load(R"(,foo)"), ParserException);
+
+  const char *doc_ok = R"(
```
jbeder, line 370: Do you want to EXPECT_EQ what these actually parse to? (And maybe put them in the above "valid" test or their own one?

jbeder (changes requested):
Looks good overall. One minor comment on the tests.

Thanks for the PR!

test/integration/load_node_test.cpp
FedeDP: Done, thanks!

jbeder (approved):

PR #1324, opened 2024-10-14
[fix] Preserving Local Tags in YAML::Load

### LINKED ISSUE #1142: Preserving Tags in YAML::Load
When I `YAML::Load` a node with a local tag, the tag type and tag contents are not preserved.

## Minimal example

I am using the 0.7.0 conan package.

```cpp
auto x = YAML::Node(42);

YAML::Emitter e;
e << YAML::LocalTag("x") << x;
std::string s = e.c_str();
auto y = YAML::Load(s);

std::cout << "before: " << s << std::endl;
std::cout << "tag:    " << y.Tag() << std::endl;
std::cout << "after:  " << YAML::Dump(y) << std::endl;
```

prints 

```console
before: !x 42
tag:    !x
after:  !<!x> 42
```

## Expected:

```console
before: !x 42
tag:    x
after:  !x 42
```

I suppose two things are going wrong here:
1. The leading exclamation mark is added to the tag contents
2. The tag type changes from `_Tag::Type::PrimaryHandle` to `_Tag::Type::Verbatim`

So I have two questions: 
1. Is this a bug or am I doing it wrong? *I am not 100% sure I correctly understand the complicated yaml specs for tags...*
2. Is there any way to set the `_Tag::Type` for an existing node as a workaround?

IMaloney:
* fixes #1142 
* adds a test case that applies a local tag to a node using the emitter, then loads the output of the emitter as a node


jbeder (approved):

src/emitfromevents.cpp
```diff
@@ -116,8 +116,13 @@ void EmitFromEvents::BeginNode() {
 }
 
 void EmitFromEvents::EmitProps(const std::string& tag, anchor_t anchor) {
-  if (!tag.empty() && tag != "?" && tag != "!")
-    m_emitter << VerbatimTag(tag);
+  if (!tag.empty() && tag != "?" && tag != "!"){
+      if (tag[0] == '!') {
```
davidzchen, line 120: Sorry I'm late to this, but could you fix the indentation here? The codebase uses 2 spaces for indentation, but there is 4 here. Thanks in advance.

jbeder: Can you just submit a new PR? Agreed, this PR should have had clang-format run.

PR #1325, opened 2024-10-20
emitter: Support std::string_view

dendy:
Accept Emitter::operator<<(std::string_view).

ABI remains C++11 compatible by exposing new method Emitter::Write(const char*, size_t).

All affected calls optimized to pass std::string values as pointer + size tuple into appropriate routines.

### DIFF
--- a/include/yaml-cpp/emitter.h
+++ b/include/yaml-cpp/emitter.h
@@ -9,11 +9,16 @@
 
 #include <cmath>
 #include <cstddef>
+#include <cstring>
 #include <limits>
 #include <memory>
 #include <sstream>
 #include <string>
 #include <type_traits>
+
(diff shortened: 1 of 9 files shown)


jbeder (changes requested):
This all looks good. Can you add a test case that emits a string_view?

dendy:
@jbeder Could you please take a look.

1. Since Emitter's `operator<<` are inlined it is safe to direct `std::string` and `const char*` overloads to the new `Write::(const char*, size_t)`. Older applications will preserve binary compatibility.

2. Fixed comparing null literals.

3. Fixed calling `RexEx::Matches` and added static_assert() to check for accidental passing raw C-string into the `Matches`.

4. Added 3 Emitter tests: for `std::string`, null terminated and unterminated `std::string_view`.

All tests are passing at my side when I configure yaml-cpp with extra option `CMAKE_CXX_STANDARD=17`.

jbeder (approved):

PR #1333, opened 2024-12-22
Forbid incorrect document start (---)

azat:
Previously documents like the following was simply ignored:

    ----
    # foo
    bar: bam

### DIFF
--- a/src/exp.h
+++ b/src/exp.h
@@ -82,7 +82,7 @@ inline const RegEx& Utf8_ByteOrderMark() {
 // actual tags
 
 inline const RegEx& DocStart() {
-  static const RegEx e = RegEx("---") + (BlankOrBreak() | RegEx());
+  static const RegEx e = RegEx("---");
   return e;
 }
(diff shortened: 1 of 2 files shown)


jbeder (changes requested):
I think there *is* a bug, but I don't think this is the correct fix - right now, the behavior is correct for the following cases:

```
$ echo '----' | ./util/parse
----
```

This correctly parses it as a single scalar node "----". I suspect this will be incorrect with your solution.

```
$ echo '----\nfoo: bar' | ./util/parse
yaml-cpp: error at line 2, column 4: illegal map value
```

This is your first test - it's properly marked as an error in the existing code, and I think it may switch to the wrong error in your code.

However:

```
$ echo '----\n# foo\nbaz: 1' | ./util/parse
----
```

This is your second test, and this is wrong, as you point out; it should be an error.

SGSSGene:
This I think is also fixed by PR #1406

azat:
Indeed, thanks!

PR #1355, opened 2025-04-15
Replace std::shared_ptr<T>(T*) usage in order to prevent additional memory allocation for shared_ptr control block

slavanap:


flomnes (commented):
What this change does is to replace calls to `std::shared_ptr<T>::shared_ptr(T*)` with 2 successive calls
- `std::make_sharted<T>`
- `std::shared_ptr<T>::shared_ptr(const std::shared_ptr<T>&)`. From C++17, copy elision will prevent any constructors from being actually called here.

I'm not sure how it's supposed to prevent additional memory allocations. Care to explain ?

slavanap:
Sure. When the first owner of shared_ptr is created with a constructor from a pointer, it allocates memory for a control block used to reference counting. make_shared method allocates memory for that control block and the object that pointer is stored, with the same allocation - thus 1 extra allocation is eliminated. More precise description is available in the "notes" section here https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared.html

PR #1358, opened 2025-04-29
[cmake] Require CMake 3.15 or newer

gruenich:
* [x] The currently used version of Google-test 1.13 still triggers a CMake deprecation warning with CMake 3.30 or newer. We need to decide on which version of Google-test we want to update. I see no reason to not choose the most recent version 1.16 which requires C++14 (like Google-test 1.13) and CMake 3.13. (I went for Google-test 1.16)
* [x] Adjust CI to test for both CMake 3.15 and 4.3, the new min and max required version.


CMakeLists.txt
```diff
@@ -63,10 +57,8 @@ endif()
 
 set(contrib-pattern "src/contrib/*.cpp")
 set(src-pattern "src/*.cpp")
-if (CMAKE_VERSION VERSION_GREATER 3.12)
-  list(INSERT contrib-pattern 0 CONFIGURE_DEPENDS)
-  list(INSERT src-pattern 0 CONFIGURE_DEPENDS)
-endif()
+list(INSERT contrib-pattern 0 CONFIGURE_DEPENDS)
```
henryiii, line 60: You can just put the `CONFIGURE_DEPENDS` in the `file(GLOB` below, this was probably only to conditionally include it.

test/CMakeLists.txt
```diff
@@ -18,10 +18,8 @@ endif()
 
 set(test-new-api-pattern "new-api/*.cpp")
 set(test-source-pattern "*.cpp" "integration/*.cpp" "node/*.cpp")
-if (CMAKE_VERSION VERSION_GREATER 3.11)
-  list(INSERT test-new-api-pattern 0 CONFIGURE_DEPENDS)
-  list(INSERT test-source-pattern 0 CONFIGURE_DEPENDS)
-endif()
+list(INSERT test-new-api-pattern 0 CONFIGURE_DEPENDS)
```
henryiii, line 21: Same on these

henryiii:
For context, googletest does not specify a range, which is why the minimum matters there.

CMakeLists.txt
gruenich: You are right, I removed even more lines.

henryiii:
I do it in [pybind11](https://github.com/pybind/pybind11/blob/master/.github/workflows/configure.yml) and [CLI11](https://github.com/CLIUtils/CLI11/blob/3ed270937e909048a526cbd4cd674bed06a2fc5b/.github/workflows/tests.yml#L236-L381) ([local support action](https://github.com/CLIUtils/CLI11/blob/main/.github/actions/quick_cmake/action.yml)). Basically `jwlawson/actions-setup-cmake` downloads binaries pretty quickly.

gruenich:
I removed a maximum version, just requiring 3.15 seems the saner choice for now.

Two points are left:
1. Are you okay with relicensing the quick_cmake action from CLI?
2. The Action is not reported as a success. Not sure what I missed.

SGSSGene:
Thank you for updating this PR, I will take a look at it this week.
I can't decide relicensing issues, only @jbeder can do that.

>     2. The Action is not reported as a success. Not sure what I missed.

Github does not provide a runner for ubuntu 26.04. Please use 24.04 or pick any other image, see https://github.com/actions/runner-images

jbeder:
What does this mean, relicensing?

henryiii:
CMake is designed to always have an upper version due to its policy system. Upper versions keep newer CMake versions working longer. Setting a minimum version of 3.15 and not setting an upper version might start throwing a warning at some point in the near future. Be setting an upper version too, say to 4.0, should work for many many more years.

The policies will be set to 3.15 if you don't set an upper version. While they will be set to the highest version that CMake supports up until the upper version if you set an upper version. So as long as CMake supports whatever value is in the upper version as its minimum version, it will still work.

3.4...3.15 doesn't show a warning today, for example, because CMake does know how to build stuff for 3.15 without a warning. While 3.4 by itself can't be used, because CMake doesn't have those policies anymore

jbeder:
Sure, I'll leave that to @SGSSGene to decide. But there was some discussion about relicensing?

gruenich:
> But there was some discussion about relicensing?

Yes. I copied the file `.github/actions/quick_cmake/action.yml` from henryiii's repository https://github.com/CLIUtils/CLI11 which is licensed under a 3-clause BSD license. This is MIT, so we need @henryiii's permission to adopt the license.

henryiii:
You have my permission, certainly.

jbeder:
Ah, that's fine, it's pretty small and straightforward code.

SGSSGene (approved):

Issue #1276, opened 2024-03-27
Figuring out if a string will round-trip


gyorokpeter:
E.g. if I render a string with the value `true` or `1.1`, when parsed back they become a boolean and a number respectively.
Is there a way to determine if a certain string doesn't round trip and therefore needs quotes added when emitting it?

SGSSGene:
I am not sure what you are asking for.
If you read them with `node.as<std::string>()` they will roundtrip just fine.

Maybe you can clarify?

gyorokpeter:
It has been a while since I raised this, and we made a workaround in the meantime, but the situation was roughly this:
- We have a function that takes some object provided by the user and converts it to YAML
- It uses `YAML::Emitter`'s `operator<<` to output the object
- Previously, if the user object was a string, it would just call `out << str`, which blindly outputs the string without consideration of whether it has some special format such that when parsed back it turns into something else such as a number, boolean, null etc.
- This was causing problems when the user object was something like `["my", "true", "love"]`, where the `"true"` value would be printed out without the quotes, so when parsed back, it would become the boolean value `true` instead of the string `"true"`
- The workaround was to create a dedicated "render string" function that matches the input against a laundry list of regexes, such as those for all the number formats, infinities, true/false etc., and if any of them match, does `out << YAML::SingleQuoted` first. Since this looks like a common operation for anyone who would like to output YAML from user data, I was wondering if there was some level of support in the library itself, such as an "is special value" function that encapsulates the check for whether the string would round-trip without the quotes.

PR #1363, opened 2025-07-10
emitter: emit quotes for non-plain scalars

### LINKED ISSUE #1362: emitter loses type info for some strings
A quoted string value that "looks like" an integer, bool or float value loses the quotes when emitted and so changes type and is no longer a string.  The output from pyyaml is also shown for comparison; the values remain quoted.

```
$ cat lost-quotes.yaml 
["true", "false", "3"]

$ cat lost-quotes.cpp 
#include <cassert>
#include <iostream>

#include "yaml-cpp/yaml.h"

int main(int argc, char** argv)
{
   char input[1024];
   std::cin.getline(input, sizeof(input));
   assert(std::cin.good());
   std::cout << YAML::Load(input) << '\n';
   return 0;
}

# build lost-quotes here ...
$ cat lost-quotes.yaml | build/debug/lost-quotes 
[true, false, 3]

$ cat lost-quotes.yaml | python -c 'import sys; import yaml; print(yaml.safe_load(sys.stdin))'
['true', 'false', '3']
```

johncoomes:
Use the "!" tag to detect scalars that are not plain scalars and pass EMITTER_MANIP::DoubleQuoted to Emitter::Write() to ensure they are quoted.

This preserves quotes on quoted items that do not strictly need quotes. The alternative of detecting and quoting only the values that need quotes seemed error-prone and brittle.

Fixes #1362.

### DIFF
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -67,9 +67,31 @@ TEST_F(EmitterTest, UnterminatedStdStringViewScalar) {
 #endif
 
 TEST_F(EmitterTest, SimpleQuotedScalar) {
-  Node n(Load("\"test\""));
+  const std::string quoted = "\"test\"";
+  Node n(Load(quoted));
   out << n;
-  ExpectEmit("test");
+  ExpectEmit(quoted);
+}
+
+TEST_F(EmitterTest, QuotedScalarFalse) {
+  const std::string quoted = "\"false\"";
+  Node n(Load(quoted));
+  out << n;
+  ExpectEmit(quoted);
+}
+
+TEST_F(EmitterTest, QuotedScalarTrue) {
+  const std::string quoted = "\"true\"";
+  Node n(Load(quoted));
+  out << n;
+  ExpectEmit(quoted);
+}
+
+TEST_F(EmitterTest, QuotedScalarInt) {
+  const std::string quoted = "\"3\"";
+  Node n(Load(quoted));
+  out << n;
+  ExpectEmit(quoted);
 }
 
 TEST_F(EmitterTest, DumpAndSize) {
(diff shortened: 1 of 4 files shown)


test/integration/emitter_test.cpp
```diff
@@ -67,9 +67,31 @@ TEST_F(EmitterTest, UnterminatedStdStringViewScalar) {
 #endif
 
 TEST_F(EmitterTest, SimpleQuotedScalar) {
-  Node n(Load("\"test\""));
+  const std::string quoted = "\"test\"";
+  Node n(Load(quoted));
   out << n;
-  ExpectEmit("test");
+  ExpectEmit(quoted);
```
jbeder, line 73: This one is a little weird; I'd much prefer that it would display

```
test
```

rather than

```
"test"
```

since the quotes are redundant.

johncoomes: The behavior is different, but seems reasonable (to me) since the input value was already quoted; it's not adding quotes.  To maintain correct behavior while only quoting when strictly necessary requires a reliable way to detect that it's safe to omit quotes. https://github.com/jbeder/yaml-cpp/issues/1276 mentions using "a laundry list of regexes, such as those for all the number formats, infinities, true/false etc." to check.  Looking at the yaml spec, that seems error prone.

johncoomes: Looking further in the yaml spec, it has a list of regular expressions in the [Core Schema](https://yaml.org/spec/1.2.2/#103-core-schema) chapter.  The regexes would be different for the fail-safe or json schemas, but AFAICT yaml-cpp does not support those.

SGSSGene: I agree with @johncoomes .
The goal is to preserve round trips and the 'Load' already marks the the value of `test` with tag `?` and `"test"` with `!`. So why not use it to emit the same (un-)quoted style?

PR #1374, opened 2025-09-19
Update CXX requirement

mosfet80:
googletest 1.13 requires at least C++14

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -43,9 +43,9 @@ target_compile_options(yaml-cpp-tests
 target_link_libraries(yaml-cpp-tests
   PRIVATE
     Threads::Threads
-    yaml-cpp
-    gtest
-    gmock)
+    yaml-cpp::yaml-cpp
+    GTest::gtest
+    GTest::gmock)
 
 set_property(TARGET yaml-cpp-tests PROPERTY CXX_STANDARD_REQUIRED ON)
 if (NOT DEFINED CMAKE_CXX_STANDARD)


DarthGandalf:
https://github.com/jbeder/yaml-cpp/blob/80ea0028c7ba35e6a267e1778eab78a3ee82e8b9/test/CMakeLists.txt#L52 needs updating too

test/CMakeLists.txt
SGSSGene: The dependency should use the ALIAS/IMPORTED name, which will automatically use the dependency compiler version requirements.
```suggestion
yaml-cpp::yaml-cpp
GTest::gtest
GTest::gmock)
```

test/CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ target_link_libraries(yaml-cpp-tests
 
 set_property(TARGET yaml-cpp-tests PROPERTY CXX_STANDARD_REQUIRED ON)
 if (NOT DEFINED CMAKE_CXX_STANDARD)
-  set_target_properties(yaml-cpp-tests PROPERTIES CXX_STANDARD 11)
+  set_target_properties(yaml-cpp-tests PROPERTIES CXX_STANDARD 17)
 endif()
```
SGSSGene: With the other change, it should be possible to delete these lines.

SGSSGene:
> @SGSSGene New versions of googletest require C++17, therefore yaml-cpp fails test without this patch

So, the issue comes up, when using for example `YAML_USE_SYSTEM_GTEST`.
This shows that yaml-cpp is not setting up the dependencies correctly.

We should fix how the dependency are being setup, so this issue does not appear in the future.

test/CMakeLists.txt
mosfet80: fixed gtest

test/CMakeLists.txt
```diff
@@ -43,13 +43,11 @@ target_compile_options(yaml-cpp-tests
...
     Threads::Threads
-    yaml-cpp
-    gtest
-    gmock)
+    yaml-cpp)
 
 set_property(TARGET yaml-cpp-tests PROPERTY CXX_STANDARD_REQUIRED ON)
 if (NOT DEFINED CMAKE_CXX_STANDARD)
-  set_target_properties(yaml-cpp-tests PROPERTIES CXX_STANDARD 11)
+  set_target_properties(yaml-cpp-tests PROPERTIES CXX_STANDARD 17)
```
SGSSGene: This line does not to be changed. Leave the standard at 11.
```suggestion
  set_target_properties(yaml-cpp-tests PROPERTIES CXX_STANDARD 11)
```

Issue #1016, opened 2021-07-28
YAML Floating Point Precision in Output Files


holdsworth:
When saving a file with new API you cannot set explicitly the number of digits after the . thus when having a double of the form 1.0 it is saved as 1, which causes issues later on if you're deducing parameter type based on input, this badly hurts the config file (especially when you re-save it)

The fix would be to modify the last else in inner_encode function (line 102 in node/convert.h file) from stream << ... to  stream << std::setprecision(12) << std::fixed << rhs;

Note that you can expose the set precision to be externally modified, would you like me to send a PR?

jbeder:
Sure

fujitatomoya:
@jbeder we have been blocked by this problem for a long time. besides i am new for yaml-cpp, could you give us a clue how to address this issue?

jbeder:
I think the OP was going to submit PR? Anyways, patches welcome.

fujitatomoya:
I am not sure who OP is, but with suggested code, i tried with https://github.com/fujitatomoya/ros2_persist_parameter_server/issues/13#issuecomment-1843930024 but it did not work. it would be appreciated if i got anything wrong.

Issue #412, opened 2016-08-23
Emitting Float With Trailing Decimal .0


LemonPi:
Currently any float/double with no decimal component gets emitted as an integer.

```
float a = 1.0;
YAML::Emitter emitter;

// regardless of the Precision manipulator
emitter << YAML::Precision(2) << a;

// outputs 1
```
Is there a way to make it output `1.0`?

The YAML file I'm emitting should have a mix of ints and floats, which gets parsed by an external script. That script needs to be able to distinguish floats and ints.

arneboe:
Did you ever find a solution for this? I am facing the same problem

PR #1377, opened 2025-10-16
Emitter: produce string with trailing '.0' for floating point value with only integral part

### LINKED ISSUE #226: Saving Decimal Numbers as Integers
```
Save 1.0f to a YAML::Node and the file will only show 1, which when loaded will 
yield an integer if you test if .as<int> doesn't throw an exception before 
trying to convert it to a float.

It should output 1.0 to the file instead.

```

Original issue reported on code.google.com by `ARandomFurry` on 16 Nov 2013 at 10:11

romainreignier:
For example the number 1.0 will be converted to "1.0" instead of "1".
This allows to keep the correct type information when parsing back the YAML.

This is an attempt to Fix #226 #412 #1016

### DIFF
--- a/src/fptostring.cpp
+++ b/src/fptostring.cpp
@@ -4,6 +4,8 @@
 #include <array>
 #include <cassert>
 #include <cmath>
+#include <iomanip>
+#include <ios>
 #include <limits>
 #include <sstream>
 #include <tuple>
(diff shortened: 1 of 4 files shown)


SGSSGene:
Hey,
I am uncertain of this PR. I do not think this fixes a bug, but adds a functionality that many would like to have.

Reading the yaml spec I understand that it has 3 types: `sequence`, `collection` and `scalar`.
Exporting a float without dot (e.g `0.0`) is valid. So interfering the type is not a float from the missing of a dot is not valid.

The correct fix, would probably be add a explicit '!!float' tag to indicate the type, but yaml-cpp does not support this.

I have done some similar fix for some other library which uses yaml-cpp: https://github.com/fujitatomoya/ros2_persist_parameter_server/issues/13#issuecomment-3646122529

SGSSGene:
So, what we maybe actually need, is someway of configuring the Emitter to add a ".0" or not. (There might also be other things we want to configure?)

Issue #1399, opened 2026-02-11
EmitterTest fails on x86 system


DarthGandalf:
gentoo linux, building in 32 bit mode

[build-abi_x86_32.x86.log](https://github.com/user-attachments/files/25223739/build-abi_x86_32.x86.log)
[LastTest.log](https://github.com/user-attachments/files/25223740/LastTest.log)

```
[ RUN      ] EmitterTest.NumberPrecision
/mnt/portagetmp/portage/dev-cpp/yaml-cpp-0.9.0/work/yaml-cpp-yaml-cpp-0.9.0/test/integration/emitter_test.cpp:30: Failure
Expected equality of these values:
  expected
    Which is: "- 3.14\n- 54\n- 2.4e+06\n- 2e+06\n- 1e+07"
  out.c_str()
    Which is: "- 3.14\n- 54\n- 2.4e+05\n- 2.8e+05\n- 1.2e+05"
With diff:
@@ -1,5 +1,5 @@
 - 3.14
 - 54
-- 2.4e+06
-- 2e+06
-- 1e+07
+- 2.4e+05
+- 2.8e+05
+- 1.2e+05


[  FAILED  ] EmitterTest.NumberPrecision (0 ms)
[ RUN      ] EmitterTest.DefaultPrecision
/mnt/portagetmp/portage/dev-cpp/yaml-cpp-0.9.0/work/yaml-cpp-yaml-cpp-0.9.0/test/integration/emitter_test.cpp:30: Failure
Expected equality of these values:
  expected
    Which is: "- 1.3125\n- 1.23455810546875"
  out.c_str()
    Which is: "- 1.3125\n- 1.270590651e-05"
With diff:
@@ -1,2 +1,2 @@
 - 1.3125
-- 1.23455810546875
+- 1.270590651e-05


[  FAILED  ] EmitterTest.DefaultPrecision (0 ms)
[ RUN      ] EmitterTest.SetPrecision
/mnt/portagetmp/portage/dev-cpp/yaml-cpp-0.9.0/work/yaml-cpp-yaml-cpp-0.9.0/test/integration/emitter_test.cpp:30: Failure
Expected equality of these values:
  expected
    Which is: "- 1.31\n- 1.23456"
  out.c_str()
    Which is: "- 1.31\n- 1.27059e-05"
With diff:
@@ -1,2 +1,2 @@
 - 1.31
-- 1.23456
+- 1.27059e-05


[...]

PR #1401, opened 2026-02-11
fix: floating point conversion on x86 (32bit)

SGSSGene:
Fixes floating point conversion on 32bit system as mentioned in #1399.
dragonbox function `to_decimal` returns a unsigned 64bit number for the significand.
This is stored in a `size_t` which is a uint32_t on 32 bit system those causes loose of precision.
This PR simply replaces the `size_t` with an `uint64_t` type.


stephanlachnit:
@jbeder do you think it would be possible to release a 0.9.1 with this fix?

PR #1427, opened 2026-04-01
Fix bug on tags starting with "!!"

### LINKED ISSUE #1373: YAML::Node::SetTag broken on inputs starting with "!!"
Below is a minimal example to reproduce the issue. It appears using `SetTag` breaks (w.r.t console/file emitting for the very least) `string_node` when tagging with anything starting with "!!". Not only that `string_node` appears broken, but also any node where `string_node` is inserted. When the tagged node is inserted into `root_node`, writing the integer to it will not work either. Try `SetTag` to whatever does not begin with "!!" and the console output looks like you would probably expect, e.g.

```
some_string: !some_tag hello
some_int: 2
```

To reproduce:
```
    YAML::Node root_node{};
    YAML::Node string_node{"hello"};
    string_node.SetTag("!!str"); // Breaks for anything starting with "!!" Comment out this line or tag with e.g. "!some_tag" and it works as expected 

    root_node["some_string"] = string_node;
    root_node["some_int"] = 2;

    std::cout << root_node << std::endl; // Outputs "some_string: !"
```
**Version:** commit 2f86d13775d119edbb69af52e5f566fd65c6953b, latest commit on master at the time of posting
**Os:** Ubuntu 24
**Compiler:** Gcc 14.2.0

sxrzh:
Fixes #1373 .

This adds support for SecondaryTag in EmitFromEvents::EmitProps, resolving tags starting with "!!".

Now the output of the example code is

```yaml
some_string: !!str hello
some_int: 2
```


SGSSGene (changes requested):
@adfhkjafsdk Thank you so much. Good to see progress on these issues!

You are using 'tabs' instead of spaces in the source code. Please change it to spaces.

About the 'Named handles'.
If I read the specs correctly a tag `!a!` is not allowed, because it has a non-empty suffix (`!a!b` would be fine).
See [6.9.1 Node Tags](https://yaml.org/spec/1.2.2/#691-node-tags) specifically the `Tag shorthands`.
The example in `Example 6.27 Invalid Tag Shorthands` even shows this as invalid.
So I think the right thing to do is throw an exception for invalid tag.

Secondly, currently we do not really have support for named tags at all.
Strictly speaking even `!a!b` would be invalid if `!a!` wasn't specified before.
The currently `master` implementation just handles this case wrongly which might be very surprising for library users.
I think it would be much nice to also throw an exception with some thing like "!TODO: Unsupported named tag". And maybe in that code section a comment to this PR so the next person knows where we left off.

@jbeder: do you have any insights to the TAG support of yaml-cpp?

adfhkjafsdk:
@SGSSGene Thank you for your hard work. I feel sorry about the poor indentation.

I totally agree with the idea of throwing an exception for invalid tag. I wonder why the old implementation just ignore the invalid tag `!`  when meeting it.

I know that `!a!b` is a Tag Shorthand, and it is invalid without previous `TAG` directive. But since we've allowed usage like `out << LocalTag("a", "foo") << "bar";` (see LocalTagInNameHandle in test/integration/emitter_test.cpp), which outputs `!a!foo bar`, I think this should also be allowed through `SetTag`. Anyway, if we ban the two ways, there may be no way to emit tag shorthands. I think it may be better to let the emitter check whether the `TAG` directive exists.

Currently there's no `throw` in the emitter code and no Exception type for emitter exceptions. Maybe the author thought checking validity isn't what the emitter is supposed to do?

By the way I felt it inconvenient to store tag as a raw string (`YAML::detail::node_data::m_tag`), maybe in the future we should change it to more structured representations.

SGSSGene:
I just accepted #1437 which solves a subset of the problems mentioned here.
If you would/could rebase your PR I would also like to accept it.

sxrzh:
@SGSSGene Thank you! I've rebased my commits.

I moved my tests to load_node_test.cpp in consideration of consistency, but I wonder why #1437 put the tests for `SetTag` into load_node_test.cpp rather than emitter_test.cpp. After all `SetTag` is an Emitter feature, not a Loader feature.

SGSSGene (approved):
Thank you!

SGSSGene:
> @SGSSGene Thank you! I've rebased my commits.
> 
> I moved my tests to load_node_test.cpp in consideration of consistency, but I wonder why #1437 put the tests for `SetTag` into load_node_test.cpp rather than emitter_test.cpp. After all `SetTag` is an Emitter feature, not a Loader feature.

@sxrzh The tests are in the wrong location. Could you check PR #1462? That PR fixes the location.

jbeder:
I'm not sure I'm following all of the questions, but the one about emitter exceptions: I remember wanting the emitter to be exception-free, but I'm not certain of why. (I remember being sure of this!) My best guess now: when you're emitting, you want freedom to produce whatever output you like; and it's a distributed behavior (you emit in many lines) whereas parsing is a single line that you could wrap with an exception handler. But this was all a long time ago.

If you choose to change this, that seems fine, but I'd recommend:

1. Be very deliberate, and do so holistically.
2. bump the version number
3. Write a note in the changelog

sxrzh:
Thanks a lot!

PR #1429, opened 2026-04-07
Increase output buffer size for float to string conversion

PhilipDeegan:
avoid werror #1428


SGSSGene:
This is a false warning. Please use PR #1430 to suppress the warning/error when required.
It introduces a `YAML_CPP_USE_STRICT_FLAGS` cmake option.

PR #1431, opened 2026-04-09
Fix MSVC narrowing-conversion warnings (C4244/C4267) under /W3

benhillis:
Add explicit casts to silence 8 warnings triggered by MSVC /W3:

fptostring.cpp:
- Add static_cast<int>(precision) where public API (size_t) calls internal template (int)
- Add static_cast<int>(...) for pointer-arithmetic result stored in int

dragonbox.h:
- Add static_cast<uint_least32_t>(n) in umul64() calls inside if-constexpr branches guarded by is_same<UInt, uint_least32_t> to silence MSVC warnings on discarded branches when UInt=uint_least64_t
- Change trailing-zero counter 's' from size_t to DecimalExponentType in both remove_trailing_zeros specializations to match exponent type


SGSSGene:
Thank you for the contribution.

I just wanted to say I have the PR on my mind just not the time yet to check/merge it.

PR #1432, opened 2026-04-12
[ci] Test for compatibility with C++23

gruenich:


gruenich:
It might be worth reducing the number of combinations by just checking 11 and 23 and excluding the versions in between 17 and 20.

SGSSGene:
> It might be worth reducing the number of combinations by just checking 11 and 23 and excluding the versions in between 17 and 20.

I think we should keep 11, 20 and 23.
11 because its the oldest
20 because its gcc default in current compiler
23 because its the newest

What do you think?

gruenich:
Good idea. I adjusted the commit.

SGSSGene (approved):

PR #1437, opened 2026-05-17
emit secondary tag handles (e.g. !!str) on Node::SetTag

### LINKED ISSUE #1373: YAML::Node::SetTag broken on inputs starting with "!!"

c-tonneslan:
Closes #1373.

When a node's tag started with `!!` (the YAML secondary tag handle, e.g. `!!str`), `EmitFromEvents::EmitProps` stripped a single `!` and emitted the rest as a `LocalTag`. The remaining `!str` then failed the local-tag character regex on its inner `!` and the emitter bailed out with `INVALID_TAG`, truncating the output:

```
YAML::Node root;
YAML::Node s{"hello"};
s.SetTag("!!str");
root["some_string"] = s;
root["some_int"] = 2;
std::cout << root;  // -> "some_string: !"
```

Route the `!!` case through `SecondaryTag` with the rest of the tag content. After this:

```
some_string: !!str hello
some_int: 2
```

Two regression tests added under `NodeTest` covering both the secondary (`!!`) and primary (`!`) handle paths.

### DIFF
--- a/src/emitfromevents.cpp
+++ b/src/emitfromevents.cpp
@@ -117,7 +117,12 @@ void EmitFromEvents::BeginNode() {
 
 void EmitFromEvents::EmitProps(const std::string& tag, anchor_t anchor) {
   if (!tag.empty() && tag != "?" && tag != "!"){
-      if (tag[0] == '!') {
+      if (tag.size() >= 2 && tag[0] == '!' && tag[1] == '!') {
+        // Tags like "!!str" use the secondary tag handle; emit them through
+        // SecondaryTag so the rendered output keeps both bangs instead of
+        // failing the LocalTag regex check on the inner '!'. See #1373.
+        m_emitter << SecondaryTag(std::string(tag.begin()+2, tag.end()));
+      } else if (tag[0] == '!') {
         m_emitter << LocalTag(std::string(tag.begin()+1, tag.end()));
       } else {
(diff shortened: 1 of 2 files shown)


SGSSGene:
Thank you for this PR!

PR #1439, opened 2026-05-26
docs: note YAML_CPP_INSTALL for FetchContent embeds

### LINKED ISSUE #1381: Default behaviour for yaml-cpp install targets missing in README
I struggled a lot to find out why the targets for yaml-cpp did not get installed, when fetching them like described in the README in the CMakeLists.txt of my own project. 

Now I found out, that in this PR (https://github.com/jbeder/yaml-cpp/pull/1007) it was changed to not be always ON by default to install the targets. 

It would probably be really helpful to add a small comment to the README under the section 'How to integrate it within your project using CMake' that it could be necessary to add `option(YAML_CPP_INSTALL ON)` if yaml-cpp is not the top project, because it could save time for other developers as this feels like a pretty standard use case. 

If you agree it would be nice to add this.

wyf027:
## Summary

When yaml-cpp is pulled in with `FetchContent`, it is not the top-level CMake project, so `YAML_CPP_INSTALL` defaults to `OFF` and install targets are not generated (see #1007). This adds a short README note under the CMake integration section.

Fixes #1381

## Test plan

- [x] README-only change

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,10 @@ FetchContent_Declare(
   GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
   GIT_TAG <tag_name> # Can be a tag (yaml-cpp-x.x.x), a commit hash, or a branch name (master)
 )
+# Install rules (CMake package config, headers, etc.) are enabled only when yaml-cpp is
+# the top-level CMake project. When embedding via FetchContent, set this before
+# FetchContent_MakeAvailable if you need install targets in your build tree:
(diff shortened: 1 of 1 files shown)


wyf027:
Docs-only PR (FetchContent + YAML_CPP_INSTALL). Happy to adjust wording if you prefer it in CONTRIBUTING instead of README.

wyf027:
Small docs clarification for FetchContent users — ready for review when you have time.

wyf027:
Small FetchContent docs note — ready for review.

Issue #1287, opened 2024-06-22
`YAML::Literal` does not work for `YAML::Binary`


davidzchen:
I noticed that when `YAML::Literal` is used when emitting a `YAML::Binary`, it does not take effect.

Code:

```cpp
#include <cstdint>
#include <cstdlib>
#include <iostream>
#include <vector>

#include "yaml-cpp/yaml.h"

int main(int argc, char** argv) {
  std::vector<uint8_t> data;
  for (int i = 0; i < 256 * 10; i++) {
    data.push_back(i % 256);
  }

  YAML::Binary binary(&data.front(), data.size());
  YAML::Emitter out;
  out << YAML::BeginMap;
  out << YAML::Key << "data";
  out << YAML::Value << YAML::Literal << binary;

  std::cout << out.c_str() << std::endl;

  return 0;
}
```

Output:

```
[...]

PR #1447, opened 2026-06-01
feat/fix: support literal and single-quoted style for YAML::Binary emitting

### LINKED ISSUE #1445: No way to emit `YAML::Binary` in literal, single-quoted or plain form.
As mentioned in issue #1287 , `YAML::Literal` does not work for `YAML::Binary`.
In fact, `YAML::SingleQuoted` does not work for `YAML::Binary` either. Besides there is no way to emit `YAML::Binary` without quotes. Currently the emitter applies double-quoted style for all binary nodes.

sxrzh:
Fixes #1445 and a part of issue #1287. 
Support literal and single-quoted style for YAML::Binary emitting. Attempts to support plain form may change the default behavior, so are not applied.
Related tests are added to `EmitterTest.BinaryStyles` in test/integration/emitter_test.cpp.

### DIFF
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -976,8 +976,34 @@ Emitter& Emitter::Write(const Binary& binary) {
   if (!good())
     return *this;
 
+  const StringFormat::value strFormat = binary.size() == 0 ? 
+      Utils::ComputeStringFormat("", 0, m_pState->GetStringFormat(),
+                                 m_pState->CurGroupFlowType(), false) : 
+      Utils::ComputeStringFormat("a", 1, m_pState->GetStringFormat(),
+                                 m_pState->CurGroupFlowType(), false);
+
+  if (strFormat == StringFormat::Literal) 
+    m_pState->SetMapKeyFormat(YAML::LongKey, FmtScope::Local);
+  
   PrepareNode(EmitterNodeType::Scalar);
-  Utils::WriteBinary(m_stream, binary);
+
+  switch (strFormat) {
+    // case StringFormat::Plain:		// to avoid changing the default behavior
+    //   m_stream.write(EncodeBase64(binary.data(), binary.size()));
+    //   break;
+    case StringFormat::SingleQuoted:
+      Utils::WriteSingleQuotedBinary(m_stream, binary);
+      break;
+    case StringFormat::Plain:
+    case StringFormat::DoubleQuoted:
+      Utils::WriteBinary(m_stream, binary);
+      break;
+    case StringFormat::Literal:
+      Utils::WriteLiteralBinary(m_stream, binary, 
+                                m_pState->CurIndent() + m_pState->GetIndent());
+      break;
+  }
+
   StartedScalar();
 
   return *this;
--- a/src/emitter.cpp
+++ b/src/emitter.cpp
@@ -976,11 +976,9 @@ Emitter& Emitter::Write(const Binary& binary) {
   if (!good())
     return *this;
 
-  const StringFormat::value strFormat = binary.size() == 0 ? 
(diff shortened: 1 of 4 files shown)


sxrzh:
I also implemented the line wrapping as mentioned in #1287, should I commit it here or open another PR?

src/emitter.cpp
```diff
@@ -976,8 +976,34 @@ Emitter& Emitter::Write(const Binary& binary) {
...
+    case StringFormat::Plain:
+    case StringFormat::DoubleQuoted:
+      Utils::WriteBinary(m_stream, binary);
+      break;
+    case StringFormat::Literal:
+      Utils::WriteLiteralBinary(m_stream, binary, 
+                                m_pState->CurIndent() + m_pState->GetIndent());
+      break;
+  }
+
```
SGSSGene, line 1004: Could you add some comment for this block, what actually is happening?
Why do you format create a format string for "a"?
And why is there a case for `case StringFormat::Plain` that is commented out?

test/integration/emitter_test.cpp
```diff
@@ -1468,8 +1496,8 @@ TEST_F(EmitterTest, Infinity) {
   out << YAML::EndMap;
 
   ExpectEmit(
-	  "foo: .inf\n"
-	  "bar: .inf");
+      "foo: .inf\n"
+      "bar: .inf");
```
SGSSGene, line 1500: Could you explain why the indentation of these change?

SGSSGene (changes requested):
Thank you for this PR!

There are two concerns:
1. The code that does the switching is hard to read and check for correctness. Some (non-trivial) comments could be helpful.
2. Why does the indentation for pointing float type change?

SGSSGene:
> I also implemented the line wrapping as mentioned in #1287, should I commit it here or open another PR?

Hey, a second PR for #1287 would be appreciated.

test/integration/emitter_test.cpp
sxrzh: This may be caused by clang-format. Previously this case used tab for indentation and the formatter changed it to spaces.

SGSSGene: > This may be caused by clang-format. Previously this case used tab for indentation and the formatter changed it to spaces.

Ah yes! than just leave the fix in. I just noticed other PRs have the same change.

src/emitter.cpp
sxrzh: About "a"：
Just like Emitter::Write for `const char*`, we should use ComputeStringFormat to generate StringFormat. And I also thought it necessary to offer functions like WriteSingleQuotedBinary and WriteLiteralBinary in Utils and call them, rather than directly generating base64 and calling Utils functions for strings. However, ComputeStringFormat indeed requires a string, but the base64 string is not generated in Emitter::Write! If we encode twice, it may cost double the time with long binary.
In fact ComputeStringFormat returns the same value for any non-empty base64 string and `a`, so I just passed `a` for all  non-empty Binary.
Maybe adding a `ComputeBinaryFormat` will look better?

About `Plain`：
Previously, without any options, the strFormat remains `Plain` by default but emits double-quoted style for Binary. Some long-existing tests are also based on this. Changing the default behavior may also influence some downstream projects. So I left it unchanged (still double-quoted).

sxrzh: Best appreciation to you!

sxrzh:
Now it may look better.
I'll soon submit another PR after this one is merged.
:)

SGSSGene:
> Now it may look better. I'll soon submit another PR after this one is merged. :)

This I find much easier to follow and read. Thank you!

SGSSGene (approved):

PR #1449, opened 2026-06-17
Add explicit -fexceptions copts to Bazel config

mkruskal-google:
This allows users to build with -fno-exceptions, and smoothly turns exceptions into crashes at the boundary.


SGSSGene:
Firstly, thank you for this PR.
I am not 100% sure about the use of this PR. Since Bazel is not the "prime" build system I will accept it.
If there are any issues with this PR please open a new Issue or PR.

PR #1451, opened 2026-06-23
fix: parse YAML 1.2 octal scalars (0o prefix) as integers

### LINKED ISSUE #1251: YAML octal scalars not converted to ints
The [1.2.2 YAML spec](https://yaml.org/spec/1.2.2/#tags) indicates octal literals should be formatted as "0o123" which is annoyingly not consistent with what C++ thinks octal strings should look like (0123). 
The stringstream extraction currently used expects octals to have a single leading '0', not '0o'. This is used by yaml-cpp:
https://github.com/jbeder/yaml-cpp/blob/eaf72053724814be3b99d38e292fca5797a57b7b/include/yaml-cpp/node/convert.h#L142-L155
Which ultimately means "0o123" cannot be converted to an int. "0123" on the other hand _can_ be converted to an int, which is maybe not the desired behavior if strictly adhering to the 1.2 spec.
Note this is true only as of YAML 1.2, version 1.1 specifies octals the opposite way 🙃.
Somewhat related to treatment of booleans:  #1198

raphaelroshan:
Fixes #1251.

YAML 1.2 writes octal as `0o17`, but `as<int>()` failed on it because std::stringstream's octal parsing only understands the C-style `0` prefix.

Normalize `0o`/`0O` to a leading `0` before the stream conversion, so both the 1.1 (`0123`) and 1.2 (`0o123`) spellings decode. Hex and decimal are unchanged. As per the discussion in the issue, adding a function `NormalizeOctalPrefix` that should help to  support backward compatibility. Added a test covering octal, signed/unsigned, hex and decimal.

Do let me know if i misinterpreted the behavior required!

### DIFF
--- a/include/yaml-cpp/node/convert.h
+++ b/include/yaml-cpp/node/convert.h
@@ -164,6 +164,15 @@ ConvertStreamTo(std::stringstream& stream, T& rhs) {
   }
   return false;
 }
+
+// Rewrite a valid YAML 1.2 octal scalar ("0o14") to the stream-parseable "014".
+inline std::string NormalizeOctalPrefix(std::string s) {
+  if (s.size() > 2 && s[0] == '0' && (s[1] == 'o' || s[1] == 'O') &&
+      s.find_first_not_of("01234567", 2) == std::string::npos) {
+    s.erase(1, 1);
+  }
+  return s;
+}
 }
 
 #define YAML_DEFINE_CONVERT_STREAMABLE(type, negative_op)                  \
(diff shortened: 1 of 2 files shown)


SGSSGene:
@raphaelroshan thank you so much for tackling this issue!

Comments:
1. Could you change the signature to `inline std::string NormalizeOctalPrefix(std::string s)` I believe this fits the style better.
2. I am not sure if you shuld additionally check that all characters are digits. Otherwise it will accept stuff like `0oxff` and convert it to `0xff`.
3. Sadly the [yaml spec](https://yaml.org/spec/1.2.2/#tags) is offline, but ane statements like `-0o12` or `+0o12` even valid in yaml? (this needs confirmation)

SGSSGene (changes requested):
Thanks for this PR, see comment above.

raphaelroshan:
Thank you @SGSSGene for the review, have updated the signature, added a check to ensure its an octal digit and based on yaml 1.2 - leading +/- signs arent valid so did not add that in the regex check

SGSSGene (approved):
Thanks for the PR! looks great

PR #1454, opened 2026-07-02
doc: stop directing users to stackoverflow

SGSSGene:
TLDR: Removing the `## Any Problems` section from the README.

People on stackoverflow are not as active as they used to be. Not sure if they even get help there. So I would like to remove that link.
And having a explicit statement to open a ticket on github is not necessary. 
I think the shorter the README the better.


jbeder (approved):

jbeder:
good idea

PR #1456, opened 2026-07-03
Added premake support for better Visual Studio and other build systems (like Makefile) support.

AashiqMeow12:


premake5.lua
```diff
@@ -0,0 +1,43 @@
+project "yaml-cpp"
+	kind "StaticLib"
+	systemversion "latest"
+	language "C++"
+	cppdialect "C++17"
```
SGSSGene: C++11 should be enough as a requirement.

.gitignore
```diff
@@ -1,3 +1,7 @@
 build/
 /tags
 /bazel-*
+bin
+bin-int
+yaml-cpp.vcxproj
+yaml-cpp.vcxproj.filters
```
SGSSGene: Please do not change the .gitignore list. I do not think these changes are necessary.

premake5.lua
```diff
@@ -0,0 +1,43 @@
...
+		runtime "Debug"
+		symbols "on"
+
+	filter "configurations:Release"
+		runtime "Release"
+		optimize "on"
+
+	filter "configurations:Dist"
+		runtime "Release"
+		optimize "on"
```
SGSSGene: I believe there is a new line missing at the end of the file.

premake5.lua
```diff
@@ -0,0 +1,43 @@
...
+	staticruntime "on"
+
+	targetdir ("bin/" .. outputdir .. "/%{prj.name}")
+	objdir ("bin-int/" .. outputdir .. "/%{prj.name}")
+
+	files
+	{
+		"src/**.h",
+		"src/**.cpp",
+
```
SGSSGene: There are two tabs not are not needed.

SGSSGene (changes requested):
@AashiqMeow12 thank you for your PR.

Could you change everything to two-spaces indentation instead of tabs. This makes it inline with all the other files in this project.

@jbeder : Do we want or are willing to accept more build systems?

jbeder:
I'm pretty sure CMake generates these. If so, we should just defer to CMake. I made the exception for bazel because it really is fundamentally different and I actually like its fundamental difference.

premake5.lua
AashiqMeow12: So 4-spaces?

premake5.lua
AashiqMeow12: **Changed**

.gitignore
AashiqMeow12: Removed:
bin
bin-int
yaml-cpp.vcxproj
yaml-cpp.vcxproj.filters

back!

premake5.lua
AashiqMeow12: # I generally don't like new lines at the end of files, but your choice!

**CHANGED**

AashiqMeow12:
> I'm pretty sure CMake generates these. If so, we should just defer to CMake. I made the exception for bazel because it really is fundamentally different and I actually like its fundamental difference.

`I just like premake's simplicity.`
**I know CMake can do this, but some people use premake, and I personally think adding premake is not a big deal.**
**I also use premake in my project, from where I requested this change**

# THE DECISION IS YOURS!

PR #1457, opened 2026-07-04
feat: add thread-safe contains() function for Node.

### LINKED ISSUE #1092: Checking for key without mutation
Hi.  I found it lacking that there is no interface to check for the existence of a key on a YAML Node.
The only way to do it at the moment is to call operator[]() with the key, and check if the result is valid.  Although this works, it looks like it requires a mutation ( so its not really const ) and therefore cant be done safely across threads.
Would it be possible to add a purely read-only hasKey-like method?

### LINKED ISSUE #1350: [Crash] Segmentation Fault/Double Free in Multithreaded Access to Undefined YAML::Node via IsDefined() and reset()
**Problem Description**  
When using yaml-cpp in multithreaded scenarios to check undefined nodes via `IsDefined()` and `reset()` methods, the program occasionally crashes with segmentation faults or double-free errors. The crash occurs even with **read-only operations** and no explicit node modifications.

---

**Minimal Reproduction**  
```cpp
#include <yaml-cpp/yaml.h>
#include <thread>
#include <vector>

YAML::Node rootNode(YAML::NodeType::Map);

YAML::Node foo() {
  YAML::Node node;
  node.reset(rootNode); 
  if (!node.IsDefined()) {
    return node;
  }
  node.reset(node["a"]);   // undefined
  if (!node.IsDefined()) {
    return node;
  }
  return node;
}

int main() {
  std::vector<std::thread> threads;
  for (int i = 0; i < 16; ++i) {
    threads.emplace_back(foo);
  }
  for (auto& t : threads) {
    t.join();
  }
}
```

**Reproduction Script**  
```bash
#!/bin/bash
PROGRAM="./bazel-bin/src/main"
if [ ! -x "$PROGRAM" ]; then
    echo "Error: Program '$PROGRAM' not found or not executable."
    exit 1
fi
echo "Running $PROGRAM until crash occurs..."
count=0
while true; do
    count=$((count+1))
    $PROGRAM
    exit_code=$?
    if [ $exit_code -ne 0 ]; then
        echo "Program finished with exit code $exit_code"
        break
    fi
done
echo "Program crashed after $count runs"
```

---

**Discovery Context**  
The crash manifests in real-world code resembling this pattern:
```cpp
YAML::Node GetNode(std::string_view keyPath) {
  std::vector<std::string> keys = split(keyPath, ".");
  YAML::Node node;
  node.reset(rootNode);
  for (const auto& key : keys) {
    if (!node) break;
    node.reset(node[key]);
  }
  return node; 
}
```
This pattern is used extensively for configuration parsing. Despite being read-only and not modifying any nodes, concurrent access causes crashes.

---

**Requests**  
1. Is this a known thread-safety limitation of yaml-cpp's reference-counted nodes?
2. What is the recommended pattern for concurrent read-only access to YAML configurations?

sxrzh:
Add `contains()` function to check if `node[key]` exists.
Close #1350 and close #1092.

### DIFF
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -394,6 +394,12 @@ inline void Node::force_insert(const Key& key, const Value& value) {
   m_pNode->force_insert(key, value, m_pMemory);
 }
 
+template <typename Key>
+inline bool Node::contains(const Key& key) const {
+  EnsureNodeExists();
+  return ((const detail::node*)m_pNode)->get(key, m_pMemory) != nullptr;
+}
+
 // free functions
 inline bool operator==(const Node& lhs, const Node& rhs) { return lhs.is(rhs); }
 }  // namespace YAML
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -396,8 +396,10 @@ inline void Node::force_insert(const Key& key, const Value& value) {
 
 template <typename Key>
 inline bool Node::contains(const Key& key) const {
-  EnsureNodeExists();
-  return ((const detail::node*)m_pNode)->get(key, m_pMemory) != nullptr;
+  if (!m_isValid)
+    throw InvalidNode(m_invalidKey);
+  if (!m_pNode) return false;
+  return (static_cast<const detail::node*>(m_pNode))->get(key, m_pMemory) != nullptr;
 }
 
 // free functions
--- a/include/yaml-cpp/node/node.h
+++ b/include/yaml-cpp/node/node.h
@@ -114,6 +114,9 @@ class YAML_CPP_API Node {
   template <typename Key, typename Value>
   void force_insert(const Key& key, const Value& value);
 
+  template <typename Key>
+  bool contains(const Key& key) const;
+
  private:
   enum Zombie { ZombieNode };
   explicit Node(Zombie);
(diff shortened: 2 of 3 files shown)


sxrzh:
Note that for a sequence node and an integer within [0, size - 1] as the key, `contains` still returns true. I think this is reasonable.

include/yaml-cpp/node/impl.h
```diff
@@ -394,6 +394,12 @@ inline void Node::force_insert(const Key& key, const Value& value) {
   m_pNode->force_insert(key, value, m_pMemory);
 }
 
+template <typename Key>
+inline bool Node::contains(const Key& key) const {
+  EnsureNodeExists();
+  return ((const detail::node*)m_pNode)->get(key, m_pMemory) != nullptr;
+}
```
SGSSGene, line 403: This is not thread safe. The problem is that the function `EnusreNodeExists` is not thread-safe. Even though it is marked as 'const' is manipulates the member variable `m_pMemory` which is marked as `mutable`.

I think instead of calling `EnsureNodeExists` it should do the check it self and return `false` if the node doesn't exists.

SGSSGene (changes requested):
Thank you @sxrzh! I think this is a much needed function!

The current way is not thread-safe.
Also could you instead of a C-Cast use a `static_cast` as Node::operator[] uses it?

Your PR is highly appreciated!

PR #1459, opened 2026-07-10
fix: emit uint8_t as an integer

### LINKED ISSUE #1245: `uint8_t` not emit as an integer using YAML::Emitter
# Bug description
With yaml-cpp version 0.8.0, when I use `YAML::Emitter` to emit a `uint8_t` in a yaml document, the result is different from other integer types. It seems that it is emitted in a quoted hexadecimal representation. 

A side effect of this is that a `BadConversion` exception is thrown when you try to convert a `uint8_t` emitted with `YAML::Emitter`.

Might be link with #1081 

# Reproduction
## Test code
```cpp
#include <yaml-cpp/yaml.h>
#include <iostream>
#include <cstdint>

template<typename Int>
void test(Int integer)
{
  YAML::Emitter out;
  out << integer;
  std::cout << out.c_str() << std::endl;
}

int main()
{
  // Comparison with other interger types
  int8_t int8 {1};
  int16_t int16 {1};
  int32_t int32 {1};
  int64_t int64 {1};

  uint8_t uint8 {16};
  uint16_t uint16 {1};
  uint32_t uint32 {1};
  uint64_t uint64 {1};

  test(int8);
  test(int16);
  test(int32);
  test(int64);
  test(uint8);
  test(uint16);
  test(uint32);
  test(uint64);

  // BadConversion exception
  YAML::Emitter out;
  out << uint8;
  try
  {
    auto test = YAML::Load(out.c_str()).as<uint8_t>();
  }
  catch (const YAML::BadConversion &e)
  {
    std::cout << e.what() << std::endl;
  }

  return 0;
}
```

## Result
With yaml-cpp version 0.8.0
```
1
1
1
1
"\x10"
1
1
1
yaml-cpp: error at line 1, column 1: bad conversion
```

hydraxman:
## Summary
- Emit `uint8_t`/`unsigned char` through the integral emitter path instead of treating the value as a character.
- Keep the existing plain `char` behavior unchanged and add a regression test for numeric output.

## Reproduction
- Before: `./build/test/yaml-cpp-tests --gtest_filter=EmitterTest.UnsignedEightBitInteger` → expected `- 16`, got `- "\\x10"`.
- After: the same command passes and emits `- 16`.

## Tests
- `./build/test/yaml-cpp-tests --gtest_filter=EmitterTest.UnsignedEightBitInteger`
- `./build/test/yaml-cpp-tests` (1027 tests passed; 8 disabled)

Fixes #1245

### DIFF
--- a/include/yaml-cpp/emitter.h
+++ b/include/yaml-cpp/emitter.h
@@ -234,7 +234,7 @@ inline Emitter& operator<<(Emitter& emitter, char v) {
   return emitter.Write(v);
 }
 inline Emitter& operator<<(Emitter& emitter, unsigned char v) {
-  return emitter.Write(static_cast<char>(v));
+  return emitter.WriteIntegralType(static_cast<unsigned int>(v));
 }
 inline Emitter& operator<<(Emitter& emitter, const _Alias& v) {
   return emitter.Write(v);
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -119,7 +119,7 @@ TEST_F(EmitterTest, IntBase) {
   ExpectEmit("- 1024\n- 0x400\n- 02000");
 }
 
-TEST_F(EmitterTest, EightBitIntegers) {
+TEST_F(EmitterTest, UnsignedEightBitInteger) {
   out << BeginSeq;
   out << std::uint8_t{16};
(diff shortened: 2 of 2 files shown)


SGSSGene (approved):
Thank you for this PR! Looks great!

PR #1460, opened 2026-07-10
fix: preserve original scalar text for null nodes so as<string> returns the source text

### LINKED ISSUE #1290: Empty strings are silently replaced with 'null' now
I've been upgrading from an older yaml-cpp version to 0.8.0.
What I found different is that for empty strings, behavior has changed, they are now replaced with `"null"`.

For example, given the following YAML:

```yaml
message:
```

```cpp
node["message"].as<std::string>()  //=> "null"
```

I know that such a value can be interpreted as `null`, hence `IsNull() -> true` is expected.

However, my parsing for this particular option does not allow for nulls, I always interpret this as a string. In my case `message=""` vs `message="null"` makes a difference, and I can't easily distinguish between them.

Can yaml-cpp revert to the old behavior in this case?
For an empty value, I want `Scalar() == as<std::string>() == ""`
For `null` value, I want `Scalar() == as<std::string>() == "null"`

mvanhorn:
## Summary
Plumb the original token text through the null path without breaking the public EventHandler API: add a non-pure virtual `OnNull(const Mark&, anchor_t, const std::string& value)` overload to include/yaml-cpp/eventhandler.h whose default implementation delegates to the existing two-arg `OnNull`, so third-party handlers and the emitter/graphbuilder adapters keep working unchanged. `SingleDocParser::HandleNode` passes `token.value` at the null-plain-scalar site; `NodeBuilder` overrides the three-arg `OnNull` and calls a new `node_data::set_null(const std::string& scalar)` overload (include/yaml-cpp/node/detail/node_data.h + src/node_data.cpp) that sets `m_type = NodeType::Null` while storing the source text in `m_scalar`, exposed via the existing `Scalar()` accessor. Finally, the two `as_if<std::string>` specializations in include/yaml-cpp/node/impl.h return `node.Scalar()` for Null nodes instead of the literal `"null"`, so `key:` yields `""`, `key: ~` yields `"~"`, and `key: null` yields `"null"`, while `IsNull()` stays true for all of them.

## Why this matters
Since yaml-cpp 0.8.0, any null-like plain scalar (`<empty>`, `~`, `null`, `Null`, `NULL`) loses its original text: `SingleDocParser::HandleNode` (src/singledocparser.cpp, ~line 101) detects `IsNullString(token.value...)` and emits `OnNull(mark, anchor)`, discarding `token.value`. `NodeBuilder::OnNull` then calls `node.set_null()`, which leaves `m_scalar` empty, and `as_if<std::string>` in include/yaml-cpp/node/impl.h (~lines 119 and 151) hard-codes the return value `"null"` for `NodeType::Null`. The reporter cannot distinguish `message:` (empty) from `message: null` because both `Scalar()` and `as<std::string>()` return fabricated values. Owner jbeder agreed the original text representation should be kept in `Scalar()` and returned by `as<std::string>()` ("Yeah that's reasonable").

See https://github.com/jbeder/yaml-cpp/issues/1290.

## Testing
- Happy path: load `message: null` and assert `IsNull()` is true, `Scalar() == "null"`, and `as<std::string>() == "null"` (test/integration/load_node_test.cpp). - Variants: `key:` yields `""`, `key: ~` yields `"~"`, `key: Null` / `key: NULL` yield their exact source text; quoted `"null"` remains a plain string scalar unaffected. - Edge cases: flow-map and block-seq missing values (structural nulls with no token) return `""` from `as<std::string>()` and stay `IsNull()`; a default-constructed `Node()` still reports Null with empty scalar (test/node/node_test.cpp). - Error paths: `as<std::string>()` on a map/sequence still throws `TypedBadConversion`; `as<std::string>(fallback)` on non-scalar non-null still returns the fallback.

Fixes #1290

### DIFF
--- a/include/yaml-cpp/eventhandler.h
+++ b/include/yaml-cpp/eventhandler.h
@@ -23,6 +23,10 @@ class EventHandler {
   virtual void OnDocumentEnd() = 0;
 
   virtual void OnNull(const Mark& mark, anchor_t anchor) = 0;
+  virtual void OnNull(const Mark& mark, anchor_t anchor,
+                      const std::string& /*value*/) {
+    OnNull(mark, anchor);
+  }
   virtual void OnAlias(const Mark& mark, anchor_t anchor) = 0;
   virtual void OnScalar(const Mark& mark, const std::string& tag,
                         anchor_t anchor, const std::string& value) = 0;
--- a/include/yaml-cpp/node/detail/node.h
+++ b/include/yaml-cpp/node/detail/node.h
@@ -82,6 +82,10 @@ class node {
     mark_defined();
     m_pRef->set_null();
   }
+  void set_null(const std::string& scalar) {
+    mark_defined();
+    m_pRef->set_null(scalar);
+  }
   void set_scalar(const std::string& scalar) {
     mark_defined();
     m_pRef->set_scalar(scalar);
--- a/include/yaml-cpp/node/detail/node_data.h
+++ b/include/yaml-cpp/node/detail/node_data.h
@@ -38,6 +38,7 @@ class YAML_CPP_API node_data {
   void set_type(NodeType::value type);
   void set_tag(const std::string& tag);
   void set_null();
+  void set_null(const std::string& scalar);
   void set_scalar(const std::string& scalar);
   void set_style(EmitterStyle::value style);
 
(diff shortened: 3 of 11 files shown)


mvanhorn:
Addressed in 30eade1. Full CTest suite (1029 tests) passes locally.

PR #1461, opened 2026-07-11
fix: allow '~' in emitted anchor and alias names

raphaelroshan:
`IsAnchorChar` (the emitter's `ns-anchor-char` check) rejects `0x7E` (`~`) because of an off-by-one: it tests `ch < 0x7E` before falling through to the DEL/C1 exclusion. But `~` is a valid `ns-char` (YAML 1.2 printable range `0x21–0x7E`) and is not a flow indicator, so it's a legal anchor-name character — and the scanner's anchor rule (`Exp::Anchor()`) already accepts it.

The result is an asymmetry: `&anchor~ x` parses fine but the emitter can't write it back out, failing with `INVALID_ANCHOR` — so yaml-cpp fails to round-trip its own valid input.

Fix: `ch < 0x7E` → `ch <= 0x7E`. DEL (`0x7F`) and the C1 controls (`0x80–0x9F`) remain excluded. Added `EmitterTest.AnchorWithTilde` (fails before, passes after).

### DIFF
--- a/src/emitterutils.cpp
+++ b/src/emitterutils.cpp
@@ -39,7 +39,7 @@ bool IsAnchorChar(int ch) {  // test for ns-anchor-char
     return false;
   }
 
-  if (ch < 0x7E) {
+  if (ch <= 0x7E) {
     return true;
   }
 
--- a/test/integration/emitter_test.cpp
+++ b/test/integration/emitter_test.cpp
@@ -535,6 +535,15 @@ TEST_F(EmitterTest, AliasAndAnchor) {
   ExpectEmit("- &fred\n  name: Fred\n  age: 42\n- *fred");
 }
 
+TEST_F(EmitterTest, AnchorWithTilde) {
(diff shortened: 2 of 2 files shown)


SGSSGene:
Nice find! Could you provida a link to the yaml-specs which defines the printable range?

raphaelroshan:
Sure —   https://yaml.org/spec/1.2.2/#character-set
  
`~` (U+007E) sits in the printable range  `[#x20-#x7E]` of the `c-printable`  production so it's a valid `ns-char`. Anchor names are `ns-anchor-char`, defined as `ns-char` minus the flow indicators `,[]{}` ([https://yaml.org/spec/1.2.2/#692-node-anchors](https://yaml.org/spec/1.2.2/#692-node-anchors)), and `~` isn't a flow indicator, so it's valid in an anchor name.

SGSSGene (approved):
Fantastic! Thank you!

PR #1465, opened 2026-07-18
Avoid production source glob checks in no-op builds

### LINKED ISSUE #1463: Globs trigger unnecessary CMake re-run when rebuilding
My code uses yaml-cpp via `add_subdirectory` and I was attempting to speed-up `cmake --build . --parallel` for the case when nothing has changed.   One of the things that came up was that yaml-cpp consistently triggers a cmake re-run,

```
ninja explain: .../VerifyGlobs.cmake_force is dirty
[0/2] Re-checking globbed directories...
[1/2] Re-running CMake...
```

Would there be interest in avoiding this by eliminating the Glob?   I see #1349 , but it looks like it's over a year old without any feedback, so wanted to open up a discussion here.

```
-set(contrib-pattern "src/contrib/*.cpp")
-set(src-pattern "src/*.cpp")
-if (CMAKE_VERSION VERSION_GREATER 3.12)
-  list(INSERT contrib-pattern 0 CONFIGURE_DEPENDS)
-  list(INSERT src-pattern 0 CONFIGURE_DEPENDS)
-endif()
-
-file(GLOB yaml-cpp-contrib-sources ${contrib-pattern})
-file(GLOB yaml-cpp-sources ${src-pattern})
+set(yaml-cpp-contrib-sources
+  src/contrib/graphbuilder.cpp
+  src/contrib/graphbuilderadapter.cpp)
+
+set(yaml-cpp-sources
+  src/binary.cpp
+  src/convert.cpp
+  src/depthguard.cpp
+  src/directives.cpp
+  src/emit.cpp
+  src/emitfromevents.cpp
+  src/emitter.cpp
+  src/emitterstate.cpp
+  src/emitterutils.cpp
+  src/exceptions.cpp
+  src/exp.cpp
+  src/fptostring.cpp
+  src/memory.cpp
+  src/node.cpp
+  src/node_data.cpp
+  src/nodebuilder.cpp
+  src/nodeevents.cpp
+  src/null.cpp
+  src/ostream_wrapper.cpp
+  src/parse.cpp
+  src/parser.cpp
+  src/regex_yaml.cpp
+  src/scanner.cpp
+  src/scanscalar.cpp
+  src/scantag.cpp
+  src/scantoken.cpp
+  src/simplekey.cpp
+  src/singledocparser.cpp
+  src/stream.cpp
+  src/tag.cpp)
```

floze-the-genius:
## Summary

- replace the production `CONFIGURE_DEPENDS` source globs with an explicit source list
- add a CTest check that compares the explicit list with the C++ files under `src/`
- keep test source discovery unchanged

This prevents consumers that use `add_subdirectory` from running CMake's glob verification on every no-op build, while the regression test catches newly added production sources that are missing from the explicit list.

Fixes #1463.

## Validation

- Ubuntu 24.04, GCC 13, CMake, and Ninja: configured with `YAML_CPP_BUILD_TESTS=ON`, built all targets, and passed both CTest tests
- configured with tests and tools disabled: the second Ninja build reported `ninja: no work to do.` and generated no `VerifyGlobs.cmake`
- external `add_subdirectory` consumer with `YAML_CPP_BUILD_CONTRIB=OFF`: built successfully and had a clean second no-op build
- negative regression check: an unlisted temporary `src/forgotten.cpp` caused `yaml-cpp::source-list` to fail as intended

## AI assistance disclosure

OpenAI Codex was used as a coding tool to inspect the repository, implement this change, and run validation. The patch was checked with the validation listed above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,8 +55,7 @@ if (NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
     MultiThreaded$<$<CONFIG:Debug>:Debug>$<${msvc-shared_rt}:DLL>)
 endif()
 
-file(GLOB yaml-cpp-contrib-sources CONFIGURE_DEPENDS "src/contrib/*.cpp")
-file(GLOB yaml-cpp-sources CONFIGURE_DEPENDS "src/*.cpp")
+include("${CMAKE_CURRENT_LIST_DIR}/cmake/yaml-cpp-sources.cmake")
 
 set(msvc-rt $<TARGET_PROPERTY:MSVC_RUNTIME_LIBRARY>)
 
--- a/cmake/yaml-cpp-sources.cmake
+++ b/cmake/yaml-cpp-sources.cmake
@@ -1,0 +1,35 @@
+set(yaml-cpp-contrib-sources
+  src/contrib/graphbuilder.cpp
+  src/contrib/graphbuilderadapter.cpp)
+
+set(yaml-cpp-sources
+  src/binary.cpp
+  src/convert.cpp
+  src/depthguard.cpp
+  src/directives.cpp
+  src/emit.cpp
+  src/emitfromevents.cpp
+  src/emitter.cpp
+  src/emitterstate.cpp
+  src/emitterutils.cpp
+  src/exceptions.cpp
+  src/exp.cpp
+  src/fptostring.cpp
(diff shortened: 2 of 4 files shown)


andrewcorrigan:
Thank you for fixing this!

SGSSGene (approved):
Very nice, Thank you!

PR #1468, opened 2026-07-26
fix: remove unverified potential simple key when beginning parsing a literal scalar

### LINKED ISSUE #1163: Loading of block sequence with tag fails
Loading of block sequence with tag (standard or application) and less than 3 spaces (at the line begin) fails with exception: "end of sequence not found".
Such YAML is declared valid with number of online validators.

### More details and examples

All following examples are successfully loaded

1.  not block sequence
```
key: !t |
 multiple
 words
```

2. no tag
```
key: 
- |
 multiple
 words
```

3. more than 2 spaces (3)
```
key: 
- !!str |
   multiple
   words
```

but the following **fail** (**1 or 2 spaces** before "multiple" and "words"):

1.  application tag
```
key: 
- !t |
 multiple
 words
```

2. standard tag
```
key: 
- !!str |
  multiple
  words
```

### Possible solution

The change in `Scanner::ScanTag`:
- remove `InsertPotentialSimpleKey();` call

seems to solve this problem, but breaks some of existing tests.

sxrzh:
fixes #1163

### DIFF
--- a/src/scanner.cpp
+++ b/src/scanner.cpp
@@ -168,6 +168,17 @@ void Scanner::ScanNextToken() {
   // special scalars
   if (InBlockContext() && (INPUT.peek() == Keys::LiteralScalar ||
                            INPUT.peek() == Keys::FoldedScalar)) {
+    // if we begin parsing a literal scalar with an unverified potential 
+    // simple key pushed, that may be a tag to the literal scalar, and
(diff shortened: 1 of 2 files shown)


SGSSGene (approved):
Very cool! Thank you!

PR #1469, opened 2026-07-27
Add some missing constexpr on static constants

jwnimmer-tri:
This avoids the Static Initialization (and Destruction) Order Fiasco.

https://en.cppreference.com/cpp/language/siof
https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

### DIFF
--- a/src/binary.cpp
+++ b/src/binary.cpp
@@ -3,7 +3,7 @@
 #include <cctype>
 
 namespace YAML {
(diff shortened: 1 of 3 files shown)


SGSSGene (approved):
Sure, why not, I do not see the benefit, but also isn't wrong. So will accept.

PR #1470, opened 2026-07-27
Fix version declaration in MODULE.bazel

jwnimmer-tri:
Amends 56e3bb550c91fd7005566f19c079cb7a503223cf.


SGSSGene (approved):
Uff! What a stupid small bug!
Thank you for reporting and fixing!

Issue #1281, opened 2024-05-17
YAML Emitter using YAML::DoubleQuoted when emitted in a destructor of a staticly created struct/class


Krausler:
**What steps will reproduce the problem?**

1. Create a struct/class containing a destructor and a function to serialize an arbitrary type of data, with the destructor calling the serialization function.

2. Create a static instance of that class and call the serilization function at least once.

**What is the expected output? What do you see instead?**
The expected output would be to see the serialized data, as many times as requested, beeing exactly the same, however, the last time the yaml string is printed, all keys, as well as all strings, are wrapped in double quotes.

**Example**
```
struct EmitterTest
{
    ~EmitterTest()
    {
        Serialize();
    }

    void Serialize()
    {
        YAML::Emitter out;

        out << YAML::BeginMap << YAML::Key << "Test";
        out << YAML::BeginMap;
        out << YAML::Key << "String1" << YAML::Value << "String1";
        out << YAML::Key << "String2" << YAML::Value << "String2";
        out << YAML::Key << "String3" << YAML::Value << "Foo";
        out << YAML::EndMap;
        out << YAML::EndMap;

        std::cout << out.c_str() << std::endl;
    }
}

static EmitterTest s_Test;

int main()
{
    s_Test.Serialize();
}
```
Expected output:
```
Test:
  String1: String1
  String2: String2
  String3: Foo
Test:
  String1: String1
  String2: String2
  String3: Foo
```

Actual output:
```
Test:
  String1: String1
  String2: String2
  String3: Foo
"Test":
  "String1": "String1"
  "String2": "String2"
  "String3": "Foo"
```

**yaml-cpp version: 0.8.0**

**Additional Information**
Note that this only happens when Serialize() is called once before destruction. Digging into the problem makes the problem pretty obvious: In emitter.h the function YAML::Emitter::Write(const std::string&) queries the StringFormat::value, by trying to match the string with a regex.This regex match returns false in the serialization called in the destructor, otherwise it returns true.

I'm not quite sure if this is a bug and reproducable or just me beeing stupid :)

PR #1471, opened 2026-07-27
Avoid static destructors during program exit

jwnimmer-tri:
Function-local static variables must either be marked constexpr or never_destroyed. Otherwise, their order of destruction during program exit is indeterminate and can lead to undefined behavior.

https://en.cppreference.com/cpp/language/siof
https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

### DIFF
--- a/src/emitterutils.cpp
+++ b/src/emitterutils.cpp
@@ -174,14 +174,14 @@ bool IsValidPlainScalar(const char* str, std::size_t size, FlowType::value flowT
   }
 
   // then check until something is disallowed
-  static const RegEx disallowed_flow =
+  static const never_destroyed<RegEx> disallowed_flow{
       Exp::EndScalarInFlow() | (Exp::BlankOrBreak() + Exp::Comment()) |
       Exp::NotPrintable() | Exp::Utf8_ByteOrderMark() | Exp::Break() |
-      Exp::Tab() | Exp::Ampersand();
-  static const RegEx disallowed_block =
+      Exp::Tab() | Exp::Ampersand()};
+  static const never_destroyed<RegEx> disallowed_block{
       Exp::EndScalar() | (Exp::BlankOrBreak() + Exp::Comment()) |
       Exp::NotPrintable() | Exp::Utf8_ByteOrderMark() | Exp::Break() |
-      Exp::Tab() | Exp::Ampersand();
+      Exp::Tab() | Exp::Ampersand()};
   const RegEx& disallowed =
       flowType == FlowType::Flow ? disallowed_flow : disallowed_block;
 
--- a/src/exp.h
+++ b/src/exp.h
@@ -10,6 +10,7 @@
 #include <ios>
 #include <string>
 
+#include "never_destroyed.h"
 #include "regex_yaml.h"
 #include "stream.h"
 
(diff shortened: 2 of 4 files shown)


SGSSGene (commented):
I think this PR needs some discussion.

What I understand is:
The extra `never_destroyed` template makes the data types "leak" their data on the heap by not calling `delete` and by this avoid wrong destruction order.

I think this is quiet some source code overhead. It avoid a theoretical issue that does not occur during execution and it will (hopefully) trigger memory leak detection (address sanitizer or valgrind).

So I am positioning myself against this PR.

SGSSGene:
Just saw that issue #1281 having the exact trouble this PR is trying to fix.

jwnimmer-tri:
> The extra never_destroyed template makes the data types "leak" their data on the heap by not calling delete and by this avoid wrong destruction order.

Yes in terms of disabling the destructor; no in terms of _directly_ leaking onto the heap.  The variable wrapped in `never_destroyed` is still stored statically, not on the heap.

For example the `static const never_destroyed<std::string> svalue;`, the `std::string` is static, and because it's an empty string it doesn't have a `c_str()` that points into heap-allocated storage.

Even for `RegEx` such as `static const never_destroyed<RegEx> e{RegEx(' ')};` for `Space()` there is no heap use since the `m_params` is empty.

For compound regex e.g. `Blank()`, yes the `static const never_destroyed<RegEx> e{Space() | Tab()};` does end up using heap storage for the `m_params` (just like it does today on `master`), and yes that heap storage is never free'd, but ...

> ... and it will (hopefully) trigger memory leak detection (address sanitizer or valgrind).

No.  These sanitizers are smart enough to recognize heap storage that is "still reachable" vs "definitely leaked".  It is not uncommon for programs to have global variables which are never free'd.  So long as the global variable is still reachable from the static storage rootset, it is not a leak.  Is is marked as "still reachable", instead.

> I think this is quiet some source code overhead.

Yes, unfortunately C++ has the problem that there is no great way to spell this.  I would love it if someone wrote a paper to get `std::never_destroyed` into the standard library.

The only way with less code overhead is doing e.g. `static const std::string* const  svalue = new std::string();` which avoids the need for a template wrapper by directly calling `new` at every call site, and purposefully never calling `delete`.  This is less efficient (by forcing the global variable onto the heap, removing the opportunity for the compiler to elide the constructor when possible), and leads to IMO confusing code (begging the question of why it is not being deleted).

> It avoid a theoretical issue that does not occur during execution ...

As you noted with the other issue cross-reference, at scale this becomes a real problem.  With only a small number of destructors you can often skate by unharmed, but once you have enough of them it's too easy to end up with crashes, and sometimes crashes that don't show up until production.  For me, the tradeoff of the extra noise in the code is worth it to make the software bulletproof.

SGSSGene:
> > The extra never_destroyed template makes the data types "leak" their data on the heap by not calling delete and by this avoid wrong destruction order.
> 
> Yes in terms of disabling the destructor; no in terms of _directly_ leaking onto the heap. The variable wrapped in `never_destroyed` is still stored statically, not on the heap.

I am sorry, I went through this PR way to fast! You are using placment-new! that indeed does not leak in anyway that I would be concerned. Additionally, the class is mostly used inside free-standing functions which are probably used to fix the initialization order issue, but for this PR they also provide great encapsulation and shield against all othercode.

This makes this PR a great solution. I will take a detailed look in my next review session!

alex-thiessen-for-siemens:
FYI I am working on switching to compile-time RegEx construction which should partially address this, and also improve performance a bit.

SGSSGene:
> FYI I am working on switching to compile-time RegEx construction which should partially address this, and also improve performance a bit.

wow! that is cool!

As an addition to this PR or an alternative?

alex-thiessen-for-siemens:
It's going to be a separate PR because it only addresses the `RegEx` part of this PR, and the switch to `ConstRegEx` is somewhat involved (some of which is due to the ABI preservation stance which might or might not be feasible).

A preview is here: https://github.com/alex-thiessen-for-siemens/yaml-cpp/commits/fork/fix/regex-constexpr/

PR #1472, opened 2026-07-28
Add LICENSE file to Bazel

jwnimmer-tri:
Bazel users cannot access this file (e.g., to redistribute it) unless it's explicitly marked public.


SGSSGene (approved):
A very good!
I assume you are using Bazel and know how to use it correctly!

Thank you very much!

jwnimmer-tri:
Yes, thanks!  I'm the maintainer for yaml-cpp in the Bazel Central Registry https://github.com/bazelbuild/bazel-central-registry/pull/8524.

PR #1474, opened 2026-08-01
feat: reverse_iterator support

sxrzh:
Adds `reverse_iterator` support for `YAML::Node`, including:
- `Node::reverse_iterator` and `Node::const_reverse_iterator`
- `Node::rbegin()` and `Node::rend()`, including const versions

I didn't do the same for `node_iterator`, for I think node iterators won't be used directly and there're no systematic tests for node iterators.

Currently `iterator_base` and `node_iterator_base` are incompatible with `std::reverse_iterator`, since their `operator->()` function returns a proxy object, which could cause dangling pointer with `std::reverse_iterator`. So I implemented a customized `reverse_iterator_base` wrapper in yaml-cpp/node/detail/reverse_iterator.h .

### DIFF
--- a/include/yaml-cpp/node/detail/iterator.h
+++ b/include/yaml-cpp/node/detail/iterator.h
@@ -9,6 +9,7 @@
 
 #include "yaml-cpp/dll.h"
 #include "yaml-cpp/node/detail/node_iterator.h"
+#include "yaml-cpp/node/detail/reverse_iterator.h"
 #include "yaml-cpp/node/node.h"
 #include "yaml-cpp/node/ptr.h"
 #include <cstddef>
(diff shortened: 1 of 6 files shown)


sxrzh:
> So I implemented a customized reverse_iterator_base wrapper in yaml-cpp/node/detail/reverse_iterator.h .

Note that this does not meet the requirements for [*ReversibleContainer*](https://en.cppreference.com/cpp/named_req/ReversibleContainer). Since the current YAML::Node is also far away from [*Container*](https://en.cppreference.com/cpp/named_req/Container) (maybe it is never intended to be), I think this is acceptable.

SGSSGene (approved):
Amazing! cool feature!

PR #1476, opened 2026-08-07
fix: fix stack underflow of PopIndent

### LINKED ISSUE #1475: [Bug]Scanner indent-stack underflow: PopIndent/PushIndentTo dereference *m_indents.top() on an empty stack (wild read, deterministic crash / remote DoS) (CWE-822)
**Describe the bug**

`YAML::Scanner` maintains an indentation stack (`m_indents`). When a malformed YAML document makes the scanner expect an indent that was never pushed, `YAML::Scanner::PopIndent` (src/scanner.cpp:396) and `YAML::Scanner::PushIndentTo` (src/scanner.cpp:330) call `m_indents.top()` on an **empty** `std::deque` — undefined behavior that reads a wild pointer. Under ASan/UBSan this is a deterministic crash (SEGV / UB trap); in a release build it is an out-of-bounds read on the stack container (DoS, potential info leak). A 96-byte YAML document triggers it deterministically.

**To Reproduce**

1. Clone yaml-cpp and check out the affected commit:

   ```bash
   git clone https://github.com/jbeder/yaml-cpp.git
   cd yaml-cpp
   git checkout 3eb39d5808c33aa93a113cb6c668fa2435ff3ecd
   ```

2. Build the static library with ASan + UBSan:

   ```bash
   mkdir -p build && cd build
CXX=clang++ cmake -DYAML_CPP_BUILD_TESTS=OFF -DYAML_CPP_BUILD_TOOLS=OFF \
         -DCMAKE_CXX_FLAGS="-fsanitize=address,undefined -fno-sanitize-recover=all -fno-omit-frame-pointer -g" ..
   make -j"$(nproc)" yaml-cpp
   cd ..
   ```

3. Save this harness as `harness.cpp` (feeds the YAML bytes to `YAML::Load`):

```cpp
// harness.cpp - libFuzzer harness for yaml-cpp YAML::Load(const std::string&)
//
// Selected target (rank 1 from target_functions.json):
//   YAML_CPP_API Node Load(const std::string& input)   [src/parse.cpp:12]
//
// Fuzz contract:
//   * The fuzz input bytes are the raw YAML document. We copy them verbatim
//     into a std::string using the (const char*, size) constructor so NUL bytes
//     and binary scalars survive (the const char* overload would truncate at
//     the first NUL). The std::string holds EXACTLY `size` bytes - no padding,
//     no slack. The library reads the entire buffer as the document.
//   * YAML::Load is a pure function: no global state, no databases, no file
//     handles. Therefore LLVMFuzzerInitialize performs no setup.
//   * All malformed-input exceptions (ParserException, DeepRecursion, BadFile)
//     are derived from YAML::Exception (std::runtime_error). They MUST be
//     caught here, otherwise a malformed fuzz input would terminate the fuzzer
//     process and be mis-reported as a crash.
//   * The returned Node (and its full alias/anchored node graph) is kept in
//     scope until the end of LLVMFuzzerTestOneInput, so construction and
//     destruction of the parsed graph are both exercised per input.
//
// Allocation-fidelity note:
//   The only allocation the harness controls is `input`, sized exactly `size`.
//   yaml-cpp's parser does its own internal allocations; no pre-sized output
//   buffer is passed in, so there is no buffer contract to over/under-allocate.
//   ASan/UBSan will catch any out-of-bounds access at the library's real
//   boundaries.

#include <cstddef>
#include <cstdint>
#include <exception>
#include <string>

#include "yaml-cpp/yaml.h"  // full YAML::Node definition + Load declarations + exceptions

extern "C" int LLVMFuzzerInitialize(int* argc, char*** argv) {
  // No one-time setup required: YAML::Load is stateless. This stub is provided
  // so the build can rely on the symbol existing even if a later iteration of
  // this harness needs global state.
  (void)argc;
  (void)argv;
  return 0;
}

extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
  if (size == 0) {
    return 0;  // empty document is uninteresting; skip
  }

  // Exact-length copy of the fuzz buffer. NULs preserved (length-based ctor).
  std::string input(reinterpret_cast<const char*>(data), size);

  try {
    YAML::Node node = YAML::Load(input);
    // `node` is intentionally kept alive to end of scope: the parsed node
    // graph (NodeBuilder::Root, alias/shared nodes via m_pMemory) is built
    // here and torn down on scope exit, exercising both paths.
  } catch (const YAML::Exception&) {
    // Expected error path for malformed/crafted YAML.
  } catch (const std::exception&) {
    // Defensive catch-all for any non-YAML std exception.
  }

  return 0;
}
```

4. Save this driver as `driver.cpp` (reads the file argument and calls the harness):

   ```cpp
   #include <cstdio>
   #include <cstdint>
   #include <cstdlib>
   #include <vector>
   extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size);
   int main(int argc, char** argv) {
       if (argc < 2) return 1;
       FILE* f = fopen(argv[1], "rb"); if (!f) return 1;
       fseek(f, 0, SEEK_END); long sz = ftell(f); fseek(f, 0, SEEK_SET);
       std::vector<uint8_t> buf((size_t)sz);
       fread(buf.data(), 1, (size_t)sz, f); fclose(f);
       LLVMFuzzerTestOneInput(buf.data(), buf.size());
       return 0;
   }
   ```

5. Compile and write the exact 96-byte POC:

   ```bash
   clang++ -std=c++11 -O1 -g -fsanitize=address,undefined -fno-sanitize-recover=all -fno-omit-frame-pointer \
       -I include harness.cpp driver.cpp build/libyaml-cpp.a -o yaml_load_fuzzer

   printf 'IQ06ICB8CiAgYf4fIGIKICAQIGMLOjumOjsmCjBvDd7e3t7e3t7e3iqdW1tbWw0NDQ0NDQ0NDQ0NDQ0+DQ18CiAgYQogIGJ/ISEhnVtbW1sNDQ0NDQ0NDQ0NDQ0NPg0N' | base64 -d > poc.yaml
   ```

6. Run it:

   ```bash
   ./yaml_load_fuzzer poc.yaml
   ```

   UBSan output:

   ```
   SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/include/c++/11/bits/stl_deque.h:181:17
   ```

   Exit code 1 (UB trap under `-fno-sanitize-recover=all`).

**Expected behavior**

`PopIndent` / `PushIndentTo` should guard against an empty indentation stack (e.g. return without popping, or report a parser error) instead of calling `m_indents.top()` on an empty container.

**Screenshots**

Image

**Please complete the following information:**

- OS: Linux x86-64 (Ubuntu 22.04)
- yaml-cpp version: 0.9.0 (master @ `3eb39d5808c33aa93a113cb6c668fa2435ff3ecd`)

**Additional context**

- Deterministic: yes — the same 96-byte input always triggers the UB report (verified 3/3 runs).
- Reachability: `YAML::Load(std::string)` (src/parse.cpp:12) on untrusted YAML text; the scanner runs during document parse.
- Severity: high (wild read on empty stack container, deterministic crash / remote DoS).
- CWE: CWE-822 (Untrusted Pointer Dereference).

sxrzh:
fixes #1475 

Now the POC will just cause a `ParserException: end of map not found` and abort.

I also add a guard in `PopIndent` that couldn't be reached theoretically.

### DIFF
--- a/include/yaml-cpp/exceptions.h
+++ b/include/yaml-cpp/exceptions.h
@@ -92,6 +92,8 @@ const char* const BAD_FILE = "bad file";
 const char* const UNEXPECTED_TOKEN_AFTER_DOC = "unexpected token after end of document";
 const char* const NON_UNIQUE_MAP_KEY = "map keys must be unique";
 
+const char* const INDENT_STACK_UNDERFLOW = "indentation stack underflow (please report this bug to yaml-cpp)";
+
 template <typename T>
 inline const std::string KEY_NOT_FOUND_WITH_KEY(
     const T&, typename disable_if<is_numeric<T>>::type* = 0) {
--- a/src/scanner.cpp
+++ b/src/scanner.cpp
@@ -179,7 +179,7 @@ void Scanner::ScanNextToken() {
       m_simpleKeys.top().pKey->status == Token::UNVERIFIED) {
       // if the top of the indents does not match the unverified simple key,
       // just invalidate the simple key and do not pop indent to avoid crash.
-      // eg: an unverified key on a previous line, like issue #1475.
+      // eg: an unverified key crossing lines, like issue #1475.
       if (m_simpleKeys.top().pIndent && !m_indents.empty() &&
           m_indents.top() == m_simpleKeys.top().pIndent) {
         PopIndent();
--- a/src/scanner.cpp
+++ b/src/scanner.cpp
@@ -401,6 +401,9 @@ void Scanner::PopAllIndents() {
 }
 
 void Scanner::PopIndent() {
+  if (m_indents.empty()) {
+    ThrowParserException(ErrorMsg::INDENT_STACK_UNDERFLOW);
+  }
   const IndentMarker& indent = *m_indents.top();
   m_indents.pop();
 
--- a/src/scanner.cpp
+++ b/src/scanner.cpp
@@ -177,7 +177,15 @@ void Scanner::ScanNextToken() {
     //  scalar
     if (!m_simpleKeys.empty() &&
       m_simpleKeys.top().pKey->status == Token::UNVERIFIED) {
+      // if the top of the indents does not match the unverified simple key,
+      // just invalidate the simple key and do not pop indent to avoid crash.
+      // eg: an unverified key on a previous line, like issue #1475.
+      if (m_simpleKeys.top().pIndent && !m_indents.empty() &&
+          m_indents.top() == m_simpleKeys.top().pIndent) {
         PopIndent();
+      } else {
+        InvalidateSimpleKey();
+      }
     }
     return ScanBlockScalar();
   }
--- a/test/integration/handler_test.cpp
+++ b/test/integration/handler_test.cpp
@@ -75,7 +75,7 @@ TEST_F(HandlerTest, CommentOnNewlineOfMapValueWithManySpace) {
 
 // example from issue #1475: a stale simple key (from a plain scalar spanning
 // a line break) must not pop an unrelated indent when parsing a block scalar
-TEST_F(HandlerTest, LiteralScalarWithMultilineTag) {
+TEST_F(HandlerTest, LiteralScalarWithMultiLineUnverifiedPotentialSimpleKey) {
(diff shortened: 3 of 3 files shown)


SGSSGene (approved):
Perfect!
Thank you very much!

Issue #506, opened 2017-06-28
The convert<T>::decode() and Node::as<T>() signature inconsistency.


leni536:
```cpp
template <typename T> inline T Node::as() const;
template <typename T> bool convert<T>::decode(const YAML::Node&, T&);
```

decode() needs a reference for an already constructed object's reference that it can mutate to decode the Node's contents into it. as<T>() returns a brand new T object, it works by *default constructing* a T object then it calling decode() on it. Well, it is problematic if:

* T is not default constructible.
* Even if T is default constructible, it is impossible and/or just expensive to mutate.

I suggest to use the same signature as Node::as<T>(). In case of an error a client can just throw an exception, preferebly TypedBadConversion. It could have the added benefit of marking the error's location more precisely if it happens at a nested subnode.

PR #1479, opened 2026-08-10
feat: Allow decoding for types without default constructor (V1)

### LINKED ISSUE #973: Better support for "default"
YAML is often used for configuration input, and it's a bit difficult to manage defaults:

Current:
```cpp
inline bool validField(const YAML::Node& node, const std::string& field) noexcept {
    if (node.IsDefined() && node[field].IsDefined() && !node[field].IsNull()) {
        return true;
    }
    return false;
}

// The convert struct is helpful only once: for reading the defaults
// default = node.as<Cfg>();

Cfg readConfig(const YAML::Node& node, const Cfg& default) {
    Cfg config = default;
    if (validField(node, "first")) {  config.first = node["first"].as<decltype(config.first)>();
	// ...
	// repeat for N number of fields
	return config;
}
```

Proposed:
```cpp
Cfg readConfig(const YAML::Node& node, Cfg default) {
	// Note the change from const reference to value for default
	node.into<Cfg>(default);  // 100% usage of convert struct
	return default;
}
```

Roadmap:
* Add templated update function to `Node`. Bikeshedding on name is on the table. `into` just reads nice
* Update the `as_if` struct with a function, say `update`, in `node/impl.h` to not create the default variable
* The `into` (newly added templated function) calls the `update` function.

Proposed change in `as_if`:
```cpp
template <typename T, typename S>
struct as_if {
  explicit as_if(const Node& node_) : node(node_) {}
  const Node& node;

  // current, can be replaced with the newer function
  T operator()(const S& fallback) const {
    if (!node.m_pNode)
      return fallback;

    T t;
    if (convert<T>::decode(node, t))
      return t;
    return fallback;
  }

  // new
  bool update(S& incoming) {
	if (!node.m_pNode)
	  return false;
	if (convert<T>::decode(node, incoming))
	  return true;
	return false;
  }

  // compilers are smart enough to reorder `T t` after the inlining of `node.m_pNode` check
  T operator()(const S& fallback) const {
	T t;
	if (update(t))
	  return t;
	return fallback;
  }	
};
```

I can create a PR if needed

SGSSGene:
This PR is inspired by #1010, but instead of creating a new customization point via `decode_dispatcher` it reuses `convert` by using `convert<std::optional<T>>` specialization for non-default constructible classes.

If cmake variable `YAML_CPP_USE_OPTIONAL` is set (automatically set if target is c++17 or above) this new feature is activated. It can be deactivated by setting `YAML_CPP_USE_OPTIONAL` to false to force old behavior if desired.

When decoding a type `T`, it will use `convert<std::optional<T>>` and forward to `convert<T>` if not available.
The signature requirements are as before but by using the additional `std::optional` it allows us to delay the construction of our type T.
Assume we have some type without default constructor:

```cpp
class Vec3 {
  double x, y, z;
public:
  Vec3(double x, double y, double z} : x{x}, y{y}, z{z} {}
};
```

you could write

```cpp
namespace YAML {
template<>
struct convert<std::optional<Vec3>> {
  static bool decode(const Node& node, std::optional<Vec3>& rhs) {
    if(!node.IsSequence() || node.size() != 3) {
      return false;
    }
    rhs.emplace(
        node[0].as<double>(),
        node[1].as<double>(),
        node[2].as<double>()
    );
    return true;
  }
};
}
```

To implement encoding for Vec3 one must still implement `convert<Vec3>` with the `encode` method.

Notes:
fixes #973 #993
alternative for PR #1010  and #1087

PR #1087 breaks API and relies on exception paths which I don't like, since this seems like normal control flow to me.

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -200,6 +200,38 @@ Vec3 v = node["start"].as<Vec3>();
 node["end"] = Vec3(2, -1, 0);
 ```
 
+## Non-default constructible types (requires c++17 and newer)
+Yaml-cpp also supports types that are not default constructible. For this one need to specialize `YAML::convert<std::optional<>>`.
+Assuming you have:
+
+```cpp
+class Vec3 {
+  double x, y, z;
+public:
+  Vec3(double x, double y, double z} : x{x}, y{y}, z{z} {}
+};
+```
+
+you could write (for encoding the previous `convert<Vec3>` with the `encode` method is required)
+
+```cpp
+namespace YAML {
+template<>
+struct convert<std::optional<Vec3>> {
+  static bool decode(const Node& node, std::optional<Vec3>& rhs) {
+    if(!node.IsSequence() || node.size() != 3) {
+      return false;
+    }
+    rhs.emplace(
+        node[0].as<double>(),
+        node[1].as<double>(),
+        node[2].as<double>()
+    );
+    return true;
+  }
+};
+}
+
 ## Partial specialization
 
 If you need to specialize the `convert` struct for a set of types instead of just one you can use partial specialization with the help of `std::enable_if` (SFINAE).
@@ -224,7 +256,7 @@ public:
     node["a"] = a;
     return node;
   }
-  
+
   int a;
 };
 
@@ -252,7 +284,7 @@ public:
 
 // Implementation of convert::{encode,decode} for all classes derived from or being A
 namespace YAML {
-  template<typename T> 
+  template<typename T>
   struct convert<T, typename std::enable_if<std::is_base_of<A, T>::value>::type> {
     static Node encode(const T &rhs) {
       Node node = rhs.emit();
--- a/include/yaml-cpp/node/impl.h
+++ b/include/yaml-cpp/node/impl.h
@@ -15,6 +15,17 @@
 #include <sstream>
 #include <string>
 
+// Check YAML_CPP_USE_OPTIONAL is set and language standard provides supports
+// if available and supported include required header
+// otherwise remove YAML_CPP_USE_OPTIONAL definition
+#ifdef YAML_CPP_USE_OPTIONAL
+#if __cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
+#include <optional>
+#else
+#undef YAML_CPP_USE_OPTIONAL
+#endif
+#endif
+
 namespace YAML {
 inline Node::Node()
     : m_isValid(true), m_invalidKey{}, m_pMemory(nullptr), m_pNode(nullptr) {}
@@ -94,6 +105,54 @@ inline NodeType::value Node::Type() const {
 // access
 
 // template helpers
+#ifdef YAML_CPP_USE_OPTIONAL
+template <typename T>
+struct decode_box : std::optional<T> {
+  decode_box() = default;
+  template <typename S>
+  decode_box(S&&) {}
+};
+
+template <typename T>
+struct convert<decode_box<T>> {
+  static bool decode(const Node& node, decode_box<T>& rhs) {
+    return convert<std::optional<T>>::decode(node, rhs);
+  }
+};
+template <typename T>
+struct convert<std::optional<T>> {
+  static bool decode(const Node& node, std::optional<T>& rhs) {
+    if (node.IsNull()) {
+      return false;
+    }
+    rhs.emplace();
+    if (!convert<T>::decode(node, *rhs)) {
+      rhs.reset();
+      return false;
+    }
+    return true;
+  }
+};
+
+
+#else
+
+template <typename T>
+struct decode_box {
+  T t;
+  T& operator*() {
+    return t;
+  }
+};
+template <typename T>
+struct convert<decode_box<T>> {
+  static bool decode(const Node& node, decode_box<T>& rhs) {
+    return convert<T>::decode(node, *rhs);
+  }
+};
+
+#endif
+
 template <typename T, typename S>
 struct as_if {
   explicit as_if(const Node& node_) : node(node_) {}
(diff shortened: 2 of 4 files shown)


SGSSGene:
Maybe it also enough to consider it as a fix for #506

docs/Tutorial.md
```diff
@@ -200,6 +200,38 @@ Vec3 v = node["start"].as<Vec3>();
 node["end"] = Vec3(2, -1, 0);
 ```
 
+## Non-default constructible types (requires c++17 and newer)
+Yaml-cpp also supports types that are not default constructible. For this one need to specialize `YAML::convert<std::optional<>>`.
```
alex-thiessen-for-siemens, line 204: _default constructible_ -> _default-constructible_

alex-thiessen-for-siemens, line 204: _For this one need to specialize_ -> _For this, one needs to specialize_

docs/Tutorial.md
```diff
@@ -200,6 +200,38 @@ Vec3 v = node["start"].as<Vec3>();
...
+
+```cpp
+class Vec3 {
+  double x, y, z;
+public:
+  Vec3(double x, double y, double z} : x{x}, y{y}, z{z} {}
+};
+```
+
+you could write (for encoding the previous `convert<Vec3>` with the `encode` method is required)
```
alex-thiessen-for-siemens, line 215: _for encoding the previous_ -> _for encoding, the previous_

docs/Tutorial.md
```diff
@@ -200,6 +200,38 @@ Vec3 v = node["start"].as<Vec3>();
...
+
+```cpp
+namespace YAML {
+template<>
+struct convert<std::optional<Vec3>> {
+  static bool decode(const Node& node, std::optional<Vec3>& rhs) {
```
alex-thiessen-for-siemens, line 221: this interface now has two `bool`s (one return value, one `std::optional`-embedded), what's their combinatory semantics?

docs/Tutorial.md
```diff
@@ -252,7 +284,7 @@ public:
 
 // Implementation of convert::{encode,decode} for all classes derived from or being A
 namespace YAML {
-  template<typename T> 
+  template<typename T>
```
alex-thiessen-for-siemens, line 287: is it possible to clean all whitespace in a separate effort, adding a static check for that?

include/yaml-cpp/node/impl.h
```diff
@@ -15,6 +15,17 @@
 #include <sstream>
 #include <string>
 
+// Check YAML_CPP_USE_OPTIONAL is set and language standard provides supports
+// if available and supported include required header
+// otherwise remove YAML_CPP_USE_OPTIONAL definition
+#ifdef YAML_CPP_USE_OPTIONAL
+#if __cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
```
alex-thiessen-for-siemens, line 22: That's a weird check, maybe C++20 should be the base line so that more straightforward feature-test macros could be used instead.

docs/Tutorial.md
```diff
@@ -200,6 +200,38 @@ Vec3 v = node["start"].as<Vec3>();
 node["end"] = Vec3(2, -1, 0);
 ```
 
+## Non-default constructible types (requires c++17 and newer)
```
alex-thiessen-for-siemens, line 203: _c++17 and newer_ -> _C++17 or newer_

docs/Tutorial.md
```diff
@@ -200,6 +200,38 @@ Vec3 v = node["start"].as<Vec3>();
...
+  Vec3(double x, double y, double z} : x{x}, y{y}, z{z} {}
+};
+```
+
+you could write (for encoding the previous `convert<Vec3>` with the `encode` method is required)
+
+```cpp
+namespace YAML {
+template<>
+struct convert<std::optional<Vec3>> {
```
alex-thiessen-for-siemens, line 220: I'd doubt that `std::optional` is a great choice here, `std::expected` seems to be more appropriate, as it enables users to provide informative failure data more easily than e.g. via exception handling. It's been available in mainstream toolchains for 3-4 years now.

alex-thiessen-for-siemens (commented):
I really don't like that interface, could we have something like `template<>
struct convert<std::expected<Vec3>> static auto decode(const Node& node) -> std::expected<Vec3>` instead?

SGSSGene:
@alex-thiessen-for-siemens 
Thank you for your great feedback!

I think there is still plenty of room for discussion about this (or the other PRs).
I agree that `std::expected` has a lot more charm than `std::optional`. Also I agree that `auto decode(Node& node) -> resultType` is much nice than `bool decode(Node& node, resultType& type)`.

Currently yaml-cpp is based on c++11. I feel weird to add features that only work with newer c++ versions. Which is also why I am not sure if it really matters if we use `std::optional` (c++17) or `std::expected` (c++23). We also could introduce a custom YAML-type that kind of works like `std::optional` but is c++11 compatible....

With a similar argument I used `bool decode(Node& node, resultType& type)` this is much easier to implement with c++17 (or older). The signature `auto decode(Node& node) -> resultType` is only manageable with c++20 and concept is being used. Which might be another argument going all in for c++23 and use `std::expected`.

Opinions from anyone are very welcome. Especially from people who would have concrete use cases and can easier imagine the pro/cons of the different APIs.

SGSSGene:
> I really don't like that interface, could we have something like `template<> struct convert<std::expected<Vec3>> static auto decode(const Node& node) -> std::expected<Vec3>` instead?

If we switch to `c++23` we could even go for specializing `Vec3` instead of `std::expected<Vec3>` which I think is even prettier:
```
template<> struct convert<Vec3> {
    static auto decode(const Node& node) -> std::expected<Vec3>
}
```

alex-thiessen-for-siemens:
Sounds like a policy decision could be handy regarding how to proceed with adding functionality which would benefit or require recent C++ versions' features, like `std::optional`, `std::expected` or reflection.

What this PR brings is added convenience and performance of _not_ having to dynamically allocate non-default-constructible objects, right? That would count as optimization which can be offered exclusively to users of recent C++ versions, IMHO. This would avoid the awkward double-bool situation here.

Those who stick to old yet battle-tested toolchain versions don't usually crave for most convenient library features, I'd suggest.

SGSSGene:
@alex-thiessen-for-siemens can you take a look at #1488 as a potential replacement for this PR?

SGSSGene:
Closed in favor of #1488

PR #1483, opened 2026-08-12
feat: add IWYU pragmas for public/private headers (#1386)

### LINKED ISSUE #1386: Add Include What You Use pragmas to help IWYU find the correct headers to include.
Include What You Use ([IWYU](https://include-what-you-use.org/)) (`A tool for use with clang to analyze #includes in C and C++ source files`) is able to fix (read: add and remove) includes automatically based on what types are being referenced in a source and or header file.

However, it fails to properly include yaml-cpp's headers because the forward declarations and actual implementation are separated across multiple includes. Which IWYU will not able to fix automatically. Resulting in linking errors because it didn't include the implementation of a function/class template.

An example of IWYU's usage can be seen in googletest/googlemock:
https://github.com/google/googletest/blob/1b96fa13f549387b7549cc89e1a785cf143a1a50/googletest/include/gtest/gtest-matchers.h#L35-L37
This marks the `gtest-matchers.h` as a private include, and instead of including `gtest-matchers.h` `gtest/gtest.h` should be included instead.
The `pragma: friend` directives mark this file for 'private use' from all the headers inside the `gtest/` and `gmock/` folders

https://github.com/google/googletest/blob/1b96fa13f549387b7549cc89e1a785cf143a1a50/googletest/include/gtest/gtest.h#L63-L72
Shows how that gtest/gtest.h is 'exporting' symbols found in all the 'private' headers.

Using these directives gives IWYU the tools to include `gtest/gtest.h` whenever you depend on any type defined in any of its 'private headers'.

Using IWYU on `yaml-cpp` results in the inclusion of, for example, `yaml-cpp/node/node.h`, alone. Which is correct-ish when you use `YAML::Node` and it will compile but will fail to link because it is missing the inclusion of `yaml-cpp/impl.h`

I would like to propose to add IWYU directives to private/public headers so it becomes IWYU compatible.

wangshuo-wsws:
## Summary

Adds [Include What You Use](https://include-what-you-use.org/) pragmas across yaml-cpp public headers, following the same pattern as googletest:

- `yaml-cpp/yaml.h` marks its transitive includes with `// IWYU pragma: export`
- Internal headers mark `// IWYU pragma: private, include "yaml-cpp/yaml.h"` plus `friend "yaml-cpp/.*"`

This lets IWYU redirect `#include "yaml-cpp/node/node.h"` to the umbrella header so users get `impl.h` and other required definitions, avoiding link errors.

Fixes #1386.

## Test plan

- [x] Headers compile unchanged (comments only)
- [ ] Run IWYU on a sample TU using `YAML::Node` and confirm it suggests `yaml-cpp/yaml.h`

### DIFF
--- a/include/yaml-cpp/anchor.h
+++ b/include/yaml-cpp/anchor.h
@@ -1,11 +1,20 @@
 #ifndef ANCHOR_H_62B23520_7C8E_11DE_8A39_0800200C9A66
 #define ANCHOR_H_62B23520_7C8E_11DE_8A39_0800200C9A66
 
+
+
+
 #if defined(_MSC_VER) ||                                            \
     (defined(__GNUC__) && (__GNUC__ == 3 && __GNUC_MINOR__ >= 4) || \
      (__GNUC__ >= 4))  // GCC supports "pragma once" correctly since 3.4
 #pragma once
+
+
 #endif
+
+// IWYU pragma: private, include "yaml-cpp/yaml.h"
+// IWYU pragma: friend "yaml-cpp/.*"
+
 
 #include <cstddef>
 
--- a/include/yaml-cpp/binary.h
+++ b/include/yaml-cpp/binary.h
@@ -1,11 +1,20 @@
 #ifndef BASE64_H_62B23520_7C8E_11DE_8A39_0800200C9A66
 #define BASE64_H_62B23520_7C8E_11DE_8A39_0800200C9A66
 
+
+
+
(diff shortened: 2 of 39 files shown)


SGSSGene (approved):
Thank you very much for this contribution!
Personally not using IWYU but pretty sure people would love to use it.

(Also reminds me, we should remove the code for #pragma once, that creates some noise in the files).

PR #1485, opened 2026-08-22
Fix static pkg-config compile flags

### LINKED ISSUE #1345: yaml-cpp.pc with static linking on Windows
The file should include `Cflags.private: -DYAML_CPP_STATIC_DEFINE`.  Without that, one gets undefined symbols when linking a static library, because the headers expect dynamically linked symbols (with the corresponding prefix on Windows).

henriquejsza:
## Summary
- emit `YAML_CPP_STATIC_DEFINE` through `Cflags.private` for static pkg-config metadata
- keep shared-library metadata free of the static definition
- add a CTest regression check for both build modes

Fixes #1345

## Testing
- static and shared CMake/Ninja Release builds
- `ctest` (3/3 in each build)
- install and CMake package consumers in both modes
- install and pkg-config consumers, including `--static --cflags`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,11 @@ write_basic_package_version_file(
   "${PROJECT_BINARY_DIR}/yaml-cpp-config-version.cmake"
   COMPATIBILITY AnyNewerVersion)
 
+if (YAML_BUILD_SHARED_LIBS)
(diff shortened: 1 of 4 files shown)


SGSSGene (approved):
Thank you! Hopefully this will reduce some of the windows linkage issues people have!

PR #1486, opened 2026-08-22
Reject failed input streams

### LINKED ISSUE #408: Load and LoadAll ignore failbit and badbit on istream

henriquejsza:
Fixes #408

- reject failed input streams before and during parsing
- preserve empty and EOF-only streams
- add regression coverage for `Load` and `LoadAll`

Tests:
- CMake C++11 static build and full test suite
- Clang C++23 shared build and full CTest
- GCC C++20 ASan/UBSan suite
- Bazel `build :all` and `test test`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,6 +79,9 @@ jobs:
         - name: Run Tests
           shell: bash
           run: |
+            if [[ "${{ matrix.os }}" == "windows-latest" && "${{ matrix.generator }}" == "MinGW Makefiles" ]]; then
+              export PATH="/c/mingw64/bin:$PATH"
+            fi
             ctest \
               --test-dir build \
               --build-config ${{ env.CMAKE_BUILD_TYPE }} \
--- a/include/yaml-cpp/exceptions.h
+++ b/include/yaml-cpp/exceptions.h
@@ -98,6 +98,7 @@ const char* const INVALID_ANCHOR = "invalid anchor";
 const char* const INVALID_ALIAS = "invalid alias";
 const char* const INVALID_TAG = "invalid tag";
 const char* const BAD_FILE = "bad file";
+const char* const BAD_STREAM = "bad stream";
 const char* const UNEXPECTED_TOKEN_AFTER_DOC = "unexpected token after end of document";
 const char* const NON_UNIQUE_MAP_KEY = "map keys must be unique";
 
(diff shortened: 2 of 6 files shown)


henriquejsza:
CI follow-up: the six Windows x64 MinGW jobs built successfully but hung in CTest when launched under Git Bash. In that shell, `/mingw64/bin` precedes the compiler-matching `C:\mingw64\bin`; prepending `/c/mingw64/bin` only for the affected matrix entries makes the full suite complete. This was verified in [diagnostic run 32585105051](https://github.com/henriquejsza/yaml-cpp/actions/runs/32585105051), where the job passed in 1m40s. Commit `f1c2b7f` applies only that PATH adjustment; the refreshed upstream workflow is currently awaiting maintainer approval.

SGSSGene (approved):
Thank for this fix! Especially finding the issue in the CI.

Short remark: This PR now contains three changes:
- fixing bad input streams
- fixing CI
- and cleanup raw pointers -> std::unique_ptr

I would prefer to have the raw pointers -> std::unique_ptr in a separately PR. It makes following the changes easier.

But again! This fixes are appreciated!

PR #1487, opened 2026-08-30
fix orphaned utf-16 high surrogate dropping the next code unit

soma0212:
StreamInUtf16 handles a high surrogate that isn't followed by a low surrogate by queuing a replacement character, but when the following unit is an ordinary BMP code unit it then queues `ch` (the orphaned high surrogate) instead of `chLow` (the unit that actually followed). That encodes a lone surrogate as ill-formed UTF-8 (U+D800 becomes ED A0 80) and drops the real character. It only shows up on malformed UTF-16 input, so the parser silently corrupts the scalar instead of substituting U+FFFD the way the replacement path intends.

Queue `chLow` so the orphan collapses to a single U+FFFD and the following character is still decoded. The bug lives in the decoder because that is where the replacement-character contract is enforced; callers only see the already-decoded UTF-8. Added a regression test next to the existing encoding tests covering a high surrogate followed by a plain character.

### DIFF
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -395,7 +395,7 @@ void Stream::StreamInUtf16() const {
         // Deal with the next UTF-16 unit
         if (chLow < 0xD800 || chLow >= 0xE000) {
           // Easiest case: queue the codepoint and return
-          QueueUnicodeCodepoint(m_readahead, ch);
+          QueueUnicodeCodepoint(m_readahead, chLow);
           return;
         }
         // Start the loop over with the new high surrogate
--- a/test/integration/encoding_test.cpp
+++ b/test/integration/encoding_test.cpp
@@ -179,10 +179,10 @@ TEST_F(EncodingTest, UTF32BE_BOM) {
   Run();
 }
 
-// An orphaned UTF-16 high surrogate must decode to a single replacement
-// character, and the code unit after it must still be decoded. Previously the
-// high surrogate was re-emitted as an ill-formed UTF-8 sequence and the
-// following character was dropped.
+// An orphaned UTF-16 high surrogate must decode to a single U+FFFD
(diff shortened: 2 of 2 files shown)


SGSSGene (commented):
Thank you for this PR!

Never mind: `\xEF\xBF\xBD` encodes the replacement character in UTF-8 

--- Old message:
I need some help understanding the expected result (I am not well versed in Utf16):

Why do we expect `"\xEF\xBF\xBD" "A"`? Or the question phrased slightly differently: why dos `0xD8000` translate to `\xEF\xBF\xBD`?
I see that `\xEF\xBF` is a BMP, but I do not know what `\xEF\xBF` stands for ([this does not seem right](https://www.compart.com/de/unicode/U+EFBF)).
And why does it get followed by a `\xBD`?

It seems like I am missing something important!

soma0212:
Right, `EF BF BD` is U+FFFD (the replacement character) in UTF-8, so the orphaned high surrogate collapses to a single U+FFFD and the `A` that used to get dropped comes through as-is. Happy to expand the test comment if that mapping is worth spelling out for the next reader.

SGSSGene:
> Right, `EF BF BD` is U+FFFD (the replacement character) in UTF-8, so the orphaned high surrogate collapses to a single U+FFFD and the `A` that used to get dropped comes through as-is. Happy to expand the test comment if that mapping is worth spelling out for the next reader.

That is a good I idea! That might help the next person who reads it!

soma0212:
Done. The test now says up front that the orphan should collapse to U+FFFD, and there's a note right above the EXPECT_EQ spelling out that `\xEF\xBF\xBD` is U+FFFD in UTF-8, plus what the scalar looked like before the fix (`EF BF BD ED A0 80`, with the 'A' lost) so the two byte strings are easy to compare.

PR #1489, opened 2026-09-02
Enforce whitespace consistency in maintained files

alex-thiessen-for-siemens:
## Summary

This standalone cleanup is motivated by the incidental whitespace changes mixed into #1488. It normalizes the existing maintained-file baseline and adds a permanent check so future semantic PRs do not need to carry unrelated whitespace edits.

## Policy

- No trailing spaces or tabs.
- Each line uses LF as its end-of-line character.
- Non-empty files end with exactly one final LF.
- The vendored `test/googletest-1.16.0/` tree is excluded.

The cleanup normalizes maintained text to LF and leaves the vendored tree unchanged. You may run `python3 util/check-whitespace.py` to check whitespace before submitting a pull request.

## CI

The new `Whitespace` workflow runs on pushes, pull requests, and manual dispatch. Please add the `Whitespace` check to the required status checks for `master` after this workflow is available. This account does not have permission to change the upstream branch rules.

This PR intentionally does not modify or comment on #1488.


SGSSGene:
This is awesome! Thank you!

May I ask, which files end in CRLF? the only two that I can find are [yaml-cpp.natvis](https://github.com/jbeder/yaml-cpp/blob/master/src/contrib/yaml-cpp.natvis) and [yaml-cpp.natvis.md](https://github.com/jbeder/yaml-cpp/blob/master/src/contrib/yaml-cpp.natvis.md)

This should definitely be also fixed and changed to LF.

I would say the policy is:
- No trailing spaces or tabs.
- Each line ends with a newline (LF)
- The vendored test/googletest-1.16.0/ tree is excluded.

CONTRIBUTING.md
```diff
@@ -21,6 +21,15 @@ Please verify the tests pass by configuring CMake with `-D YAML_CPP_BUILD_TESTS=
...
 If you are adding functionality, add tests accordingly. Note that the "spec tests" are reserved for examples directly from the YAML spec, so if you have new examples, put them in other test files.
 
+# Whitespace
+
+Maintained text files must not contain trailing spaces or tabs and must have
+exactly one final newline. Each file may use either LF or CRLF line endings,
+but must not mix the two. The vendored GoogleTest sources are excluded from
+this check.
+
+Run `python3 util/check-whitespace.py` before submitting a pull request.
```
SGSSGene: I would like to make it less demanding (The CI will hopefully fail anyways)

```suggestion
You may run `python3 util/check-whitespace.py` to fix white spaces before submitting a pull request.
```

CONTRIBUTING.md
```diff
@@ -21,6 +21,15 @@ Please verify the tests pass by configuring CMake with `-D YAML_CPP_BUILD_TESTS=
 
 If you are adding functionality, add tests accordingly. Note that the "spec tests" are reserved for examples directly from the YAML spec, so if you have new examples, put them in other test files.
 
+# Whitespace
+
+Maintained text files must not contain trailing spaces or tabs and must have
+exactly one final newline. Each file may use either LF or CRLF line endings,
+but must not mix the two. The vendored GoogleTest sources are excluded from
+this check.
```
SGSSGene:
```suggestion
Maintained text files must not contain trailing spaces or tabs.
Each line ends with a single newline LF character. The vendored GoogleTest sources are excluded from
this check.
```

SGSSGene (changes requested):
This is great!

I would like to pin it down to only LF symbols.
And even though it is called a 'newline' character, I think it is easier/better to think of it as a 'end of line' character, which makes it very clear when/how to use it and why there has to be one at the end of a non-empty file.

SGSSGene:
@jbeder I think this is a great change to keep maintain cleanliness of all files which reduces noise in PRs for whitespaces. This needs your approval and intervention to add there CI workflow as an requirement for PRs to pass.

CONTRIBUTING.md
alex-thiessen-for-siemens: Updated. The contributor instruction is now optional and accurately says the script checks whitespace; CI remains the enforcement mechanism.

CONTRIBUTING.md
alex-thiessen-for-siemens: Updated to make LF the only accepted end-of-line style. `.editorconfig`, the checker, and the normalized baseline now enforce LF for maintained text; the vendored GoogleTest tree remains excluded.

alex-thiessen-for-siemens:
Addressed. `src/contrib/yaml-cpp.natvis` and `src/contrib/yaml-cpp.natvis.md` were the only maintained files using CRLF; both now use LF. The policy and checker require LF for maintained text, while the vendored `test/googletest-1.16.0/` tree remains excluded. Making `Whitespace` a required status check still needs an upstream maintainer with branch-rule permissions.

alex-thiessen-for-siemens:
Initially, I decided that _natvis_-related files should remain CRLF, as these are for Windows users only. Hope Visual Studio can deal with LF natvis, else an exception to the EOL check would be needed.

SGSSGene (approved):
Thank you!

PR #1490, opened 2026-09-03
Test a stream read that fails on the first attempt

subtilitas:
FailingStreamBuf serves 32 bytes before it throws, so Stream's constructor completes and the failure is raised later, while scanning.  Stream's constructor ends in ReadAheadTo(0), so a buffer that throws immediately raises it from inside the constructor instead - a path nothing covered.

It is also the path that leaked before m_pPrefetched became a unique_ptr: a destructor does not run for an object whose constructor threw.  Restoring the raw pointer keeps every existing test passing and green, while this one reports 2048 bytes leaked under AddressSanitizer.


SGSSGene:
I am sorry, I have trouble understanding what this PR is for. To me it sounds you are raising an issue and this PR demonstrates it?

subtilitas:
> I am sorry, I have trouble understanding what this PR is for. To me it sounds you are raising an issue and this PR demonstrates it?

Yes, the issue itself has been fixed on master, but your test is incomplete so that's my suggestion.
I ran into the leak with 0.8 while testing my library and found it closed but still untested.
That's what I propose here.

SGSSGene (approved):
Thank you!

PR #1491, opened 2026-09-06
Fix Base64 padding separated by whitespace

alex-thiessen-for-siemens:
## Summary

Fix Base64 decoding when whitespace separates padding characters.

`DecodeBase64("TQ= =")` previously emitted an extra zero byte because it
checked the raw preceding input character after whitespace had been skipped.
The decoder now tracks whether the previous non-whitespace Base64 character
was padding. The focused test verifies that the result remains the single byte
`M`.

## Validation

- C++11 CMake/CTest: 3/3 tests passed.
- Pinned Debian Trixie evaluation: CMake/CTest, ASan/UBSan, Valgrind,
  clang-format, clang-tidy, exhaustive cppcheck, Bazel, and Bzlmod passed.
- PyYAML 6.0.3, using YAML 1.1 `!!binary` native-value semantics, matched
  the decoded byte `4d`.
- libyaml 0.2.5 matched the yaml-cpp parser/event stream for the fixture.

### DIFF
--- a/src/binary.cpp
+++ b/src/binary.cpp
@@ -75,6 +75,7 @@ std::vector<unsigned char> DecodeBase64(const std::string &input) {
 
   unsigned value = 0;
   std::size_t cnt = 0;
+  bool previous_was_padding = false;
   for (std::size_t i = 0; i < input.size(); i++) {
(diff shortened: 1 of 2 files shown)


SGSSGene (approved):
Amazing! thank you!

PR #1492, opened 2026-09-08
Terminate literal scalars before emitting comments

### LINKED ISSUE #800: Comment after literal string
By running this code:
```c++
#include <yaml-cpp/yaml.h>
#include <iostream>
#include <fstream>

int main() {
    {
        std::ofstream f("hola.yaml");
        YAML::Emitter emitter(f);
        emitter << YAML::Literal << "Hello\nworld" << YAML::Comment("Un comentario");
    }
    {
        std::ifstream f("hola.yaml");
        YAML::Node node = YAML::Load(f);
        std::cout << node.as<std::string>() << std::endl;
    }
    return 0;
}
```
The first part generates the following content:
```yaml
|
  Hello
  world # A comment
```
When the second part reads it back, the comment is read as part of the literal. One of them, reading or writing is wrong. Specs says that no comments can be placed inside an scalar, since a literal is an scalar and it terminates with the indentation back, it sounds like the reading is right and the writing is wrong.

When a scalar is inserted in literal mode comments inserted next should be at the next line. and outside of the indentation. Like this:

```yaml
|
  Hello
  world
# A comment
```

vzer200:
Fixes #800.

Track the output position after a literal string or binary value. If a comment immediately follows, terminate the literal before writing the comment. This prevents the comment from becoming scalar data and preserves the trailing newlines expected by the literal's chomping indicator. An explicit `Newline` or intervening output prevents a second newline from being added.

Seven regressions cover string round trips, strip/clip/keep behavior, explicit newlines, nested maps, literal keys, binary values and an external output stream. All 1,067 tests and both CMake contract checks pass in a C++11 Debug build with Clang on Windows; changed ranges pass clang-format.

Independent review also compared 648 combinations against the existing explicit-Newline workaround, with identical output in every case. Existing blank-only/leading-whitespace/CRLF literal limitations remain outside this change.

AI-assisted implementation, independently reviewed and tested by another coding agent.


alex-thiessen-for-siemens:
The commit message body is empty despite a detailed PR description. (by Gemini 3.8 Flash High)

alex-thiessen-for-siemens:
One edge case to consider is an empty literal string:

```cpp
out << Literal << "" << Comment("comment");
```

For non-empty strings, `WriteLiteralString` leaves the last line un-terminated so subsequent formatting can supply the newline. But for `size == 0`, `WriteLiteralString` already writes `|-\n`. The check in `Write(const _Comment&)` then adds another `\n`, producing an extra blank line between the header and the comment:

```yaml
|-

# comment
```

instead of:

```yaml
|-
# comment
```

While parsers strip the blank line for `|-`, it creates an unintended blank line in the output.

(by Gemini 3.8 Flash High)

PR #1493, opened 2026-09-08
Add tests for YAML 1.2 specification examples

### LINKED ISSUE #38: Full test suite for the YAML 1.2 spec
```
All examples from the YAML 1.2 spec should be written up as test.
```

Original issue reported on code.google.com by `jbe...@gmail.com` on 5 Sep 2009 at 11:17

alex-thiessen-for-siemens:
## Summary

- Add node and event-handler coverage for the remaining YAML 1.2 specification examples.
- Extend the shared specification fixtures used by the integration tests.
- Keep the coverage organized by specification example so each case is traceable to the YAML 1.2 text.

## Testing

- CMake/CTest suite passes with 1,110 tests.

Closes #38

### DIFF
--- a/test/integration/handler_spec_test.cpp
+++ b/test/integration/handler_spec_test.cpp
@@ -1690,5 +1690,13 @@ TEST_F(HandlerSpecTest, Ex8_22_BlockCollectionNodes) {
   EXPECT_CALL(handler, OnDocumentEnd());
   Parse(ex8_22);
 }
+
+TEST_F(HandlerSpecTest, Ex9_1_DocumentPrefix) {
+  EXPECT_CALL(handler, OnDocumentStart(_));
+  EXPECT_CALL(handler, OnScalar(_, "?", 0, "Document"));
+  EXPECT_CALL(handler, OnDocumentEnd());
+  Parse(ex9_1);
+}
+
 }  // namespace
 }  // namespace YAML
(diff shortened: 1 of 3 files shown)


test/integration/handler_spec_test.cpp
```diff
@@ -728,6 +728,12 @@ TEST_F(HandlerSpecTest, Ex5_9_DirectiveIndicator) {
   Parse(ex5_9);
 }
 
+TEST_F(HandlerSpecTest, Ex5_10_InvalidUseOfReservedIndicators) {
+  // SPEC_GAP: yaml-cpp currently accepts neither reserved indicator as a
+  // plain-scalar starter, so keep the specification rejection discoverable.
+  EXPECT_THROW(IgnoreParse(ex5_10), ParserException);
+}
```
SGSSGene, line 733: This seems weird, if there is a SPEC_GAP, shouldn't this test be Disabled? (Why is it passing?)

SGSSGene (changes requested):

test/integration/handler_spec_test.cpp
alex-thiessen-for-siemens: You are right to question the marker. I misclassified Example 5.10: YAML 1.2.2 explicitly marks this input invalid because `@` and backtick cannot start plain scalars. yaml-cpp correctly rejects it, which is why the enabled test passes. I removed the erroneous `SPEC_GAP` marker and kept the test enabled. The actual gaps on this branch remain disabled separately.

Issue #986, opened 2021-04-20
Single carriage return line endings cause an exception


jwelsch:
### Description
YAML files that use single carriage returns for line endings can cause an exception to be thrown. There are at least two scenarios that can cause an exception. The first is when the carriage return is on the first line. A `YAML::ParserException` is thrown when calling `YAML::LoadFile()`. The second is when the carriage return is on a nested node. It can throw a `YAML::TypedBadConversion` when calling `Node::as<T>()`.

### Reproduction Steps
To throw a `YAML::ParserException`, execute the following code with the attached file:
````
auto root = YAML::LoadFile("first-line-cr.yaml"); // Exception thrown here
````

To throw a `YAML::TypedBadConversion`, execute the following code with the attached file:
````
auto root = YAML::LoadFile("nested-node-cr.yaml");
auto foo = root["foo"];
auto bar = foo["bar"];
auto hello = bar["hello"];
auto a = hello.as<std::string>(); // Exception thrown here
````

Files containing example YAML can be found in this ZIP:
[examples.zip](https://github.com/jbeder/yaml-cpp/files/6345274/examples.zip)

**Comment by Akaame** [comment]
This turned out to be pretty ugly.

I think the following diff patches it.

```diff
diff --git a/src/exp.h b/src/exp.h
index 301449e..a0d6d76 100644
--- a/src/exp.h
+++ b/src/exp.h
@@ -37,7 +37,7 @@ inline const RegEx& Blank() {
   return e;
 }
 inline const RegEx& Break() {
-  static const RegEx e = RegEx('\n') | RegEx("\r\n");
+  static const RegEx e = RegEx('\n') | RegEx("\r\n") | RegEx('\r');
   return e;
 }
 inline const RegEx& BlankOrBreak() {
diff --git a/src/stream.cpp b/src/stream.cpp
index b1aa092..ab83293 100644
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -262,7 +262,7 @@ char Stream::get() {
   AdvanceCurrent();
   m_mark.column++;
 
-  if (ch == '\n') {
+  if (ch == '\n' || (ch == '\r' && peek() != '\n')) {
     m_mark.column = 0;
     m_mark.line++;
   }
```

SGSSGene:
This is fixed in #1312

PR #1494, opened 2026-09-09
treat a carriage return as a line break when emitting

soma0212:
The scanner takes a bare CR as a line break (`Exp::Break`, after #986 and #1309), but the emitter helpers still only look for LF:

- `WriteComment` re-prefixes `#` on LF only, so `Comment("hi\rinjected: pwned")` on a one-key map reparses as two keys, the comment tail landing as document structure
- `IsValidSingleQuotedScalar` rejects LF but not CR, so `'a\rb'` goes out raw and folds back to `a b`
- `IsValidLiteralScalar` checks for neither, so a CR in a block scalar comes back as LF
- `WriteSingleQuotedString` keeps its own bail-out in step with its validator

Plain scalars already fall back to double quotes for this (#607); the two remaining explicit formats now do the same, while the comment writer handles the break in place since it owns the `#` continuation. A CRLF pair collapses to one break instead of two.


SGSSGene:
This is a great addition! Thank you!

